PR #19, opened 2017-02-21
network.IDN_show_punycode

earthlng:
my draft for `network.IDN_show_punycode`
added under 2600 but it would maybe also fit under 0800 (?)
the title and that one line are quite long, feel free to improve the wording etc.


Atavic:
> added under 2600
seems the optimal location to me.

earthlng:
pyllyukko has the following url included, is it too much if we add it too?
http://kb.mozillazine.org/Network.IDN_show_punycode

Atavic:
Your choice is better! Also, I think that a russian/arabic/hebrew user always sees the URL in its own alphabet but I can't confirm (I have a latin system language).

Roman-Nopantski:
@earthlng  If you're happy with my little modifications, merge the pull request and them commit to master

earthlng:
thx Pants, much better with your mods

PR #22, opened 2017-02-21
network.IDN_show_punycode

earthlng:
- removed the 'period' at the end of the pref description (we usually don't have one there)
- re-aligned the 2nd line of 'the downside'


earthlng:
oh damn, ended up with a 2nd pull request lol. my bad.
I will merge it like this and delete the first one.
@Roman-Nopantski feel free to remove the extra whitespace directly in the master if you don't like the alignment

PR #38, opened 2017-03-03
signon.autofillForms.http + contextual warning

earthlng:
my proposal as a PR, let's have a discussion if necessary and maybe wait with merging this until FF52 stable is released

### DIFF
--- a/user.js
+++ b/user.js
@@ -454,7 +454,7 @@ user_pref("security.insecure_password.ui.enabled", true);
 user_pref("browser.fixup.hide_user_pass", true);
 /* 0909: disable formless login capture for Password Manager (FF51+) ***/
 user_pref("signon.formlessCapture.enabled", false);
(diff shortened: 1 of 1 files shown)


Roman-Nopantski:
Are any of these 3 commits meant to have a (FF52+) at the end?

Roman-Nopantski:
I'm also OK with committing these now. The v51 is archived, then we start on the next version. It's either that or we create our releases just prior to the next one - which means less time, relevance. Besides, we do add version numbers to prefs. We have tons of time anyway, no need to hurry.

But I will still only change the version number, code name, lyric, and date just prior to a release.

earthlng:
> Are any of these 3 commits meant to have a (FF52+) at the end?

Yes, I forgot to add that. I'll add it.

> committing these now.

I only mentioned to wait in case the prefs are backed out in stable for whatever reason.
But yeah, not really that important, we can always remove it again.

Issue #5, opened 2017-02-18
discussion: referers section 1600 & new prefs in 52


Roman-Nopantski:
The 1600: Referers section needs some love. We have some changes coming up in the next release (52). These are

- Deprecated in 52+
1601: `network.http.sendSecureXSiteReferrer`

- New in 52+ (currently in section 9999)
// 1600's: restrict the contents of referrers attached to cross-origin requests (FF52+)
   // 0- 1- 2-scheme+hostname+port
   // user_pref("network.http.referer.XOriginTrimmingPolicy", 2);
// 1600's: default referrer fallback override? (FF52+?)
   // 0-no-referer 1-same-origin 2-strict-origin-when-cross-origin
   // 3-no-referrer-when-downgrade (default)
   // https://bugzilla.mozilla.org/show_bug.cgi?id=1304623
   // user_pref("network.http.referer.userControlPolicy", 3);

I also came across this article released today: https://scotthelme.co.uk/a-new-security-header-referrer-policy/  which nicely describes what each type of referer does exactly. Might be a good ref for you @pyllyukko 

The 1600's section currently has only one active pref, and recommends using an add-on. In 52, it could be completely inactive - this doesn't seem right - surely some of these 7 prefs by default could be tightened up? And I think the section still needs better explanation (It's getting a bit confusing). I don;t mind losing all the numbering and starting from scratch, so it's logical. @fmarier 's opinion here would be great. Especially what the inactive defaults should be. Do we set them at FF defaults, or at most private or at a balance for less site breakage - and the new ones, what do we do with those: the first of which I have not filled in what 0 or 1 means, and I'd also like someone else to confirm what I already have is ok/right, including the descriptions.

publicarray:
The blog from Scott Helme talks about the [Referrer-Policy header](https://www.w3.org/TR/referrer-policy/). The header can be implemented by web developers on their site to control the referrer behavior on their domain. basically it belongs near the category of headers like the CSP (content security policy). I would assume that the setting conflict the option where the least amount is send wins. Note I have not read the standard but I suppose at the moment it's a bit of an undefined behavior where the browser vendors can decide what takes precedence. 

I would recommend an extension like [Smart Referer](http://github.com/meh/smart-referer) rather than setting the preferences in FF.

earthlng:
[...]

PR #40, opened 2017-03-03
network.http.sendSecureXSiteReferrer

earthlng:
move network.http.sendSecureXSiteReferrer to 9997 and enable a safe alternative for now until we re-work the Referrer-Section

Only merge this after FF52 stable has been released!

### DIFF
--- a/user.js
+++ b/user.js
@@ -654,14 +654,11 @@ user_pref("font.blacklist.underline_offset", "");
 user_pref("gfx.font_rendering.graphite.enabled", false);
 
 /*** 1600: HEADERS / REFERERS [SETUP]
-     Except for 1601 and 1602, these can all be best handled by an extension to block/spoof
+     Except for 1602, these can all be best handled by an extension to block/spoof
      all and then whitelist if needed, otherwise too much of the internet breaks.
      http://www.ghacks.net/2015/01/22/improve-online-privacy-by-controlling-referrer-information/
      #Required reading: https://feeding.cloud.geek.nz/posts/tweaking-referrer-for-privacy-in-firefox/ ***/
 user_pref("ghacks_user.js.parrot", "1600 syntax error: the parrot rests in peace!");
-/* 1601: disable referer from an SSL Website
- * to be deprecated in FF52+? - https://bugzilla.mozilla.org/show_bug.cgi?id=1308725 ***/
-user_pref("network.http.sendSecureXSiteReferrer", false);
 /* 1602: disable the DNT HTTP header (this is essentially USELESS and raises entropy)
  * This setting is under Options>Privacy>Tracking>Request that sites not track you
  * [NOTE] if you use NoScript MAKE SURE to set the pref noscript.doNotTrack.enabled to match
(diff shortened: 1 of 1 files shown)


crssi:
I would comment out `user_pref("network.http.sendSecureXSiteReferrer", false);` because it breaks stuff and will anyway be deprecated in FF52 acting FF52 as this would be set as true.

Roman-Nopantski:
@crssi Once 52 stable lands, and once it has been confirmed that it is deprecated, and once we do all the other things for 52 .. and when the version number changes to 52 .. that is when I will comment out the items already moved to the deprecated section.. we are still on 51

crssi:
@Roman-Nopantski I was wrong and you are true about deprecation... its not out until not confirmed.

But about sendSecureXSiteReferrer... I would still set this to true, since when a web page is written badly "half SSL / half non-encrypted" we have a breakage.
But of course its your call. I am just thinking out loud.

Everything about referers I would leave as default and set only `network.http.referer.XOriginPolicy`

`network.http.referer.XOriginPolicy = 2` effectively strips all cross domain referals where not full hostnames match.
But in for example you visit www.blahblah.com and you are redirected to passport.blahblah.com to signin where referal is needed, you will have a breakage.

`network.http.referer.XOriginPolicy = 1` effectively strips all cross domain referals where not base domain match.
So in this case the upper example will not have a problem.
There are some cases where this would not be OK... for example "blahblah **.co.uk**"

So with "2" we are covered everywhere and with "1" only a small subset is not covered.

When `sendSecureXSiteReferrer = false` and a web page is written badly "half SSL / half non-encrypted" we have a breakage.

Roman-Nopantski:
Stick to #5 where this is relevant. I can't win can I? Everyone wants something different. The only crap I care about is cross domain. The only settings not at default are the two XOrigin ones. You and earthlng go have a massive arm-wrestle and argument over it all. AFAIK what I already have proposed is just fine.

XOriginPolicy is NOT set at 2 .. it is set at 1 which is LESS breakage than when at 2 ... "some cases .. not ok" ..well, yes, I expect some breakage. It says it in the description

`sendSecureXSiteReferrer` is dead, well be commented out in a week.

"written badly "half SSL / half non-encrypted"" - well, for starters this is partially covered under active and passive XSS content. What am I supposed to do ... re-write the entire web?

see you at #5 .. am putting down a $5 on earthlng winning the arm wrestle, but both of you to be unable to make me make any changes :)

crssi:
Oh no. I am not trying to arm wrestle here, just commenting... nothing else.
Please, don't be mad on me. :)

fmarier:
> `network.http.referer.XOriginPolicy = 1` effectively strips all cross domain referals where not base domain match.
> There are some cases where this would not be OK... for example "blahblah .co.uk"

Actually, this case is handled correctly. If you look at [how the base domains are compared](https://searchfox.org/mozilla-central/rev/1bc7720a434af3c13b68b8d69f92078cae8890c4/netwerk/protocol/http/HttpBaseChannel.cpp#1665-1676), it is based on a list of [effective top-level domains](https://en.wikipedia.org/wiki/Public_Suffix_List). The *ETLDs* of `www.example.co.uk` and `www.example2.co.uk` are `example.co.uk` and `example2.co.uk` respectively. They are **not** both `co.uk`, which you would get if you used a simplistic algorithm that only removes the last suffix (`.uk`).

crssi:
Thats cool then and even better than I have anticipated :).
Thank you @fmarier.

PR #57, opened 2017-03-12
layout.spellcheckDefault

earthlng:
spellchecking in multi-line & single-line is better than multi-line only. Less typos, more better.
Both @atomGit and myself are using the value 2.

### DIFF
--- a/user.js
+++ b/user.js
@@ -1394,8 +1394,8 @@ user_pref("browser.ctrlTab.previews", true);
 /* 3011: don't open "page/selection source" in a tab. The window used instead is cleaner
  * and easier to use and move around (eg developers/multi-screen). ***/
 user_pref("view_source.tab", false);
-/* 3012: spellchecking: 0=none, 1-multi-line controls, 2=multi-line & single-line controls ***/
-user_pref("layout.spellcheckDefault", 1);
+/* 3012: spellchecking: 0=none, 1-multi-line controls (default), 2=multi-line & single-line controls ***/
+user_pref("layout.spellcheckDefault", 2);
 /* 3013: disable automatic "Work Offline" status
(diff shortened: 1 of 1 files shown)


atomGit:
again, no comment - just me personal pref since i can't spell worth a shit :)

Thorin-Oakenpants:
default is 1. I want to get more in line with atomGit's settings if they are in common, but this is really personal choice. I for one, do not want spellchecking in single lines, as these (usually, for me anyway) are usually info fields for names, towns, etc - as compared to writing sentences. Could be argued that people can update their dictionaries and add their common towns, names. Again, this is just my preference.

The whole personal settings section should probably all be inactive - I will be opening an issue over this at some stage. I think the 3200s can do with some love: add some more stuff if we go full on commented out, maybe a slight rejig of order, make defaults known, decide on what value to set as commented out.

I will close this and the other pull request when I set up the issue

earthlng:
I don't see the problem with 2 - I mean it's spellchecking not auto-correct.
We could help make the Internet a better place if all the reddit titles for example would have less spelling mistakes. I know - a bit cheesy right? - but you get my point.

Thorin-Oakenpants:
reddit titles without a typo would suck. I see a problem 1. its not the default so will probably be more likely to spring a change on users than not - but its not a harmful change 2. you might like it, but others may not (I find underlined wavy red lines highly annoying for names/towns, not titles) - .. but 3. moot point because I think I'm going to comment out all of 3200

Thorin-Oakenpants:
^^ that's I'm *thinking* about it, but would have a discussion first. Will do after ghacks article to get a broader feedback

earthlng:
ok, np, closed and deleted the branch

PR #58, opened 2017-03-12
security.dialog_enable_delay

earthlng:
set to 700ms for a slight speedup and to match @atomGit's user.js


atomGit:
i have no comment - that's just my personal pref

Issue #73, opened 2017-03-28
browser.privatebrowsing.autostart: means no "normal" mode right?


Thorin-Oakenpants:
If you start FF in "normal" mode, any new private windows will display a badge/icon in the top right (a mask image). (Side note: themes do not load either)

If you start FF in PB mode, there is no PB mode icon shown anywhere. If you then open a new "normal" window, there is no difference to that visually from the PB mode window (I have never started in PB mode so I would like to clarify this).

I have tested this with multiple vanilla FF52s. To me, this either a bit of a failure and/or ambiguous. @fmarier 

==
below are three concurrent profiles (vanilla FF52), all with `Options>Privacy>History>Firefox will use custom settings`, starting in "normal" mode (i.e "always use PB mode" is unchecked), with spawned PB windows (just reusing the image from the wiki to show the PB mode badge and loss of theme)

![](https://raw.githubusercontent.com/ghacksuserjs/ghacks-user.js/385aeaa54e082ccba39d5363372fa99db7577331/wikipiki/concurrent02.png)

And here is a `browser.privatebrowsing.autostart`->true. If you spawn a "new window" or a "new private window", they are indistinguishable

![untitled-2](https://cloud.githubusercontent.com/assets/16656956/24390314/a92acf3a-13e4-11e7-88e2-bbdc5c5efd01.png)

==
So am I missing something here?
- The wording in options is "*ALWAYS* use PB mode", which at face value means that if you start in PB you can only spawn PB mode windows - regardless of the command (new window or new private window).
- ^^ Is this correct? @fmarier 
- If this is true (start PB mode = ONLY PB mode windows), then why even allow the command "new window". This is misleading. It should be disabled IMO or (to keep the ctrl-n which people use) the icon changed to the same as "new private window" - whatever, also needs to cover context menus/menus/hamburger buttons/key-combos?. It needs to be clear.
- Also, why is the PB mode badge missing. This should, IMO, be enforced. See [this](https://support.mozilla.org/t5/Protect-your-privacy/Private-Browsing-Use-Firefox-without-saving-history/ta-p/4473) from mozilla support: "Private Browsing windows have a purple mask at the top", but also "When Firefox is set to Never remember history you won't see a purple mask at the top of each window, even though you are effectively in Private Browsing mode". Who dreamed that up?
- While people should be responsible for their own tweaks etc, what if an addon or script changed the start mode back to normal. An end-user wouldn't pick up on it.

[...]

PR #74, opened 2017-03-28
0001: extra info re starting in PB mode

Thorin-Oakenpants:
Because PB mode still allows the key-commands/menus/context menus/buttons to do both "new window" and "new private window", and because in this mode the PB mode purple mask icon is never shown, this is not clear. See issue #73


earthlng:
Doesn't need the comma I think but other than that I like it

earthlng:
FYI PK added this note recently: `Private browsing breaks Kerberos authentication`

Thorin-Oakenpants:
PB mode breaks a lot of stuff: btw, OT: did the concurrent profiles wiki - check it out. Will do implementation soon, and revamp readme (hopefully both today) - then it's article time for Martin. I want him to publish the article on Saturday - weekends are good

earthlng:
Ok, cool :)

Issue #115, opened 2017-05-13
ToDo: standardize descriptions [within reason]


Thorin-Oakenpants:
suggestion - go thru the entire js and streamline how we describe prefs - key word is the first word

We current have enable, disable, set, enforce, use, stop (see 0607), limit, force (see 0907 - could easily be enable) etc etc... disable/enable/set are perfect. Suggest we change as many of the others as possible to use either a standardized set of words where possible. I haven't gone thru the entire js yet.

-update-
OK, so I have 110 numbered items NOT starting with `enable` or `disable` or `set`. Exclude the 7 from 2697* about navigator objects, and exclude the 10 from 2698* about isolating. That leaves 93. I haven't actually read the list in full, just dumping it here.

```js
see PR Standardize Part1 & Part2
```

Thorin-Oakenpants:
OK, so I have 124 numbered items NOT starting with enable or disable. I will exclude the 7 from 2697* about navigator objects, and exclude the 10 from 2698* about isolating. That leaves 107.

I think "set" is a good one to keep, as is "sanitize". Some of these can easily use enable/disable. I haven't actually read the list in full, just dumping it here.

```
/* 0001: Start Firefox in PB (Private Browsing) mode
/* 0102: set start page (0=blank, 1=home, 2=last visited page, 3=resume previous session)
/* 0204: set APP locale ***/
/* 0207: set language to match ***/
/* 0208: enforce US English locale regardless of the system locale
/* 0301c: ensure update information is not suppressed ***/
/* 0330b: set unifiedIsOptIn to make sure telemetry respects OptIn choice and that telemetry
/* 0331: remove url of server telemetry pings are sent to ***/
/* 0335: remove telemetry clientID ***/
/* 0607: stop links launching Windows Store on Windows 8/8.1/10
/* 0803: display all parts of the url in the location bar - helps SECURITY ***/
/* 0804: limit history leaks via enumeration (PER TAB: back/forward) - PRIVACY
/* 0902: use a master password (recommended if you save passwords)
/* 0903: set how often Mozilla should ask for the master password
/* 0904: how often in minutes Mozilla should ask for the master password (see pref above)
/* 0906: ignore websites' autocomplete="off" (FF30+)
/* 0907: force warnings for logins on non-secure (non HTTPS) pages
/* 0908: When attempting to fix an entered URL (see 0802: browser.fixup.alternate.enabled),
[...]

PR #122, opened 2017-05-21
Standardize Part2 #115

Thorin-Oakenpants:


earthlng:
2604: why [MAC]? see [OS-diff](https://earthlng.github.io/FFprefs-diffs/Firefox-v53.0.html) - also exists in Win+Linux
2605: why [WINDOWS]? it's not in the OS diff
2669: strip is the correct word to use and is not the same as remove

Thorin-Oakenpants:
2604 read the ticket (I actually read about 7 of these damn things, most to do with thunderbird) - it is on Mac that a temp file is created and not removed, on the desktop (maybe @grauenwolfe can help) - the pref is only useful in Mac as far as i can tell

2605 - ohh, then if this applies to Mac and Linux then the wording of the pref is fuked

2669: "remove paths" is the same as "strip paths" especially to a layman. Yes strip is better, but some compromises have to be made. I'll think about it.

Thorin-Oakenpants:
2604: browser.helperApps.deleteTempFileOnExit - default in windows/linux is true - but on a MAC is false, but I'll remove the tag

2605: http://kb.mozillazine.org/Browser.download.manager.addToRecentDocs - I know this is old, but it says WINDOWS only. That page says it four times. But I will look further .. https://www.lifewire.com/modifying-download-settings-in-firefox-445716  .. says windows only ...  https://dxr.mozilla.org/mozilla-central/source/toolkit/components/jsdownloads/src/DownloadPlatform.cpp#110

>// On Windows and Gtk, add the download to the system's "recent documents" list, with a pref to disable.

So it's [WINDOWS] [ANDROID] then?

earthlng:
disable integration of activity into ~~windows~~ system's recent documents
or:
disable adding downloads to the system's "recent documents" list (to use the same words as mozilla)

Idk if GTK is Android too but it's definitely used in (some?) Linux and since MAC is based on Linux maybe there too. I think we could remove the OS tags. Everyone will know if their system has "recent documents" or not

PR #137, opened 2017-06-03
9999 revamp

Thorin-Oakenpants:
Nits, review syntax etc. Note: 2 items missing deprecation bugzilla tickets, we can get those in time. Note if each section number is made active, the prefs are also - except those which either match the current js (eg TP/SB not active but we do block reporting) or they make no sense or were inactive originally (eg personal 3000 settings etc) - might want to review those choices as well. Also, a few numbers etc changed to match current numbers (eg replaced by items etc, new sections)

### DIFF
--- a/user.js
+++ b/user.js
@@ -1600,179 +1600,230 @@ user_pref("browser.urlbar.decodeURLsOnCopy", true);
 /* END: internal custom pref to test for syntax errors ***/
 user_pref("ghacks_user.js.parrot", "No no he's not dead, he's, he's restin'! Remarkable bird, the Norwegian Blue");
 
-/*** 9999: DEPRECATED / REMOVED
-     Confirmed by resetting as well as via documentation, bugzilla tickets, and DXR searches.
-     [NOTE] Numbers may get re-used ***/
-/* 2607: (23+) disable page thumbnails, it was around v23, not 100% sure when
- * this pref was replaced with browser.pagethumbnails.capturing_disabled ***/
-   // user_pref("pageThumbs.enabled", false);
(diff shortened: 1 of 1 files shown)


earthlng:
* 2502 - newline for `Initially a Linux issue ...`
* 1266 - [1] should be a [-] and [2]=>[1]
* 0308 - maybe change title: disable ~~update~~ plugin **update** notifications
* 1005 - maybe rename to 1021b and also copy the extra description from 1021
* security.tls.insecure_fallback_hosts.use_static_list - we can remove the first link because all they did there was remove 1 #ifdef; the 2nd link removed the pref and all the code (from the sticky aswell!)
* 9999 - move the ***/ on a new line ? add RENAMED to the title ?
* 0415 - remove FF from the bracket => (41+) and maybe rename title to `disable **SafeBrowsing** reporting URLs`
* dom.disable_image_src_set - removed the [TEST] and [WARNING] ?
* 0806: can activate the pref IMO
* mozTCPSocket ??
* network.http.spdy.enabled.http2draft - was removed in FF41, see my last comment in the sticky

I didn't find any syntax errors. Good job! Thanks!

Thorin-Oakenpants:
Will check it all out when i'm sober :) Good spotting for a newbie

Thorin-Oakenpants:
nits patched in https://github.com/ghacksuserjs/ghacks-user.js/commit/ed9bc336fa8d9643c30fe1c716e2b0b0163f452d

All changes also reflected in sticky issue first post

Only nit not changed is `2502 - newline for `Initially a Linux issue..`,  Trying to keep lines to a minimum

PR #146, opened 2017-06-17
Convert line endings to LF

### LINKED ISSUE #145: Why does user.js have Windows line endings?
This is a bit inconvenient. When I modify the file locally on a Linux machine, I get mixed line endings (LF on the lines I modified, CRLF on the lines I did not modify). Or, when I configure my editor to input CRLF line endings, Git starts to throw a hissy fit when I want to commit my changes.

[This article](https://help.github.com/articles/dealing-with-line-endings/) is a good guideline on how to deal with line endings in Git (basically, LF line endings in the repo, native line endings in the working directory). But `user.js` is CRLF in the repo and CRLF in the working directory.

Can this be fixed? A single `dos2unix user.js` should do it, but that would (1) cause a massive commit that would break merges from/to previous revisions and (2) not solve the problem if the thing that _caused_ the file to have CRLF file endings is not resolved (probably a configuration error on the main contributor's machine?).

Thanks!

carmenbianca:
Fixes #145

Proposed solution. It's a bit nuclear as described in the issue, but it works.


Gitoffthelawn:
What was CRLF messing up?

carmenbianca:
The only downside that comes to mind is that it makes merges with past revisions a headache. Otherwise not really.

PR #182, opened 2017-07-28
0515+0516 - screenshots already bundled with FF54

earthlng:
moved screenshots up to 0515 and added the FF54 pref. I know I said we can remove that pref but the item needs to be for FF54+ regardless. To make it less confusing I added the pref back in.


Thorin-Oakenpants:
All we need to do is this: we are currently on v55

/* FF55
// 0516: disable Screenshots (FF54+) - replaced by extensions.screenshots.disabled
   // user_pref("extensions.screenshots.system-disabled", true); // (FF54+)
***/

PR #190, opened 2017-08-07
FF55+ deprecation

Thorin-Oakenpants:
- should check if 0360 `user_pref("browser.newtabpage.directory.source", "data:text/plain,");` is still around since the ping pref is gone.
- A bit iffy about 2507 - this spilts two prefs and there's a lot of text. Not sure if FF38+ refers to the second one. We should investigate the still active 2507 and fix that up with some info and version


earthlng:
> should check if 0360 `browser.newtabpage.directory.source` is still around since the ping pref is gone

Yes it's still around in [nightly](https://dxr.mozilla.org/mozilla-central/search?q=browser.newtabpage.directory.)

Thorin-Oakenpants:
`dom.keyboardevent.dispatch_during_composition` = false anyway, just ditch it = :+1: 

Sounds good to me. Will do as a separate commit

PR #192, opened 2017-08-08
9999 - JS-comment re-grouping for ESR users

earthlng:
makes it even easier for ESR users to re-enable the prefs they still need


Thorin-Oakenpants:
!bump @earthlng Did you not see my comments about this?

earthlng:
Modify it as you see fit.

Issue #264, opened 2017-11-09
override.js file no longer overriding user.js file in FF58+


overdodactyl:
edit: [1413413](https://bugzilla.mozilla.org/show_bug.cgi?id=1413413) is the cause - Thorin

edit II:  I submitted a bug report report ([1416279](https://bugzilla.mozilla.org/show_bug.cgi?id=1416279)) to hopefully address this - overdodactyl

edit III: bug was labeled WONTFIX

>The code that implements this functionality has been removed. After sending an intent to remove email to dev.platform and discussing, we decided that it wasn't worth the effort of continuing to support as we re-implement the preference service.

edit IV: [wiki link](https://github.com/ghacksuserjs/ghacks-user.js/wiki/1.5-Keeping-Up-To-Date)

---
Greetings,

In the past, I've been using an `override.js` file to override some preferences from `user.js` as suggested in [Section 1.5](https://github.com/ghacksuserjs/ghacks-user.js/wiki/1.5-Keeping-Up-To-Date) of the wiki.  

Sometime within the past week (maybe two?), the settings in my override.js file are no longer being used to override the user.js file in Firefox Nightly.  

I can reproduce this issue in a new Nightly profile, however, if I copy my user.js file and my preferences folder containing override.js to a new profile and use the stable version of Firefox, I do not experience this issue.

Has something changed? Can anyone else confirm they are experiencing this in Nightly (or, alternatively, that they aren't having the problem)?

I apologize if this has already been addressed...I did a quick search here and on bugzilla and couldn't find anything relevant .

Thorin-Oakenpants:
That sucks. I haven't got 58 here to test with to confirm (not that I doubt you!). Wouldn't surprise me if some tweak to shave a ms eg lazy loading of prefs caused this. Or someone removed it in their line of work as it's undocumented.

Might be worth looking at all the bugzillas resolved for 58

overdodactyl:
Thanks for the reply @Thorin-Oakenpants, I'll take a closer look at bugzilla and report back with anything I find!

Hopefully it wasn't removed on purpose...I've been relying on that :)

Thorin-Oakenpants:
[...]

PR #286, opened 2017-11-26
Updater for Windows v3.0

claustromaniac:
As per #264,

### _(Totally unnecessary)_ changes:

- Fixed label clash with obsolete **Break** command. _(Who cares?)_
- Smarter and case-insensitive version checks.
- Case-insensitive parameter checks.
- Overall syntax improvements and fixes, for consistency and to prevent circumstancial weird behaviour. Plus etiquette. _(except for code comments, I suck at that)_
- Some cleanup of unnecesary lines. _(Don't get too excited, the file is still way larger than before... like.. 5 KB larger or so)_
- By default the updater now attempts to auto update itself on execution. This new behaviour can be skipped with the new `-SkipBatchUpdate` switch. _~~(So that I can slip malicious code into your computer without you knowing. Oops.)~~ (This will help you get the fixes for all of the bugs possibly introduced in this useless update)_
- Added new `-LogP` switch, that works just like `-Log`, but also opens the log file after updating. _(again, who cares?)_
- Added new `-MultiOverrides` switch, that uses any and all .js files in a `user.js-overrides` sub-folder as overrides instead of the default `user-overrides.js` file. Files are appended in alphabetical order.
- Added new `-Merge` switch, that uses a new algorithm to merge overrides instead of appending them. Files are parsed in alphabetical order and when there are conflicting records, the values of the last ones declared are the ones used. The order of said records doesn't change in the resulting file. Comments are appended normally. When the script is run with both `-Merge` and `-MultiOverrides` switches, an `user-overrides-merged.js` file is also generated in the root directory for quick reference. Said file contains only the merged data from override files and can be safely discarded after updating.
- Probably some other minor stuff I can't remember right now.

### DIFF
--- a/updater.bat
+++ b/updater.bat
@@ -3,54 +3,104 @@ TITLE ghacks user.js updater
 
 REM ### ghacks-user.js updater for Windows
 REM ## author: @claustromaniac
-REM ## version: 2.1
+REM ## version: 3.0
 
+SET _myname=%~n0
+SET _myparams=%*
 SETLOCAL EnableDelayedExpansion
-SET "_ua="
-SET "_log="
 :parse
-IF "%~1"=="" GOTO endparse
-IF "%~1"=="-unattended" SET "_ua=true"
-IF "%~1"=="-log" SET "_log=true"
+IF "%~1"=="" (
+	GOTO endparse
+)
+IF /I "%~1"=="-unattended" (
+	SET _ua=1
+)
+IF /I "%~1"=="-log" (
+	SET _log=1
+)
+IF /I "%~1"=="-logp" (
+	SET _log=1
+	SET _logp=1
+)
+IF /I "%~1"=="-multioverrides" (
+	SET _multi=1
+)
+IF /I "%~1"=="-merge" (
+	SET _merge=1
+)
+IF /I "%~1"=="-skipbatchupdate" (
+	SET _skipbu=1
+)
 SHIFT
 GOTO parse
 :endparse
-SET "_name="
-SET "_date="
-SET "_version="
-SET /A "_line=0"
 ECHO.
-IF EXIST user.js (
-	FOR /F "delims=" %%i IN (user.js) DO (
-		IF !_line! EQU 1 SET "_name=%%i"
-		IF !_line! EQU 2 SET "_date=%%i"
-		IF !_line! EQU 3 SET "_version=%%i"
-		SET /A "_line+=1"
-		IF !_line! GEQ 4 GOTO break
+IF NOT DEFINED _skipbu (
+	ECHO Checking updater version...
+	ECHO.
+	IF EXIST "!_myname!-updated.bat" (
+		DEL /F "!_myname!-updated.bat"
 	)
-	:break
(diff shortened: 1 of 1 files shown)


earthlng:
Holy shit dude, you couldn't just pass the salt, could you? xD

Ok so for starters I won't merge this with the Opt-out batch update. Downloading an executable and running it, should not be the default behavior. Running the script without any parameters should just update the user.js and nothing more.

I haven't done DOS scripts in a while and need some help with this ...
- The double `%` in `%%G` is necessary because it's used inside a script, right?
- how does this work? `IF /I NOT "!_name!"=="!_name:ghacks=X!"` - the _name**:ghacks=X** part ?

earthlng:
I committed the working parts of your PR. Things that aren't working in your v3 version:
- the "Handling backup" part - I kept the old style which works and took the new formatting
- the merge part is broken. I used multi and merge and ended up with a user.js with all active user_pref() lines stripped and replaced with either `-------------- TEMP2` or `----------- USER-OVERRIDES`
- _logp is unable to open the file because it's locked by the script. it prints the error message to the logfile.

merging also takes a very long time. What's the reason that you do it twice when multi+merge are used? Wouldn't it be better to create a concat with all files + the user.js and then do the processing?

It also strips empty lines and leading whitespace on all lines which destroys our formatting.
Not that it matters when I end up with a user.js completely void of all user_pref lines anyway ;)

earthlng:
I'm closing this because it now conflicts with the latest version and I have no idea how "resolving conflicts" works. New PRs are welcome of course. Sorry for not taking everything and thanks for the parts that work.

claustromaniac:
Wow. What the hell did I do?

The non-working parts are definitely due to last-minute changes. The batch interpreter is so unpredictable. I'll fix those ASAP, sorry.
> 
> The double % in %%G is necessary because it's used inside a script, right?

Right.

> how does this work? IF /I NOT "!_name!"=="!_name:ghacks=X!" - the :ghacks=X part ?

`!_name!"=="!_name:ghacks=X!` means replace any instance of 'ghacks in _name with X' it's a fast way to check if the string includes that substring.

> Downloading an executable and running it, should not be the default behavior.

I assumed it was OK to do it this way since the whole using a script thingy is not really default behaviour at all. Also, I can't merge commits by myself (thank gods), so no risk there. I admit it's prone to bring about chaos so I'll change that.

> merging also takes a very long time. What's the reason that you do it twice when multi+merge are used? Wouldn't it be better to create a concat with all files + the user.js and then do the processing?

That's actually how it works (or how it would work if I didn't mess up before uploading my changes). 

I repeated a few lines of code to account for the different scenarios, using different combinations of settings, but they weren't that many lines considering how long the script already was. 

Merging takes long because the script builds a new user.js file from scratch and has to make a lot of comparisons, and there are two consecutive chunks of code that look the same, but each builds a separate file: `user-overrides-merged.js` and `user.js`. Building the first one doesn't take that long unless you have like lots of lots of overrides. Should I remove that part? or maybe make it optional? lol

claustromaniac:
> EDIT: I also noticed that it can't handle user instead of user_pref lines from user-overrides in the merging part. But the problems above happened after I removed my override.js files with user lines.

That was an oversight. An easily fixable one, thankfully.

PR #287, opened 2017-11-26
Updater for Windows v3.0 (the real deal)

claustromaniac:
- Added `-updatebatch` switch so the updater can auto update itself on execution. This is the only case-sensitive switch.
- Added new `-LogP` switch, that works just like -Log, but also opens the log file after updating.
- Added new `-MultiOverrides` switch, that uses any and all .js files in a 'user.js-overrides' sub-folder as overrides instead of the default 'user-overrides.js' file. Files are appended in alphabetical order.
- Added new `-Merge` switch, that uses a new algorithm to merge overrides instead of appending them. Files are parsed in alphabetical order and when there are conflicting records, the values of the last ones declared are the ones used. The order of said records doesn't change in the resulting file. Comments are appended normally. When the script is run with both `-Merge` and `-MultiOverrides` switches, an 'user-overrides-merged.js' file is also generated in the root directory for quick reference. Said file contains only the merged data from override files and can be safely discarded after updating.

For a more full list of changes, with jokes and nasty surprises, see my epic fail (#286).

@earthlng have fun testing 😆

### DIFF
--- a/updater.bat
+++ b/updater.bat
@@ -3,7 +3,7 @@ TITLE ghacks user.js updater
 
 REM ### ghacks-user.js updater for Windows
 REM ## author: @claustromaniac
-REM ## version: 3.0-alpha92
+REM ## version: 3.0
 
 SET _myname=%~n0
 SET _myparams=%*
@@ -59,7 +59,6 @@ IF DEFINED _updateb (
 		EXIT /B
 	)
 )
-GOTO begin
 :begin
 SET /A "_line=0"
 IF NOT EXIST user.js (
@@ -231,7 +230,5 @@ FOR /F "tokens=1,* delims=]" %%G IN ('find /n /v "" ^< "%~1"') DO (
 )
 ENDLOCAL
 DEL /F %~1 temp123 >nul
-GOTO EOF
+GOTO :EOF
 REM ############################
-
-:EOF
--- a/updater.bat
+++ b/updater.bat
@@ -65,7 +65,7 @@ DEL /F %2 2>nul
 FOR /F "tokens=* delims=" %%G IN (%1) DO (
 	SET _pref=%%G
 	SET "_temp=!_pref: =!"
-	IF /I "user.js.parrot"=="!_temp:user.js.parrot=!"
+	IF /I "!_temp!"=="!_temp:user.js.parrot=!" (
 		IF /I "user"=="!_temp:~0,4!" (
 			FOR /F "delims=," %%S IN ("!_pref!") DO (
 				SET _pref=%%S
--- a/updater.bat
+++ b/updater.bat
@@ -3,7 +3,7 @@ TITLE ghacks user.js updater
 
 REM ### ghacks-user.js updater for Windows
 REM ## author: @claustromaniac
-REM ## version: 2.2
+REM ## version: 3.0
 
 SET _myname=%~n0
 SET _myparams=%*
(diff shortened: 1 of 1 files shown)


earthlng:
Duuuuuude, chillax mate xD Let me commit the stuff I'm working on and then see what you think.

earthlng:
Wow, it's so much cleaner now, isn't it! I love it. ❤️ 💋 
It still deletes empty lines but that's no biggie. It also doesn't seem to like exclamation marks in the final merged user.js and removes them. Any idea if that can be fixed?
I'm side-by-side comparing your latest commit and the latest version I just committed...
* we could delete the "inputfile" at the end of the merge function ... 👍 👎 ?
* does it need to be `$1` or `$~1` in the function? do you know? what's the difference? what's better?
* `>nul 2>&1` or `2>nul` in the `DEL` lines? do we need to suppress the standard-out?
* can we do `REN "!_myname!.bat" "!_myname:-updated=!.bat"` after `DEL /F "!_myname!.bat" >nul 2>&1` and before the exit? It looks weird atm

claustromaniac:
> It still deletes empty lines but that's no biggie. 

My latest version didn't delete empty lines last time I checked. I simply added "delims=" to the FOR call and it stopped messing with spaces.

> It also doesn't seem to like exclamation marks in the final merged user.js and removes them. Any idea if that can be fixed?

I didn't notice that, I'll check it out.

> we could delete the "inputfile" at the end of the merge function ... 👍 👎 ?

👍 

> does it need to be $1 or $~1 in the function? do you know? what's the difference? what's better?

%~1 is the same, only that it removes quotes. They should be interchangeable in this context since the  names are always the same.
>nul 2>&1 or 2>nul in the DEL lines? do we need to suppress the standard-out?

No need to. It's a matter of preference. I don't even remember if DEL has any standard out at all and I didn't bother checking lol.

> can we do REN "!_myname!.bat" "!_myname:-updated=!.bat" after DEL /F "!_myname!.bat" >nul 2>&1 and before the exit? It looks weird atm

Nope, the preceding line deletes the currently running batch script, nothing can go after that. In fact I put the `EXIT /B` there mostly as a reminder, it shouldn't even be necessary. The renaming is for the updated batch (a new instance) so it can go back to being just updater.bat after it finishes.

claustromaniac:
So... I found the cause for the missing exclamation marks, but it is a genuine pain in the butt to fix. I will look into it in the next few days if I can, but no promises.

Unfortunately there seems to be no way to fix the missing empty lines within a pure batch solution, though. It's just the way FOR loops work.

earthlng:
> Nope, the preceding line deletes the currently running batch script, nothing can go after that.

Oh okay, thanks, I didn't know that or rather I probably forgot about it over the years. So that means it reads and executes the script line by line then, huh? Well I guess I shouldn't save changes while the script is still running in the background ... good to know.

> So... I found the cause for the missing exclamation marks, but it is a genuine pain in the butt to fix. I will look into it in the next few days if I can, but no promises.

Don't worry I already got it, at least for the most part. And I have a native fix for the empty lines as well.

As for the -updatebatch part - I think we still need to strip it from the params, no? Otherwise it would result in an endless loop I think, right? I haven't tested that part yet at all.

I will do a commit and then it would be nice if you could adjust your local copy you're working on with my changes that you're happy with (like are you okay with having the function at the end, the function name and all these other small details where we still have slight differences) and then we can figure out the rest.

claustromaniac:
Hey, thanks for the ingenious way to deal with spaces 👍

This is finally looking pretty good!

Thorin-Oakenpants:
I have a question .. when you check for the current version and online version - do you detect the "-beta" parts - i.e you could grab line 4, char 13
- if char 13 is a hyphen `-` then use the string from chars 11-17 `xx-beta` else use the string from chars 11-12 `xx`

claustromaniac:
@Thorin-Oakenpants As it is now it detects everything after the first space (the first 2 characters in that line) up to the colon.

claustromaniac:
@Thorin-Oakenpants I read that in a hurry earlier and didn't notice this part:

> when you check for the current version and **online** version

I didn't bother with number comparisons because there are normally multiple commits between version numbers (so I really want to update multiple times for each version # of the file), and to even check the online version number it has to download the file first anyway. 

The only comparison this makes is line-per-line between the backup and the new file, to see if they are the same or not.

I think there's still room for improvements, especially in that regard. Even with all the added console output it is still kind of a shy script. So if you have any suggestions, just shoot.

claustromaniac:
@earthlng It took me a while to notice your comment here [1c99769 ](https://github.com/ghacksuserjs/ghacks-user.js/commit/1c997693ffef7d4e3610fa6e0038b113409e6e63). 

Github is already making me dizzy.

earthlng:
damn that trick to rebase the PR did not work. I'm closing this and will create a new PR with your latest file so that it only shows the parts that are different.

PR #292, opened 2017-11-28
Updater for Windows - misc fixes and improvements

claustromaniac:
FIxed `-updatebatch` switch, which is now case-insensitive too.

Other fixes below. Don't be lazy and just read the commit notes.

There's more to come, but you can merge this in the meantime if you want (or ditch it). Up to you.

@earthlng Thanks again for your help testing and ironing out my initial mess, and for the fancy solution to the missing empty lines! 

I often manage to pull off relatively advanced stuff like I did here with the merge function but I'm really a newbie when it comes to this language, and scripting/programming in general. I guess I'm kinda crazy for even trying lol. I feel like a mad scientist tinkering with stuff I don't fully understand.

### DIFF
--- a/updater.bat
+++ b/updater.bat
@@ -56,9 +56,10 @@ IF DEFINED _updateb (
 		)
 	) ELSE (
 		IF EXIST "!_myname:~9!.bat" (
-			DEL /F "!_myname:~9!.bat"
+			REN "!_myname:~9!.bat" "!_myname:~9!.old"
 			CALL :begin
 			REN "!_myname!.bat" "!_myname:~9!.bat"
+			DEL /F "!_myname:~9!.old"
 			EXIT /B
 		) ELSE (
 			ECHO.
--- a/updater.bat
+++ b/updater.bat
@@ -207,38 +207,40 @@ EXIT /B
 
 REM ###### Merge function ######
 :merge
-DEL /F %2 2>nul
 SETLOCAL disabledelayedexpansion
-FOR /F "tokens=1,* delims=]" %%G IN ('find /n /v "" ^< "%~1"') DO (
-	SET "_pref=%%H"
-	SETLOCAL enabledelayedexpansion
-	SET "_temp=!_pref: =!"
-	IF /I "user_pref"=="!_temp:~0,9!" (
-		IF /I NOT "user.js.parrot"=="!_temp:~12,14!" (
-			FOR /F "delims=," %%S IN ("!_pref!") DO (
-				SET "_pref=%%S"
-			)
-			SET _pref=!_pref:"=""!
-			FIND /I "!_pref!" %~2 >nul 2>&1
-			IF ERRORLEVEL 1 (
-				FOR /F "tokens=* delims=" %%X IN ('FIND /I "!_pref!" %~1') DO (
-					SET "_temp=%%X"
-					SET "_temp=!_temp: =!"
-					IF /I "user_pref"=="!_temp:~0,9!" (
-						SET "_pref=%%X"
-					)
+(
+	FOR /F "tokens=1,* delims=]" %%G IN ('find /n /v "" ^< "%~1"') DO (
+		SET "_pref=%%H"
+		SETLOCAL enabledelayedexpansion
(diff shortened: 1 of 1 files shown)


earthlng:
Thanks for the new PR. I didn't have the time or the nerve yesterday to deal with the buggy -updatebatch part so I just removed it again.
I like most of the changes here except the PING part, it triggered my firewall and I'd prefer to use TIMEOUT. We also need to change 2 of the EXITs to GOTO :EOF because they prevent the code from reaching the renaming part. The exit at the end and the one at `IF ERRORLEVEL 2 EXIT /B`.

The PAUSE here also comes before the renaming part which prevents the renaming if you close the window at that point:
```
IF NOT DEFINED _log (
	IF NOT DEFINED _ua PAUSE
)
```
With these changes the autoupdating part seems to work.
But I also noticed a more serious problem with the merging - it swallows user_pref lines from overrides that are commented out in the ghacks user.js?!

claustromaniac:
> I like most of the changes here except the PING part, it triggered my firewall and I'd prefer to use TIMEOUT.

127.0.0.1 is the localhost address, which means your firewall is either dumb or super nosy 😆 It's ok, though. Let's change that.

> We also need to change 2 of the EXITs to GOTO :EOF because they prevent the code from reaching the renaming part. The exit at the end and the one at IF ERRORLEVEL 2 EXIT /B.

The only difference between EXIT /B and GOTO :EOF is that the former can modify the ERRORLEVEL variable. Otherwise they are the same.

I'll see what I can do about the final pause. Good catch.

> But I also noticed a more serious problem with the merging - it swallows user_pref lines from overrides that are commented out in the ghacks user.js?!

It shouldn't, but maybe I'm not understanding what you mean correctly - I'll have a look.

earthlng:
Hey don't you insult my firewall mate! It's exactly the way I like it. :)

Are you sure about the EXIT? AFAIK exit kills the script and going to :EOF will return after a CALL, no?

earthlng:
try with something like `user_pref("browser.privatebrowsing.autostart", true);` as an override and it will simply not end up in the final user.js at all

claustromaniac:
@earthing I found two separate reasons for it to happen. One is my latest commit, but even after reverting it I still will have to add some code to prevent it. Nice.

> Are you sure about the EXIT? AFAIK exit kills the script and going to :EOF will return after a CALL, no?

Yes, EXIT does that, but EXIT /B exits just the script (or subroutine), which is the same GOTO :EOF does.

earthlng:
no need to revert that part IMO. To debug you can use `ECHO DEBUG .... >>debug.txt`

claustromaniac:
No, I really did have to revert it. The function relies on the target file being generated on the go to check if it already merged a specific preference or not, otherwise it's works as a terribly inefficient append function lol. But even after removing it, the FIND call that I use to make that specific check is unreliable because it counts commented-out preferences.

earthlng:
> Yes, EXIT does that, but EXIT /B exits just the script (or subroutine), which is the same GOTO :EOF does.

Thanks, I did not know that.

TIMEOUT 300 - really??? well whatever, I've bitched enough already I'll just take it :)
Thanks for the quick fix for the merging problem!

So I figured out what the problem was with the auto-updater part failing to rename the file on my system.
There's this thing called Application Compatibility Engine on Vista + Windows 7 which somehow keeps a handle on executables and that leads to delayed deletions. I've just disabled it and now everything works as expected. But not everyone may want to disable ACE and an easy workaround in the code is to first rename the bat file to .old (or something non-executable) and delete it after the final renaming happened.

earthlng:
Congrats on the first PR not getting rejected xD No seriously thanks a lot!
How do you feel about making a nice little header for the script. Add some ASCII art if you want or whatever. I think it really deserves *something*. It should probably go into the `IF NOT DEFINED _ua (...` part. ... Or not, up to you.

earthlng:
I stand corrected

> Are you willing to run more tests? xD

anyone?? please?? no? ok then, keep 'em coming I guess, I'll take one for the team

PR #298, opened 2017-12-01
Updater for Windows 3.1

claustromaniac:
Just some more polishing and some changes to prevent some potential issues when the user closes the console window too soon.

I still need @earthlng to test this because the `-updatebatch` switch didn't work for him before and it may be so again.

### DIFF
--- a/updater.bat
+++ b/updater.bat
@@ -36,13 +36,13 @@ GOTO parse
 :endparse
 ECHO.
 IF DEFINED _updateb (
-	REM THe normal flow here goes from phase 1 to phase 2 and then phase 3.
+	REM The normal flow here goes from phase 1 to phase 2 and then phase 3.
 	IF NOT "!_myname:~0,9!"=="[updated]" (
-		REM Phase 3
-		REM The new script, with the original name, should:
-		REM 	Delete the [updated]*.bat script
-		REM 	Begin the normal script routine.
 		IF EXIST "[updated]!_myname!.bat" (
+			REM Phase 3
+			REM The new script, with the original name, should:
+			REM 	Delete the [updated]*.bat script
+			REM 	Begin the normal script routine.
 			REN [updated]!_myname!.bat [updated]!_myname!.bat.old
 			DEL /F "[updated]!_myname!.bat.old"
 			ECHO Script updated^^!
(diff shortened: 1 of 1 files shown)


earthlng:
> do you think I should still rename the scripts to .old or something before overwriting/deleting?

Yes. And I think adding the "how it works" you wrote [here](https://github.com/ghacksuserjs/ghacks-user.js/pull/298/commits/00e17281dc439c1e517671eeeb5bfc4e96f90104) at the top of the `IF DEFINED _updateb (` block would help a lot with understanding that part of the code.
---
instead of this:
```
IF EXIST user.js.old.bak ( DEL /F user.js.old.bak )
IF EXIST user.js (
	IF EXIST user.js.bak ( REN user.js.bak user.js.old.bak )
```
how about this? :
```
IF EXIST user.js (
	IF EXIST user.js.bak ( MOVE /Y user.js.bak user.js.old.bak )
```

---
re: carets - this
```js
user_pref("caret.x3x3", "!! ^^^true^^^");
```
gets merged to this:
```js
user_pref("caret.x3x3", " ^true^");
```

Thorin-Oakenpants:
@earthlng Is it worthwhile adding a version/update date to both files

earthlng:
only the version is enough

claustromaniac:
> how about this? : 

The user could have bak files and no user.js file.

Regarding carets, I didn't test exhaustively but it only happens when exclamation marks are also involved, right?

earthlng:
> The user could have bak files and no user.js file.

exactly, and you're deleting it even if there's no need to.

> it only happens when exclamation marks are also involved, right?

yes, but theoretically a pref that requires such a value could exist and at least nobody can blame us for not mentioning that merging could mess with that.

claustromaniac:
> exactly, and you're deleting it even if there's no need to.

The script is never meant to leave a .bak.old backup file behind. If it did, the most likely reason is that the user closed it too soon. Deleting that file in that situation means doing what it would've done before.

> yes, but theoretically a pref that requires such a value could exist and at least nobody can blame us for not mentioning that merging could mess with that.

The point is that exclamation marks are still the real cause for issues, not carets.

earthlng:
1. Oh okay, good point. Let's keep it as it is then.
2. that's true. Ok, I'll remove it from the howto.

claustromaniac:
1. On second thought, the script will still delete that file later on in the code, and it will compare the files first, so it would be handled better that way. I will adopt the change you suggested.

earthlng:
Lol, was just about to merge this but if you think it's better this way I'll take it. Anything else before I merge your 3rd non-rejected PR?? :)
The "**THe** normal flow here" maybe? last chance xD

earthlng:
> New changes since you last viewed
> 3.1 ccaf069
> New changes since you last viewed
> Revert "3.1" a6c93b3 

looool what is happening?

claustromaniac:
There you go. lol

earthlng:
taking it in 3 ... 2 ... 1 ... Speak Now or Forever Hold Your Peace ... 0.9 .... 0.8 ....

earthlng:
3 out of 5 - congrats! but more importantly - THANKS! good stuff

Issue #303, opened 2017-12-03
Script to clean deprecated prefs from prefs.js


crssi:
Anyone interested in that kind of script, before I dig into?
If so, I can make powershell one.

Cheers

Thorin-Oakenpants:
See #242 which I hope to do one day

crssi:
It a bit ugly. But WTH, will be good for first post.

```powershell
if (Get-Process firefox -ErrorAction SilentlyContinue) { Write-Host 'Firefox must be closed to proceed. Exiting...'; Exit }

if (($InstalledVersion = Get-ItemProperty 'HKLM:\Software\Microsoft\Windows\CurrentVersion\Uninstall\Mozilla Firefox *', 'HKLM:\SOFTWARE\Wow6432Node\Microsoft\Windows\CurrentVersion\Uninstall\Mozilla Firefox *' | Sort-Object DisplayName | Select-Object @{Name='Version';Expression={$_.DisplayVersion.Split('.')[0]}}, @{Name='Name'; Expression={$_.DisplayName}}) -eq $null) { Write-Host 'No Firefox installed. Exiting...'; Exit }

Try { $ghacks_user_js = (Invoke-WebRequest -Uri https://raw.githubusercontent.com/ghacksuserjs/ghacks-user.js/master/user.js).Content } Catch { Write-Host 'There might be no internet connection. Exiting...'; Exit }

$Section9999 = ([regex]'\/\*\*\*\s9999:.*\/\*\sEND:').Match($ghacks_user_js.Replace("`n", "°")).Value # get deprecated section
$Section9999 = ([regex]'°\/\*\sFF.*\/\*').Match($Section9999).Value # strip header and kill the parrot
$ValidVersions = @(($Section9999.Split('°') | Select-String -Pattern '^\/\S\sFF.*') | ForEach { $_ -match '^\/\S\sFF(\d{2}).*' | Out-Null; $Matches[1] })


Write-Host 'Firefox installation(s) found:'
$InstalledVersion | ForEach { "$($_.Version) : $($_.Name)" }
Write-Host ''

if ($InstalledVersion.Count -ne 1) { $Version = Read-Host -Prompt "Enter Firefox version (Default: $($InstalledVersion[0].Version); 0: Exit)" }
if ($Version -eq 0) { Exit }
if (!$Version) { $Version = $InstalledVersion[0].Version }
[...]

PR #312, opened 2017-12-13
prefs.js cleaner script for Windows 1.0b2

claustromaniac:
#303 

First actual working version.

If my research and tests bear fruit I'll have some nice changes to make to both this script and the updater later.

### DIFF
--- a/prefs cleaner.bat
+++ b/prefs cleaner.bat
@@ -9,7 +9,7 @@ SETLOCAL EnableDelayedExpansion
 :begin
 CLS
 CALL :message "This batch should be run from your Firefox profile directory."
-CALL :message "It will remove from any entries from prefs.js that also exist in user.js."
+CALL :message "It will remove any entries from prefs.js that also exist in user.js."
 CALL :message "This will allow inactive preferences to be reset to their default values."
 CALL :message "Firefox needs to stay closed during the process."
 ECHO:
@@ -83,28 +83,18 @@ REM ########### Help Function ###########
 CLS
 CALL :message "This script creates a backup of your prefs.js file before doing anything." 
 CALL :message "It should be safe, but you can follow these steps if something goes wrong:"
-ECHO   1. Make sure Firefox is closed.
-ECHO:
+CALL :message "  1. Make sure Firefox is closed."
 ECHO   2. Delete prefs.js in your profile folder.
-ECHO:
-ECHO   3. Delete Invalidprefs.js if you have one in the same folder.
-ECHO:
+CALL :message "  3. Delete Invalidprefs.js if you have one in the same folder."
 ECHO   4. Rename or copy your latest backup to prefs.js.
-ECHO:
-ECHO   5. Run Firefox and see if you notice anything wrong with it.
-ECHO:
+CALL :message "  5. Run Firefox and see if you notice anything wrong with it."
 ECHO   6. If you do, restart it again, and check back.
-ECHO:
-ECHO   7. If you still notice something wrong, especially with your extensions,
-ECHO:
+CALL :message "  7. If you still notice something wrong, especially with your extensions,"
 ECHO      and/or with the UI, go to about:support, and restart Firefox with
-ECHO:
-ECHO      add-ons disabled. Then, restart it again normally, and see if the problems
-ECHO:
+CALL :message "     add-ons disabled. Then, restart it again normally, and see if the problems"
 ECHO      were solved.
 ECHO:
-ECHO If you are able to identify the cause of your issues, please bring it up on
-ECHO:
+CALL :message "If you are able to identify the cause of your issues, please bring it up on"
 ECHO ghacks-user.js GitHub repository.
 ECHO:
 ECHO:
--- a/prefs cleaner.bat
+++ b/prefs cleaner.bat
@@ -3,18 +3,27 @@ TITLE prefs.js cleaner
 
 REM ### prefs.js cleaner for Windows
 REM ## author: @claustromaniac
-REM ## version: 1.0b2
+REM ## version: 1.0b3
 
 SETLOCAL EnableDelayedExpansion
-CALL :message "This batch should be run from your Firefox profile directory. It will remove from prefs.js any entries that exist in user.js, allowing inactive preferences to reset to default value."
-CHOICE /M "Continue"
-IF ERRORLEVEL 2 ( EXIT /B )
+:begin
 CLS
+CALL :message "This batch should be run from your Firefox profile directory."
+CALL :message "It will remove from prefs.js any entries that also exist in user.js."
+CALL :message "This will allow inactive preferences to reset to default value."
+ECHO:
+CHOICE /C SHE /N /M "Start [S] Help [H] Exit [E]"
+CLS
+IF ERRORLEVEL 3 ( EXIT /B )
+IF ERRORLEVEL 2 (
+	CALL :showhelp
+	GOTO :begin
+)
 IF NOT EXIST "user.js" ( CALL :abort "user.js not found in the current directory." 30 )
 IF NOT EXIST "prefs.js" ( CALL :abort "prefs.js not found in the current directory." 30 )
-IF EXIST "webappsstore.sqlite-shm" ( CALL :abort "Running this script while Firefox is also running is useless. Close Firefox and try again." 60 )
+CALL :FFcheck
 CALL :message "Backing up prefs.js..."
-COPY /B /V /Y prefs.js "prefs-backup-!date:/=-!, !time::=.!.js"
+COPY /B /V /Y prefs.js "prefs-backup-!date:/=-!_!time::=.!.js"
 CALL :message "Cleaning prefs.js... (this can take a while)"
 CALL :cleanup
 CLS
@@ -22,6 +31,27 @@ CALL :message "All done."
 TIMEOUT 5 >nul
 EXIT /B
 
+REM ########## Abort Function ###########
+:abort
+CALL :message %1
+TIMEOUT %~2 >nul
+EXIT
+REM ########## Message Function #########
+:message
+ECHO:
+ECHO:%~1
+ECHO:
+GOTO :EOF
+REM ####### Firefox Check Function ######
+:FFcheck
+TASKLIST /FI "IMAGENAME eq firefox.exe" 2>NUL | FIND /I /N "firefox.exe">NUL
+IF NOT ERRORLEVEL 1 ( 
+	CLS
+	CALL :message "Please, close Firefox to continue."
+	TIMEOUT 3 >nul
+	GOTO :FFcheck
+)
+GOTO :EOF
 REM ######### Cleanup Function ##########
 :cleanup
 SETLOCAL DisableDelayedExpansion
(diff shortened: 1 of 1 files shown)


Thorin-Oakenpants:
I'll leave all of this up to you and earthlng, if that's ok

crssi:
@claustromaniac how do you decide which pref should be dropped out and which not?

claustromaniac:
@crssi This loops over the contents of prefs.js, line-by-line. The flow goes like this:

1. Store the line in a temporary variable (A).
2. Trim spaces in that line and store it in a second temporary variable (B).
3. Check if var B starts with "user_pref". If it doesn't, just add var A exactly as it is to the new prefs.js file and start processing a new line. Otherwise, go to step 4.
4. Trim var B, from the first comma (,) to the end of the line.
5. Look for any instance of var B in user.js. If it exists, drop that line and begin processing a new line in prefs.js. If it's not in user.js, add var A to the new prefs.js file.

The result: all preferences that exist in both prefs.js and user.js are removed from prefs.js, and all active preferences in user.js are re-applied to prefs.js once Firefox is restarted.

crssi:
I see, but what is the point of that?
You actually remove everything that exists in user.js from prefs.js and when FF is started again you get the same prefs.js as was before "cleaning".

Am I missing something in this logic?

claustromaniac:
The FIND command used for step 5 will return _true_ whether it finds the string in the beggining of a line or in the middle, which means the script effectively drops preferences that are commented out with // and preferences that appear in deprecated sections in user.js. Those are the preferences this script is targeting, and this method works because those are not re-applied once Firefox is restarted.

crssi:
Very nice approach indeed, but it has flaws that needs to be polished.
For example
1. Lets say that I am using ghacks vanilla user.js
In user.js, for example 2504, that is `dom.vr.enabled` is inactive, but I for some reason set this manually through about:config to `true`.
After processing this setting in my profile is lost.
The same goes if I set any pref to inactive (not to force to different) and set it trough about:config on FF settings GUI to some other value than default, it will be lost.

2. Lets say that I am still (for some reason) on FF version 56 and I am using ghacks vanilla user.js
All settings that are deprecated with FF57, but still needed in FF56 will also get lost.

claustromaniac:
Thanks for the feedback.

1. I'm aware of that. It's a trade-off. I don't see any reliable way to prevent that, but a simple workaround would be to compare the new prefs.js and the backup and simply add the removed lines that you'd like to keep to the overrides, so those don't get removed again the future.
2. That can be done, but it would only work if Firefox is installed, and not on portable profiles. Also, if you had a portable profile and Firefox installed at the same time, the script would treat all profiles as being from the installed version of Firefox. It's probably worth looking into, anyway.

crssi:
In the script published at #303 all of this is covered.

claustromaniac:
Apparently, the approach in that script involves ignoring all preferences above section 9999, and I'm not sure that's what @Thorin-Oakenpants wants. Pants, you should re-state your preference considering this consequence of not ignoring everything above 9999.

> Um .. so what does this do exactly?
> 
> Take commented out preferences from user.j section 9999 and remove them from pref.js if they exist in there? Is that what it is meant to do? And does it account for the fact that the ESR sections in deprecated do not use // but instead have a whole section switch?
> 
> ^^ If that is what this does (section 9999) then why not just do the entire user.js?

earthlng:
1. what @claustromaniac said.
2. If you're not using the latest FF you'll need to make adjustments in your user.js. It's the same for ESR users. An automated solution can never be perfect because for example all the prefs that we removed from the user.js. However we can adjust to that by making prefs inactive for 1 or 2 versions before removing them from the user.js. Then something like this script will be the best solution for users with the latest FF+user.js and an override file.
Point 1 will always be a problem but as @claustromaniac said, you can compare the prefs.js backup + the new prefs.js after having the user.js applied to it again, to find all the prefs that you changed manually in about:config, so that you can add them to the override file.

earthlng:
`IF EXIST "webappsstore.sqlite-shm"` - is that really a reliable way to test if FF is running?

claustromaniac:
@earthlng Before Pants suggested that method, I was going to look into the running processes to determine if Firefox was running, but after thinking it, the advantage of this method is that you can run the script in a secondary profile while Firefox is running, which _shouldn't_ pose any problems.

crssi:
@carmenbianca Yes, ESR in considered, in fact the section9999 is dissected per FF version, so you get only part thats applicable for your FF version as default or version by your desired by your input.
Why not whole user..js and only section 9999?
Because section9999 is deprecated and all before that is not.
Going through everything will produce undesired removals, where user might not even be aware for a few days what has happened.

@earthlng point 1 is not problematic if you go only though section9999

But I am just saying,

earthlng:
@crssi yes your script is very nice in that it allows the user to select the profile to process and the version until which deprecated prefs should be reset. :+1: But it does not reset prefs that we moved to the 4600 section and/or prefs that we made inactive. That's the big difference between this script and yours.
For example at one point we made the indexedDB pref inactive because we noticed that webextensions now also rely on that. If someone used that previous user.js and updates to a new version, then with your script he will still have IDB disabled.

claustromaniac:
Even if this script is accepted as the standard solution, I see no reason not to keep crssi's script as an alternative. Also, it doesn't suffer from CMD's limitations, so it could be the only real option in some rare cases.

earthlng:
Sure we can do that but I just don't really see the usefulness of a script that only resets the least important group of prefs out of the 4 categories that should be reset (made inactive, deprecated, moved to 4600, and removed ones).
For Windows users who use the latest user.js as is, and have all their overrides neatly organized in a .js file, this script is a very nice solution to automatically reset all 4 categories. They can run the updater.bat first and then this script afterwards and don't need to go download + run our latest scratchpad [changes-only] script. I think that's exactly what Pants wanted.
For people who make changes in about:config without adding them to the user.js or an override.js, no script can do the cleanup for them without the possibility of also resetting some of their manual changes.
Even without a cleanup script, if they manually set a pref that's currently inactive in the user.js and we later make that pref active with possibly another value than they had, their change will be gone when they update the user.js.

earthlng:
I haven't tested the script yet but I can already tell that it should only be used when you don't have legacy addons installed anymore. Because all SDK addons have a pref like this which includes an `!` :
```
extensions.<extensionname>.sdk.rootURI;jar:file:///<profilefolderpath>/extensions/<extensionname>.xpi!/
```

claustromaniac:
@earthlng I haven't tested exhaustively myself but, unlike the merge function in the updater script, the cleanup function in this one doesn't need to modify the lines before adding them. It only modifies them temporarily to search for them in user.js, which means that every line added by this script is actually added the same way that comments are added by the merge function in the updater: preserving !'s.

I have legacy addons installed and didn't lose any !'s by running the script myself.

Anyway, I'm researching a seemingly possible way to finally stop having issues with !'s when using such functions. Can't promise anything yet, but there is hope.

earthlng:
Ok I've tested it and it works perfectly! There are only 2-3 things that I don't like or could be improved:
1. the comma+space in the backup filename - I'd prefer an underline - do you mind?
2. I'm still not convinced the webappstore.sqlite-shm check is reliable enough - this probably only works atm because uBlock keeps some its config in that file. We can keep that check and abort if it detects it but I would add an additional note before the prompt, like "make sure to close Firefox before continuing".
3. the first message would be nicer if it wasn't all on the same line.

Other than that I love it! I especially like the way you structured it with all these calls - makes it shorter overall and super easy to read. Good stuff, well done!

I still think both your scripts deserve a nice header with the script name, author and version outputted to the console, maybe even some ASCII art. I only bring that up again because you didn't respond to it the last time, but of course it's totally up to you if want to do it or not.

Thorin-Oakenpants:
On a brand new profile `webappsstore.sqlite-shm` still exists if FF is running. You could code for an extra step

my shitty code example
```
if `webappsstore.sqlite-shm` is present
   then msg ("FF is runing", OK, abort)
else
  msg ("blah blah", OK, cont)
end if
```

claustromaniac:
I took a few liberties with the latest commit. Be sure to read the notes, test it if you want, and tell me what you think. I wouldn't mind rolling back changes.

The most significant one in this commit : it now detects if firefox.exe is running.

@earthlng Any specific suggestions or ideas for some ASCII art? It's not that I have anything against that, I just tend to be lazy when it comes to that kind of thing.

Thorin-Oakenpants:
>  ideas for some ASCII art?

I once wrote an encryption script in VB that emulated an [enimga machine](https://en.wikipedia.org/wiki/Enigma_machine) .. and one of my tests produced a result, that when loaded in a jpg viewer, showed a nude redhead on a Harley .. true story xD

http://patorjk.com/software/taag/#p=display&f=Elite&t=claustro

earthlng:
I think we ditch the ASCII art idea because ECHO is just too stupid for anything semi-nice.

The changes look good but I wonder what could ever cause points 5-7 in showhelp to be necessary?
I mean prefs.js isn't even touched until everything is done and then it's a simple MOVE /Y cmd. Even if something could go wrong (which I just don't see happening), after restoring the backup prefs.js in point 4 everything should be exactly as it was, no? What am I missing?

earthlng:
Instead of looping the ffcheck I think it would better to use PAUSE and a message like "Firefox is still running. If that process is using a different profile you can continue, otherwise close Firefox first!"

claustromaniac:
>  I wonder what could ever cause points 5-7 in showhelp to be necessary?

If you try the initial commit, without the fix for long lines, you'll see what I'm talking about. The longest lines in prefs.js are typically `browser.uiCustomization.state` and `extensions.webextensions.uuids` and those could get truncated by the FIND command, which just killed the entire prefs.js file when you launched Firefox.

I ditched the FIND command for FINDSTR, which doesn't truncate lines so as long as they are read from a file, but I read somewhere the CMD shell still has some (very large) limit of characters per line, and since the lengths of those lines vary from user to user, I decided it was prudent to mention that just in case.

This is one of the limitations I was referring to before, when I said that PowerShell doesn't suffer from CMD's limitations.

earthlng:
> ... I decided it was prudent to mention that just in case.

oh okay, yeah I didn't think about that. Thanks for clarifying :+1: 
I've just retested the script and only found one thing that could be improved... the default window size (at least for me) is too small to display the full showhelp text and it auto-scrolls down a bit. A `MODE 80,34` at the top of the `:begin` part would solve this.
I also experimented with different COLOR changes throughout the file but IDK if we want to go there.

@Thorin-Oakenpants is the filename with a space okay, or should we name it `prefsCleaner.bat` or something like that? I personally prefer filenames without spaces.

Thorin-Oakenpants:
>  I personally prefer filenames without spaces.

ditto

earthlng:
Damn I still have 300+ prefs in my stripped prefs.js xD

earthlng:
Pants, have you seen this part of my comment above?

> However we can adjust to that by making prefs inactive for 1 or 2 versions before removing them from the user.js.

To get a clean slate, Windows users now only need to run `ghacks-clear-all-up-to-57-[removed].js` + this script once. If we do what I quoted above, the next time they can simply update their user.js + run the prefsCleaner.bat and it will reset all 4 categories (removed, deprecated, 9999 + 4600).

earthlng:
@claustromaniac 

> **FINDSTR Line Length limits**
> Files specified as a command line argument or via the /F:FILE option have no known line length limit. Searches were successfully run against a 128MB file that did not contain a single <LF>.
> 
> Piped data and Redirected input is limited to 8191 bytes per line. This limit is a "feature" of FINDSTR. It is not inherent to pipes or redirection. FINDSTR using redirected stdin or piped input will never match any line that is >=8k bytes. Lines >= 8k generate an error message to stderr, but ERRORLEVEL is still 0 if the search string is found in at least one line of at least one file.

Since you're specifying the file as a command line argument, I think we should be gucci. The variable should be able to hold the full string, I assume.

Thorin-Oakenpants:
> Damn I still have 300+ prefs in my stripped prefs.js xD

My prefs.js went from 120+kb (didn't keep exact track) to a current 46k because as I removed legacy prefs, I also cleared loads of extensions prefs (noscript, findbartweak, scrapbookx, fvdspeeddial each had 50 entries as well as uBo, uM and others). If i remove my user.js prefs, I still have 173 left in prefs

> To get a clean slate, Windows users now only need to run ghacks-clear-all-up-to-57-[removed].js + this script once. If we do what I quoted above, the next time they can simply update their user.js + run the prefsCleaner.bat and it will reset all 4 categories (removed, deprecated, 9999 + 4600).

Ummm .. yeah. I get that. But future removed items will be in *changes scratchpads. But yes, this script is good because it works on ANY user.js including ones people have changed from ours, even in major ways. The scratchpad scripts still come in handy for users though

earthlng:
https://github.com/ghacksuserjs/ghacks-user.js/commit/fbd9f344af3cb6d0fb320de57497278887cc4b83
took your v1.0b4 + added `MODE 80,34` to the showhelp function. Thanks for another great script!

claustromaniac:
It was entertaining. Thanks for the feedback, and for everything, as always. 😸

PR #313, opened 2017-12-16
Updater for Windows v4.0

claustromaniac:
You wouldn't believe the things I had to experiment with to finally come up with this new merge function. I really leveled up during the making of this version.

I still marked it as 4.0b because I didn't test it too much, and as always, I'm open to suggestions, etc.

By the way, I always allow edits from maintainers, in case you want to make changes yourselves.

(changes listed in the commit description)

### DIFF
--- a/updater.bat
+++ b/updater.bat
@@ -3,7 +3,7 @@ TITLE ghacks user.js updater
 
 REM ### ghacks-user.js updater for Windows
 REM ## author: @claustromaniac
-REM ## version: 4.0b1
+REM ## version: 4.0b2
 
 SET _myname=%~n0
 SET _myparams=%*
@@ -27,7 +27,7 @@ IF DEFINED _updateb (
 			REM 	* Begin the normal routine
 			REN "[updated]!_myname!.bat" "[updated]!_myname!.bat.old"
 			DEL /F "[updated]!_myname!.bat.old"
-			CALL :message "Script updated^^!
+			CALL :message "Script updated^!"
 			TIMEOUT 3 >nul
 			CLS
 			GOTO begin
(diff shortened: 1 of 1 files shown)


earthlng:
Holy shit, this is quite ingenious. But I'm not sure if this really fixes the long lines issue:

> a batch file cannot set a variable that is longer than the maximum command line length (8192).

[source](https://stackoverflow.com/questions/21723275/how-to-override-batch-variable-number-of-characters-limit)

But the speed improvements are massive, and I think 8KB should be more than enough anyway. The question is, is there a space limit on total length of all variables? Apparently there's a 32KB limit for the combined length of all Environment variables but IDK if that also applies to variables in the script that are stored with SET.

The other issues are that empty lines are removed and colons in comment + inactive user_pref lines are replaced with a comma.

earthlng:
`CALL :message "Script updated^^!` is also missing a quotation mark at the end, or does it have to be this way because of the exclamation mark?

earthlng:
keeping token 1 in the 2nd FOR loop and adjusting the script accordingly, can retain empty lines. No idea how to fix the colon/comma issue though. Probably needs another FOR loop to delim first with the colon and then with the comma.

claustromaniac:
> I'm not sure if this really fixes the long lines issue

Hmm. I didn't consider that.

> Apparently there's a 32KB limit for the combined length of all Environment variables but IDK if that also applies to variables in the script that are stored with SET.

I wasn't aware of that. If that's true, it should apply to those. I'll have to test a bit.

The other issues were simply because I didn't have time to look carefully for all the details when I made all those formatting changes yesterday. I knew I was probably missing stuff but I decided to push a first commit anyway so you could see what I made and try it out.

earthlng:
If those length limits are not a problem then I think a mix between the old version and the new would be the way to go. I don't think splitting at colon and/or comma like the new version does will ever work, because any line that ends with a colon or comma will have that char removed.
So I think we need to keep the old way of using FIND and split at ] and then, if possible, do the lookup with the variables for the speed improvements.

claustromaniac:
> Apparently there's a 32KB limit for the combined length of all Environment variables

Before I begin testing, where did you read that? The link above doesn't say anything about it being a combined limit.

earthlng:
Dude you're a GOD! It works, keeps all the lines intact **and** is incredibly fast - freakin amazing!

https://blogs.msdn.microsoft.com/oldnewthing/20100203-00/?p=15083

I've tested it with a copy of greprefs.js which contains most of FF's default prefs (nearly 2200 in the file)
and as far as I can tell it worked fine. But you should probably do some testing yourself just to make sure.
I will add some long comments behind a bunch of those user_pref lines so that the total env size should be longer than 32K and see how that goes.

earthlng:
> I don't think splitting at colon and/or comma like the new version does will ever work, because any line that ends with a colon or comma will have that char removed.

apparently what I said above was wrong because your b2 version still used FINDSTR and did not remove those chars. The only problem with that version was that empty lines got stripped but that can be easily fixed with a `if "%%J"=="" ( ECHO: ) ELSE ( ...`. So if you think FINDSTR is better than FIND we can go back to using that IMO. up to you

earthlng:
> I've tested it with a copy of greprefs.js

forgot to say that I stripped that file of all comments and it was still ~130kb in size.
The longest line in that file is just under 3000 bytes long. 8KB **IS** the limit but that should be plenty enough. With the filesize of 130KB I think it's safe to assume that the 32KB limit does not apply to local variables.

claustromaniac:
> apparently what I said above was wrong because your b2 version still used FINDSTR and did not remove those chars.

True. I just tested.

earthlng:
YUUUUUGE improvements! xD Thanks @claustromaniac

PR #314, opened 2017-12-18
Convert BAT line-endings to CRLF

claustromaniac:
LF is messing with calls to :message in the updater, at least on my end.

EDIT: Curiously, the updater seems to have CRLF EOL's in the repository, but GitHub still hands out LF versions [here](https://raw.githubusercontent.com/ghacksuserjs/ghacks-user.js/master/updater.bat) and [here](https://github.com/ghacksuserjs/ghacks-user.js/raw/master/updater.bat), which makes me wonder if this can be fixed at all. I guess it ultimately means the `-updatebatch` switch produces unexpected results.

### STR:
1. Start the updater with `-updatebatch` switch. Let it update itself.
2. Press H in the menu to see the available switches, and you should see CALL errors where the names of the switches would be.
3. Close the updater.
4. Change line endings to CRLF using Notepad++ or the like.
5. Start the updater **without** `-updatebatch` switch.
6. Repeat step 2. Errors should be gone.

I'm leaving this PR open because even if this may not fix the issue, it can't make things worse.

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -4,5 +4,6 @@
 *.md text
 *.yml text
 *.txt text
+*.bat text eol=crlf
 
 *.png binary


earthlng:
doesn't seem to work. I also edited the file to see if that would update it but it doesn't.
https://github.com/ghacksuserjs/ghacks-user.js/commit/1571e067e752da31dcb1c420f49548a1ed67c095 as a workaround. Either that or we need to remove `* text=auto` from .gitattributes again, I think.

claustromaniac:
I was going to push a similar solution but you beat me to it. It works, thanks. 👍

claustromaniac:
On second thought, manual downloads from those links will still fail, but a native solution to that would get pretty complex, I think. 

🤔

claustromaniac:
Well, another workaround would be to simply replace _message_ calls in the showhelp function with ECHO's. It should work since it seems it's the only place where LFs are giving trouble, but it wouldn't prevent such issues from happening again in future updates (if we ever need to update this again). Also, I don't know how it works in other setups. Maybe the raw with LF endings works worse (or better) for other people. Who knows.

earthlng:
I think it's possible that we would just need to use git to push the bat files again. Maybe then the batches in the repo downloaded from raw would be CRLF by default, but I'm not sure. And I don't have git and it's been years since I last used it, so I'm worried I might fuck things up.

earthlng:
from http://adaptivepatchwork.com/2012/03/01/mind-the-end-of-your-line/

> *.txt -text Unset all files matching the filter. These files will never run through the CRLF to LF replacement.

let's try that

claustromaniac:
That seems to have worked, as in, no more issues with CALL :message, but it's super weird because the raw files still have LF endings, according to Git and Notepad++ LOL

earthlng:
@carmenbianca can you help us out here, please? How can we have all batch files in the repo itself with CRLF so that raw will not provide LF'ed batch files?
Back when we added the .gitattributes file you also made a pull request that changed the user.js to LF.
If it's not too much to ask, could you fix the gitattributes for us, so that it still works the way you wanted it to work, and if necessary create a PR for the 2 batch files.
It's fine if everything else is still LF-ed in the repo but the batch files should always have CRLF, if that's even possible. Thanks

Thorin-Oakenpants:
https://github.com/ghacksuserjs/ghacks-user.js/blob/master/.gitattributes already says `*.bat eol=crlf`

If you download the zip the file is CRLF. I do not think you can change how github displays the RAW version - I think @nodiscc explained this once

look-ma-no-hands

claustromaniac:
> already says *.bat eol=crlf

Yeah, because we added it yesterday.

If what @nodiscc said is true, it means that either..

- the opposite is currently true, or...
- GitHub hands out LF versions to Windows users, and CRLF versions to Unix users. Which would be plain stupid.

If there is no way to change line endings in the raws, we're left with very few options. We can...
1. Warn people that raw versions _may not_ work as intended on some systems.
2. Host those files elsewhere. 👎 
3. Zip the batchfiles individually, or compile them into EXEs, and put the current sources in a folder? Like the typical _source_ folder. 👎 👎 

Any other ideas?

carmenbianca:
@earthlng `*.bat eol=crlf` seems like the correct solution to me. When I clone the repo, I get:

```
carmen@carmen-thinkpad:/tmp/ghacks-user.js> file updater.bat
updater.bat: DOS batch file, ASCII text, with CRLF line terminators
```

This seems right to me. Perhaps it's a bug with GitHub as @claustromaniac says? I don't know. I frankly don't use GitHub very often.

Though the current .gitattributes file doesn't look quite right. [This article](https://help.github.com/articles/dealing-with-line-endings/) shows the supposed correct way:

```
# Set the default behavior, in case people don't have core.autocrlf set.
* text=auto

# Explicitly declare text files you want to always be normalized and converted
# to native line endings on checkout.
*.c text
*.h text

# Declare files that will always have CRLF line endings on checkout.
*.sln text eol=crlf
```

That is: Just declare `* text=auto` and then explicitly declare some other file types as text, and optionally their eol. Either way the current .gitattributes seems to work fine as well, though.

earthlng:
I've fixed it by re-uploading the files. Also removed the workaround part again in the updater.bat.

The fact that Carmen cloned the repo on linux (or mac) and git converted the file to CRLF seems to suggest that the .gitattributes works as intended as well. Yeah I know it doesn't look quite right but I read somewhere that `* text=auto` is not really necessary if you explicitly declare the file extensions used in a project. And from what I understand (I could be wrong) it might be the line that tells github to always use LF for text files in the repo, which is exactly what we don't want for the raw batch files.

claustromaniac:
Finally!

I tried re-uploading the files on my fork yesterdey, to see if that worked, but it didn't. I must've done something wrong.

It's a relief. I really didn't want to add more workarounds in the scripts lol.

earthlng:
Yeah, finally. I should have thought of that a lot earlier. Well better late than never, right xD
And thanks @carmenbianca for chiming in here, your comment clarified a few things for me. :+1:

carmenbianca:
@earthlng You're welcome :) Glad it works now. Seems very strange to me that a re-upload would do it. Maybe the file was internally stored with LF line endings, and Git converted it to CRLF in the working directory on-the-fly? I have no idea. Line endings are messy.

earthlng:
I think the re-upload only worked because the gitattributes now tells it **not** to convert it to LF.

> Git converted it to CRLF in the working directory on-the-fly

yes that's what the gitattributes file is for, afaik. If it only says "text" it will convert the file to LF on linux and mac, and to CRLF on Windows when you pull it with git. And when you push a changed file again, it will convert it back to LF for storing in the repo. Our line for batch files now tells it to use CRLF for all OS'es.
At least that's how I *think* it works.

Issue #283, opened 2017-11-25
Task: wiki page on updater scripts


Thorin-Oakenpants:
- [x] complete updater scripts to required standard
   * Updater.bat v3.0 FINAL :tada:
- [x] does the mac version need finalizing?
   * No, it does the "downloading + appending" part but doesn't support multiple .js files or any of the other parameters that the Windows version has.
- [x] rearrange wiki pages to make more sense
- [x] update user.js readme header 
- [x] rewrite bulk pref resetting wiki - add in prefsCleaner bat
- [x] rewrite wiki page on extra .js file(s)
- [x] write updater scripts page

earthlng:
I wrote a HOWTO for the Windows version to which we can link in the wiki to keep that nice and short.
As the person in charge now, I delegate the updating of the Wiki and the other tasks to Lady Thorina of House Oakenpants. When finished, report to my quarters for debriefing and nightly duties, sincerest your Sir Earthling of House Terra

Thorin-Oakenpants:
>  to which we can link in the wik

Wondered why you used a new issue. Looks good. But it **must** be in the wiki. I will re-arrange the wiki numbering and create a new page for windows updater and one for mac/linux (unless you think we can combine them) eg (off the top of my head, just bouncing ideas)

### Getting Started
* 1.1 - 1.4
* 1.5 - Setting Extension Permissions

### Updating, Maintenance
* 1.7 - Bulk Pref Resetting [Scratchpad]
* 1.8 - Updating (stuff about keeping your overrides in separate files)
   * 1.8a - old method up to ff58 (eventually we can strikethru it when ESR59 lands)
   * 1.8b - windows
   * 1.8c - mac/linux

### How To
* 2.1 etc

earthlng:
I'd keep the wiki short and just link to this issue.

Thorin-Oakenpants:
I think you mean the other issue you created, not this one. But nah, the info should be in one location (wiki), not two (wiki + issues). You'll see. I'm not in any hurry, I have other things to deal with. As long as I get it done before FF58 stable lands

earthlng:
Look, the wiki is already way too confusing with too much text, even for the ones who actually try to read it. Lets try to keep it relevant and short, with links to "more details" for the people that it concerns:
for example:

[...]

PR #317, opened 2017-12-24
Updater for Windows 4.1

claustromaniac:
Just very minor changes, mostly adding some missing information. including the changes suggested in https://github.com/ghacksuserjs/ghacks-user.js/issues/283#issuecomment-353766065.

@Thorin-Oakenpants I reply to you here, since I can't post comments on #283 because it says the conversation _"has been locked and limited to collaborators"_ (strange, I know). Your work on the wiki looks good to me. I especially like what you said about hidden prefs in section 3.2, good call!

### DIFF
--- a/updater.bat
+++ b/updater.bat
@@ -146,6 +146,10 @@ IF EXIST user.js (
 				CALL :message "Appending..."
 				COPY /B /V /Y user.js+"user.js-overrides\*.js" user.js
 			)
+			(
+				ECHO:
+				ECHO:user_pref("_user-overrides.js.parrot", "Parrot successfully overridden");
+			)>>user.js
 		) ELSE (CALL :message "No override files found.")
 		ECHO:
 	) ELSE (
@@ -157,6 +161,10 @@ IF EXIST user.js (
 			) ELSE (
 				CALL :message "user-overrides.js appended."
 			)
+			(
+				ECHO:
+				ECHO:user_pref("_user-overrides.js.parrot", "Parrot successfully overridden");
+			)>>user.js
 		) ELSE (CALL :message "user-overrides.js not found.")
 		ECHO:
 	)
--- a/updater.bat
+++ b/updater.bat
@@ -246,8 +246,8 @@ ECHO:     together, a user-overrides-merged.js file is also generated in the roo
 ECHO:     directory for quick reference. It contains only the merged data from
 ECHO:     override files and can be safely discarded after updating, or used as the
 ECHO:     new user-overrides.js. When there are conflicting records for the same
-ECHO:     pref, the value of the last one declared will be used. Visit the GitHub
-ECHO:     repository for usage examples and more detailed information.
+ECHO:     pref, the value of the last one declared will be used. Visit the wiki
+ECHO:     for usage examples and more detailed information.
 CALL :message "  -multiOverrides"
 ECHO:     Uses any and all .js files in a user.js-overrides sub-folder as overrides
 ECHO:     instead of the default user-overrides.js file. Files are appended in
--- a/updater.bat
+++ b/updater.bat
@@ -1,9 +1,10 @@
 @ECHO OFF & SETLOCAL EnableDelayedExpansion
 TITLE ghacks user.js updater
 
-REM ### ghacks-user.js updater for Windows
+REM ## ghacks-user.js updater for Windows
 REM ## author: @claustromaniac
-REM ## version: 4.0
+REM ## version: 4.1
+REM ## instructions: https://github.com/ghacksuserjs/ghacks-user.js/wiki/3.3-Updater-Scripts
 
 SET _myname=%~n0
 SET _myparams=%*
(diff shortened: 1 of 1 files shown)


Thorin-Oakenpants:
conversation is unlocked now, forgot about that. I'll let earthlng handle the PR commit

updater.bat
```diff
@@ -84,10 +85,10 @@ IF NOT EXIST user.js (
...
 	FOR /F "skip=1 tokens=1,2 delims=:" %%G IN (user.js) DO (
 		SET /A "_line+=1"
-		IF !_line! GEQ 4 ( GOTO exitloop )
-		IF !_line! EQU 1 ( SET _name=%%H )
-		IF !_line! EQU 2 ( SET _date=%%H )
-		IF !_line! EQU 3 ( SET _version=%%G )
+		IF !_line! GEQ 4 (GOTO exitloop)
+		IF !_line! EQU 1 (SET _name=%%H)
+		IF !_line! EQU 2 (SET _date=%%H)
+		IF !_line! EQU 3 (SET _version=%%G)
```
claustromaniac: ( SET _version=%%G ) adds an unnecessary space after the version number, resulting in an output like:

> ghacks user.js version 57 , 20 November 2017

earthlng:
> ~~Comments and _user.js.parrot lines are appended normally~~
> One-line comments and ...

re: merge - why did you change this? multi-line comments are appended normally, too.

I also don't really like the parrot change. Users can add them themselves if + where they want them, IMO.

claustromaniac:
> re: merge - why did you change this? multi-line comments are appended normally, too.

It's not technically accurate, because of the way it handles user_pref lines within multi-line comments. With this change it neither affirms nor denies that multi-line comments are appended. A full explanation would take many lines more. It would be redundant, too.

> I also don't really like the parrot change. Users can add them themselves if + where they want them, IMO.

For me, either way is OK. It's up to you folks.

Thorin-Oakenpants:
We can drop the parrot, I kinda like it though, because the last line in the user.js is a SUCCESS

parrot

And if we kept it (parroting the last section), it should be a diff name because it's not ghacks user.js cannon

Side question: I hope the parrot code (if we keep it) does not keep appending more parrots every update until you end up with dozens of parrots lines at the start and end (I assume you have this covered)

Thorin-Oakenpants:
Lets say we drop the parrot thing - If users add their own parrots, I thought the code did not append parrots (well at least not OUR parrot pref name) - right?

claustromaniac:
My reasoning (and I didn't actually test shit, this is just the way I understand things): Firefox will only read `prefs.js` and then `user.js`, and the updater just appends the overrides to `user.js`. Using the same parrot for that purpose should be ok, because when Firefox correctly reads the last parrot line, you know everything above that line was also read correctly. If something fails in the middle it will not get to the part with the overrides anyway. Isn't that how the parrot thingy is meant to work?

> Side question...

The script DLs `user.js`, appends overrides, and **then** sticks the parrot line at the end of the new modified `user.js`. The parrots can't multiply that way.

> Lets say we drop the parrot thing - If users add their own parrots, I thought the code did not append parrots (well at least not OUR parrot pref name) - right?

The script appends parrot lines regardless of the mode (append or merge), which means that parrots are never merged, otherwise the merge algorithm would render them completely useless. I've been adding parrots to the overrides before I even wrote the first version of this script, so I think it is a good practice, I just don't mind if the script takes care of that part for me or not because it won't really save me much time by doing so anyway. I mean, my override files are only touched by me, so the parrots I add there won't go anywhere if I don't do anything.

claustromaniac:
Maybe it would be a good idea to just suggest that practice of adding manual parrots to the overrides in the wiki instead (if it's not there already). It is good advice and it can help people that don't want or can't use this script, too.

Thorin-Oakenpants:
I'll let you and earthlng work it all out :)

The parrot works exactly as you said, but you would still need to add them at BOTH the start and end of the overrides appendage

```js
/ /ghacks user.js
/* END: internal custom pref to test for syntax errors ***/
user_pref("_user.js.parrot", "SUCCESS: No no he's not dead, he's, he's restin'!");

// overrides appended by code
user_pref("_user.js.parrot", "you need something in here so the above SUCCESS is cleared");
   // ^^ I would use "overrides: syntax error"
...
...
user_pref("_user.js.parrot", "SUCCESS: No no he's not dead, he's, he's restin'!");
```

But I think we should just forget about the whole parrot thing - it was a silly idea, completely SILLY ... Monty Python SILLY

earthlng:
> I thought the code did not append parrots

@Thorin-Oakenpants to help you understand how the script works, you should really give it a try. You don't have to run it in your real profile, you can just create a new empty folder somewhere, place the updater, a user.js and an override file in there and run the script with or without some of the cmd-line parameters.

Thorin-Oakenpants:
^^ true, but I have the utmost confidence in you two :) That and time factors .. or maybe I'm just making excuses

PS: closing wiki issue, will rejig numbers/order in section 2700 (persistent storage) and close that issue, and move to 58beta - if you're ok with that

Thorin-Oakenpants:
added a `PRO TIP` to 3.2 Applying Your Changes re re-using our parrot, and included the lines in the example - so we can drop the parrot thing in the updater.bat

earthlng:
got rid of the parrot stuff and merged it. Thanks

PR #321, opened 2017-12-26
Daily round of tweaks

claustromaniac:
Just to keep @earthlng  busy.

### DIFF
--- a/prefsCleaner.bat
+++ b/prefsCleaner.bat
@@ -3,7 +3,7 @@ TITLE prefs.js cleaner
 
 REM ### prefs.js cleaner for Windows
 REM ## author: @claustromaniac
-REM ## version: 1.2
+REM ## version: 1.1
 
 SETLOCAL EnableDelayedExpansion
 :begin
@@ -12,7 +12,7 @@ ECHO:
 ECHO                 ########################################
 ECHO                 ####  prefs.js cleaner for Windows  ####
 ECHO                 ####    author: @claustromaniac     ####
-ECHO                 ####          version: 1.2          ####
+ECHO                 ####          version: 1.1          ####
 ECHO                 ########################################
 ECHO:
 CALL :message "This script should be run from your Firefox profile directory."
@@ -72,10 +72,8 @@ SETLOCAL DisableDelayedExpansion
 	FOR /F "tokens=1,* delims=:" %%G IN ('FINDSTR /N "^" prefs.js') DO (
 		SET "_line=%%H"
 		SETLOCAL EnableDelayedExpansion
-		SET "_pref=!_line: =!"
-                SET "_pref=!_pref:	=!"
-		IF /I "user_pref"=="!_pref:~0,9!" (
-			FOR /F "delims=," %%X IN ("!_pref!") DO (SET "_pref=%%X")
+		IF /I "user_pref"=="!_line:~0,9!" (
+			FOR /F "delims=," %%X IN ("!_line!") DO (SET "_pref=%%X")
 			SET _pref=!_pref:"=""!
 			FIND /I "!_pref!" user.js >nul
 			IF ERRORLEVEL 1 (
--- a/prefsCleaner.bat
+++ b/prefsCleaner.bat
@@ -73,7 +73,7 @@ SETLOCAL DisableDelayedExpansion
 		SET "_line=%%H"
 		SETLOCAL EnableDelayedExpansion
 		SET "_pref=!_line: =!"
-                SET "_pref=!_line:	=!"
+                SET "_pref=!_pref:	=!"
 		IF /I "user_pref"=="!_pref:~0,9!" (
 			FOR /F "delims=," %%X IN ("!_pref!") DO (SET "_pref=%%X")
 			SET _pref=!_pref:"=""!
--- a/prefsCleaner.bat
+++ b/prefsCleaner.bat
@@ -3,7 +3,7 @@ TITLE prefs.js cleaner
 
 REM ### prefs.js cleaner for Windows
 REM ## author: @claustromaniac
-REM ## version: 1.0
+REM ## version: 1.1
 
 SETLOCAL EnableDelayedExpansion
 :begin
@@ -12,7 +12,7 @@ ECHO:
 ECHO                 ########################################
 ECHO                 ####  prefs.js cleaner for Windows  ####
 ECHO                 ####    author: @claustromaniac     ####
-ECHO                 ####          version: 1.0          ####
+ECHO                 ####          version: 1.1          ####
 ECHO                 ########################################
 ECHO:
 CALL :message "This script should be run from your Firefox profile directory."
@@ -20,19 +20,20 @@ ECHO   It will remove any entries from prefs.js that also exist in user.js.
 CALL :message "This will allow inactive preferences to be reset to their default values."
 ECHO   This Firefox profile shouldn't be in use during the process.
 CALL :message ""
+TIMEOUT 1 /nobreak >nul
 CHOICE /C SHE /N /M "Start [S] Help [H] Exit [E]"
 CLS
-IF ERRORLEVEL 3 ( EXIT /B )
-IF ERRORLEVEL 2 ( GOTO :showhelp )
-IF NOT EXIST "user.js" ( CALL :abort "user.js not found in the current directory." 30 )
-IF NOT EXIST "prefs.js" ( CALL :abort "prefs.js not found in the current directory." 30 )
+IF ERRORLEVEL 3 (EXIT /B)
+IF ERRORLEVEL 2 (GOTO :showhelp)
+IF NOT EXIST "user.js" (CALL :abort "user.js not found in the current directory." 30)
+IF NOT EXIST "prefs.js" (CALL :abort "prefs.js not found in the current directory." 30)
 CALL :FFcheck
 CALL :message "Backing up prefs.js..."
 COPY /B /V /Y prefs.js "prefs-backup-!date:/=-!_!time::=.!.js"
 CALL :message "Cleaning prefs.js...
 CALL :cleanup
 CLS
-CALL :message "All done."
+CALL :message "All done^!"
 TIMEOUT 5 >nul
 EXIT /B
 
@@ -43,9 +44,11 @@ TIMEOUT %~2 >nul
 EXIT
 REM ########## Message Function #########
 :message
+SETLOCAL DisableDelayedExpansion
 ECHO:
 ECHO:  %~1
 ECHO:
+ENDLOCAL
 GOTO :EOF
 REM ####### Firefox Check Function ######
 :FFcheck
@@ -54,9 +57,7 @@ IF NOT ERRORLEVEL 1 (
 	CLS
 	CALL :message "Firefox is still running."
 	ECHO   If you're not currently using this profile you can continue, otherwise
-	ECHO:
-	ECHO   close Firefox first^^!
-	ECHO:
+	CALL :message "close Firefox first^!"
 	ECHO:
 	PAUSE
 	CLS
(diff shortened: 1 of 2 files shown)


prefsCleaner.bat
```diff
@@ -68,12 +69,13 @@ REM ######### Cleanup Function ##########
 :cleanup
 SETLOCAL DisableDelayedExpansion
 (
-	FOR /F "tokens=1,* delims=:" %%G IN ( 'FINDSTR /N "^" prefs.js' ) DO (
+	FOR /F "tokens=1,* delims=:" %%G IN ('FINDSTR /N "^" prefs.js') DO (
 		SET "_line=%%H"
 		SETLOCAL EnableDelayedExpansion
 		SET "_pref=!_line: =!"
+                SET "_pref=!_line:	=!"
```
earthlng: this looks wrong, it makes the previous line obsolete. At this point _line is a line from prefs.js, right? FF doesn't write TABS into the file afaik. So what is this supposed to do and why is this necessary?

claustromaniac: 75 removes spaces from the input.
76 removes tabs from the input.

The output stays unchanged.

It's not currently necessary. It's just completeness. robustness, because leading tabs (wherever they come from) don't ruin the syntax in javascript files. If there *are* leading tabs, for whatever reason, Firefox will have no issues parsing those lines, but without this addition the script won't treat those lines as it should.

earthlng: To achieve what you want, the new line should be `SET "_pref=!_pref:	=!"`. See what I mean?
But since FF never writes lines with prepended TABS anyway, there's no need to add this line at all IMO

claustromaniac: Oh, you're right. I messed up lol. I've been doing so many things at once today that I'm not getting anything right.

Yeah, I agree with you. In typical usage scenarios there is currently no *need* for this line.

earthlng:
thanks for the changes and for keeping me busy :)

Thanks to our little conflict here, I noticed that what Pants wrote in the wiki is not necessarily correct:
> prefsCleaner.bat (**works on any local user.js**)

it works perfectly fine for the ghacks user.js because all our user_pref lines have the same "style" ie no spaces and always using double-quotes. But the same can not be said or assumed for *any* local user.js.

@Thorin-Oakenpants you should remove that part in the brackets from the wiki page.

claustromaniac:
Glad to be of service.

That is a very good observation, and it is interesting because there might be a reliable way to make a version of this script that can actually parse any kind of user.js. I'm thinking of something with regex, maybe.

It could be another interesting little project for me.

earthlng:
It only needs to work for *our* user.js, that's all I really care about.

> parse any kind of user.js

... is a bold claim ;) I seriously doubt that's possible with DOS.

```js
   user_pref    (  'prefname'   , // blahblah
                /* try to parse this - good luck */   true );
```

claustromaniac:
> It only needs to work for our user.js, that's all I really care about.

It's clearly not all @Thorin-Oakenpants cares about.

>  I seriously doubt that's possible with DOS.

Did you think the merge function in the updater was possible before I made it? (rethorical question)

```
   user_pref    (  'prefname'   , // blahblah
                /* try to parse this - good luck */   true );
```

*Challenge accepted.*

earthlng:
> Challenge accepted.

I was afraid you were gonna say that :)
I guess that means no more pull requests for me to review for the next couple months or so - yippie ("always look on the bright side") xD

Thorin-Oakenpants:
> you should remove that part in the brackets from the wiki page.

Done

> It's clearly not all @Thorin-Oakenpants cares about.

I just assumed it would be universal. I don't really care about anyone else's user.js. However, end users can modify the core of the user.js they get from here, and can fuk-it-up

earthlng:
> using the well known standard syntax of `//`

that doesn't matter. It's about the format of user_pref lines, it must be exactly like this `user_pref("prefname",` - anything before or after that doesn't matter.
For example the following all won't get reset:
```js
// user_pref ("prefname", ... // space(s) after user_pref
// user_pref( "prefname", ... // space(s) after (
// user_pref("prefname" , ... // space(s) after "
// user_pref('prefname', ... // using single quotes
```

claustromaniac:
OK so I had a good look, tried a few things, and I came up with a method that should be able to deal with spaces and tabs, and single quotes. The problem is FINDSTR's support for regular expressions is ridiculously limited and ridden with bugs. So, for the sake of what little is left of my own sanity, I will just leave the experiment behind (for now at least). I'm sorry @earthlng, you may have to keep reviewing stuff after all.

Anyway, the format in this`user.js` is the same that Firefox itself uses in its own javascript files, so I think removing that statement from the wiki is not really necessary. If you wanted, you could add the missing bit of relevant information instead, something like:

> prefsCleaner.bat (works on any local user.js as long as they are formatted like prefs.js)

...but it is now perfectly clear that nobody really cares about other `user.js`', so feel free to ignore this altogether.

Over and out.

earthlng:
glad you chose your sanity over proving a point :)

claustromaniac:
It's not so much about proving a point: it's the challenge. Why do you think I find writing batchfiles so much fun? It's an awful language! The sense of accomplishment delivered is pretty sick, like disarming a bomb with chewing gum and a paper clip - all MacGyver style.

I do admit that from experience I know it kind of triggers me when others tell me what I can and cannot do, but I usually don't have much trouble hitting the brakes. That's the only reason I think there still is some semblance of sanity left in me.

I forgot to mention that the easiest way to accomplish this **is very possible**, but it has a significant drawback. To elaborate, the script could simply look for the name of the preference itself in `user.js` (without the rest of the syntax, only what is inside the first set of quotes/double quotes). There can be neither spaces nor tabs, nor new lines in there, right? The drawback in question is pretty obvious: if the name of the preference is in the file but only mentioned in a comment or something, the script would still treat it like a declared `user_pref` line.

earthlng:
Yes, without proper regex support that's the only way I could think of myself as well, with the obvious drawback that you mentioned. Exactly because of that drawback I was hoping you wouldn't waste time on that solution because I was gonna have to shoot you down ie not accept a PR based on only looking at the prefname. (because it's arguably worse than what we have at the moment)
The current solution works perfectly fine for this user.js and apart from the intellectual challenge there's really no need to spend any more time on this.
I'll maybe look into finding a way to use regex but it sounds like you already tried that and it's unlikely that I can come up with something if you couldn't.

claustromaniac:
In case you're interested, the first problem I had in my tests was that I couldn't search for any number of spaces and tabs **specifically**. The only way I could make it work was by looking for any number of **characters** instead, like this:

`FINDSTR /I /R /C:".*user_pref.*\(.*[\"']browser\.rights\.3\.shown[\"'].*," user.js`

And while this works, it also means that lines like the following one would match (when they shouldn't):

`user_pref QWERTY( I sometimes break the syntax just because I can "browser.rights.3.shown", true);`

I tried using ranges (with square brackets), and many different ways of escaping spaces and tabs but nothing worked.

earthlng:
```dos
FINDSTR /I /R /C:"user_pref[ ]*\([ ]*[\"']browser\.rights\.3\.shown[\"'][ ]*," user.js
```
ignoring TABS for now, this works for lines with any number of spaces

claustromaniac:
That's odd. Did you try that in a batch file? I'm pretty sure I tried exactly the same from the command line and it didn't work for me.

earthlng:
yes I tried that with a batch file:
```dos
@ECHO OFF
SET "_pref=browser\.rights\.3\.shown"
FINDSTR /I /R /C:"user_pref[ 	]*\([ 	]*[\"']%_pref%[\"'][ 	]*," user.js
```

> that it needs not be the beginning of the line

why shouldn't it be? currently we're searching for `user_pref("pref.name"` anywhere in the user.js, which is exactly what we need to do to find all active and inactive user_prefs.

I'm also wondering if we should get rid of `/I` because I think FF preferences are case-sensitive

claustromaniac:
> why shouldn't it be? 

Single-line comments, mostly, and it goes hand in hand with the whole idea of making the thing more _universal_, because other user.js files may be formatted prepending stuff.

earthlng:
I don't understand. I just tried it with this user.js and it still detects all 7 lines:
```
 /* !!! */ user_pref("browser.rights.3.shown", true);
user_pref ("browser.rights.3.shown", true);
   user_pref( "browser.rights.3.shown", true);
	user_pref("browser.rights.3.shown" , true);
  //  user_pref('browser.rights.3.shown', true);
	//	 user_pref    (    'browser.rights.3.shown'    , true);
user_pref 	 ('browser.rights.3.shown'	, "space-tab-space in front + tab at the end");
```
What exactly do you think should be excluded ie limited with .* in front of the regex or whatever?

claustromaniac:
Yeah, it was me who didn't understand in the first place, because I'm not paying sufficient attention lulz. I misinterpreted your question. I just mentioned that when I tried before I made it explicit that it didn't need to look for pattern in the beginning of the line, but it isn't necessary, and I never said it was necessary. It was just a misunderstanding.

So, back to my previous question, would this be a worthwhile addition, after all?

earthlng:
> It was just a misunderstanding.

oh, okay nvm

> would this be a worthwhile addition, after all?

Yeah, since it's clearly an improvement and it can apparently be done with less additional code-complexity than I expected, I think this would be a worthwhile addition to the script.

claustromaniac:
Alright. So, I'm thinking this through, and there are a couple of things to consider.

1. You are right about Firefox preferences being case-sensitive, we definitely need to get rid of the `/I`.
2. Merely matching spaces and tabs doesn't cover the following scenario:

`user_pref ( /* Etiquette? What the hell is that? */ "browser.rights.3.shown", true);`

The pattern from the first snippet that I posted here covers this, but has the (minor?) drawback that it matches comments like this too:

_// I'm just innocently referring to_ **user_pref** _lines in a comment_ **(** _in case you didn't notice). This is a self-explanatory example that also mentions_ **"browser.rights.3.shown",** _because that's what needs to happen._

So, we have a pattern that can't parse inline comments that exist anywhere between `user_pref` and the comma, and we have a pattern that **can**, but that can also get mixed up any time the necessary ingredients are scattered in order across the same line in a comment.

What is worse?

earthlng:
I prefer the one I posted [here](https://github.com/ghacksuserjs/ghacks-user.js/pull/321#issuecomment-354195037) but without the `/I`

claustromaniac:
I ran some tests and discovered some limitations both in the current version of the script, and in the regex-using version that we've been discussing lately.

### Current limitations:
**Firefox allows special characters like** `! " \` **in preference names.**
1. **[ can fix ]** Delayed Expansion removes `!`'s from the search string, thus the script can't reset preferences that include !'s.
2. **[ cannot fix ]** When Firefox reads preferences in `user.js` with names in this format `'prefname"dq\bl'` it writes them in `prefs.js` in this format instead `"prefname\"dq\\bl"`. Firefox is escaping those special characters, which means there are at least two different valid ways to declare those preferences in `user.js`. Attempting to tweak the algorithm to adapt to that would be insane.

### FINDSTR /R limitations:
**Firefox allows all of the different regex special characters in preference names.**
AFAICT all characters can be escaped with \ in the search string without problems, with one exception: **asterisks**. Escaping them is possible, but requires a LOT more code, would probably slow down the algorithm to a crawl, and requires delayed expansion, which would in turn kill any `!`'s in the search string. The current version of the script doesn't have this limitation (it can properly clean preferences that include asterisks in the name).

Expect a new pull request soon.

earthlng:
> **Firefox allows special characters like** `! " \` **in preference names.**

true, but there's not a single pref with any of these chars in the name, nor will there probably ever be.

> Firefox allows all of the different regex special characters in preference names.

same as above, no existing pref with any of these chars `* ^ $ ] [ \ > <` in the name. Did I forget any?
DOT is the only one that we need to escape in the prefname.

claustromaniac:
@earthlng Good thing we cleared up all these points before I pushed any updates. Many of these are really a matter of preference after all, and if I notice that my scripts have issues or limitations I don't ever want to keep that information to myself.

Thorin-Oakenpants:
```js
user_pref("!!!!!!!!!!!^$$][\><<>>!!$!!!!", "!!!") // !!!
```

Issue #235, opened 2017-09-13
discussion: section persistent [web] storage items


Thorin-Oakenpants:
Just an idea: section all persistent storage items - just re-purpose the current 2700: COOKIES & DOM STORAGE section

I read a [long article](https://queue.acm.org/detail.cfm?id=3136953) yesterday and it re-enforced to me how we should look at things - in other words, instead of lumping stuff together under "conventional" or current rules, let's define the problem.

This would cover them under one umbrella <sup>1</sup> : a "solution" to a "problem" (did I miss anything?)
- Cookies
   * :x: Limited Control:
     * not accessible in PB mode
     * not accessible with FPI
     * no access to system cookies (eg GA for SB) - WONTFIX
- Local DOM Storage
   * :x: Little Control:
     * not accessible in PB mode?
     * Extensions (not legacy) - not accessible by host (yet) but coming
     * FPI may cause access permission problems for extensions
- IndexedDB (Offline Website Data)
   * :x: No or LIttle Control: 
     * 56+57 have various fixes plus new permissions UI coming
     * Extensions (not legacy) - not accessible by host (yet) but coming
     * FPI may cause access permission problems for extensions
     * settings to force sites to ask for permission are broken (also 50mb limit is ignored I think) - these are being looked at
- Storage API
   * No Idea: Lets just assume everything is broken!

This would also allow a section header with comments that make sense
- FPI effects
- PB Mode effects
- Container effects
- Overall strategies
- The killer extension <sup>2</sup>

Especially as the WE APIs are fixed/expanded: eg opening up access to FPI objects, IDB access. Some items are intertwined (eg cookies and local dom, something going on with IDB+FPI?/StorageAPI?) and we should consolidate. I also see something like C-AD becoming a killer recommended Extension to cover all these. It just makes sense to me

<sup>1</sup> To be clear, by persistent storage, I mean set by websites, not FF "internals" such as various histories, cache, CA's, session data etc. Keep it simple: items like workers using storage would stay under workers section

<sup>2</sup> ***It's not a question of if, but when*** - If not C-AD expanding in time, there will be others. Without looking up the tickets/names, there is one guy (Thomas Wisniewski) pushing a lot of change in these areas and even doing some of the tickets for Mozilla (pretty sure he's not a zilla employee) and he wants to create this killer extension to cover all the above as well as serviceWorkers, cacheStorage and applicationCache. He's currently landing a patch for by domain access to IDB.

[...]

PR #322, opened 2017-12-27
2700's section revamp #235

Thorin-Oakenpants:


Ratator:
Is IndexedDB automatically deleted or not? Can we expect a small summary for beginners without having to read through it completely, please?

PR #324, opened 2017-12-29
prefsCleaner 1.2

claustromaniac:
- Search string made case-sensitive, because Firefox preferences are.
- The script now uses regex, which allows it to understand `user.js` files formatted using single quotes, spaces and/or tabs in `user_pref` lines.

Trade-off: it can no longer reset preferences that include some special characters in their names. Not an issue for now, just something to remember.

See full discussion [here](https://github.com/ghacksuserjs/ghacks-user.js/pull/321).

### DIFF
--- a/prefsCleaner.bat
+++ b/prefsCleaner.bat
@@ -3,7 +3,7 @@ TITLE prefs.js cleaner
 
 REM ### prefs.js cleaner for Windows
 REM ## author: @claustromaniac
-REM ## version: 1.1
+REM ## version: 1.2
 
 SETLOCAL EnableDelayedExpansion
 :begin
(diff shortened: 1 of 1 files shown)


earthlng:
`FOR /F tokens^=2^ delims^=^"` - damn, DOS syntax is sooo strange xD
Very nice how you managed to extract the prefname AND do the escaping in a single line.
You made the script more universally useful and with fewer lines than before, fantastic!

claustromaniac:
> damn, DOS syntax is sooo strange xD

Yeah. I'm pretty sure that's the only way to use double quotes as delimiters. 

The double quotes around the FOR options are normally used to pass all of the options as a single argument to FOR. In other words, they are there because the options (like tokens, delims, etc) are delimited by spaces. Escaping double quotes inside those double quotes doesn't work as expected, so the alternative involves escaping the spaces and all of the special characters in the argument. That way the options can still be passed as a single argument without having to use double quotes for that.

**TL;DR:** I guess after all the tinkering and reading I'm just really getting the hang of this. lol

Thanks for the help and the patience.

claustromaniac:
By the way, I guess it should be ok for @Thorin-Oakenpants to mention in the wiki that this works with other user.js' if they want, right?, I still wouldn't say that this works with **any** user.js, but it certainly should work with most.

Thorin-Oakenpants:
Well, TBH, I think we should leave it out - otherwise people will think we support any old user.js. People can find out for themselves

claustromaniac:
Good point.

PR #358, opened 2018-02-08
Fix comment typos

Marc05:
Absolutely critical changes


Thorin-Oakenpants:
It's actually called "frecency" - see [here](https://dxr.mozilla.org/mozilla-central/search?q=frecency&redirect=false) and that's why it's in quote marks

supercede: is actually quite common and has been around since the 17th century - it is acceptable, but generally regarded as a misspelling (and it's the only English word ending in -sede, while there are several ending in -cede). So I'll fix that one in a commit.

I actually always used cede, and if you look at the wiktionary page it mentions Latin, which I did for 6 years including university, which is probably where I picked it up from. Cheers. I'll fix it in a separate commit (otherwise I'll edit this and then emerge and that will be 3 commits just for a single character change). Feel free to delete/close the PR.

PR #364, opened 2018-02-19
default values cleanup

earthlng:
These default values are the same in all OSes and all current Firefox versions (ESR, Release, Beta, Nightly). Apart from alerts.showFavicons these defaults are most likely never gonna change


earthlng:
@Thorin-Oakenpants I would like to commit this to get rid of some of the differences between the template and my own user.js, if you don't mind.

earthlng:
thanks :kissing:

PR #374, opened 2018-03-13
move most of the FF59 deprecated/removed prefs

earthlng:
I used the short-form for the bugzilla links and added the part-info for tickets with multiple parts. If you don't like it, feel free to change it.

### DIFF
--- a/user.js
+++ b/user.js
@@ -1805,7 +1805,7 @@ user_pref("dom.workers.sharedWorkers.enabled", false);
 user_pref("browser.sessionstore.privacy_level_deferred", 2);
 // ***/
 /* FF46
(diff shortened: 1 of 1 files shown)


Thorin-Oakenpants:
I did the deprecated sticky earlier. This is spot on! I can do the other 5 in a separate commit

Thorin-Oakenpants:
(part x) that's fine
> and added the part-info for tickets with multiple parts

any reason why we don't do this to the whole user.js?
> I used the short-form for the bugzilla links

I count 135 instances of "show_bug.cgi?id=" we could remove .. that's 2160 chars (need to check those linking to a comment first)

PR #388, opened 2018-03-26
Options> and [settings]

Thorin-Oakenpants:
While I'm at it, I'm changing the 21 instances of
- `[SETTING-56+]` to just `[SETTING]`
- `[SETTING-ESR]` to `[SETTING-ESR52]` because we'll leave those in until 62 (yes I know they may apply to earlier ESRs, but people should be upgraded). Thus no ambiguity with ESR60 vs ESR52 users for the overlap

Note: some [settings] / Options> removed as they no longer exist (did you see they removed the DRM UI option). Double check when you have time (not that they exist, but if I missed anything etc). At some stage I will go thru all these and check them as per the **FF60** interface

### DIFF
--- a/user.js
+++ b/user.js
@@ -49,8 +49,8 @@ user_pref("_user.js.parrot", "START: Oh yes, the Norwegian Blue... what's wrong 
 user_pref("general.warnOnAboutConfig", false);
 
 /* 0001: start Firefox in PB (Private Browsing) mode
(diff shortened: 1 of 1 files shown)


earthlng:
> did you see they removed the DRM UI option

no they didn't - we did :)

Thorin-Oakenpants:
> no they didn't - we did :)

doh!

Issue #323, opened 2017-12-28
FACT: CHEF-KOCH plagiarizes and passes off other people's work as his own


Thorin-Oakenpants:
edit: Oct 2108, he was banned from GitHub (may not be permanent) - FYI: he also is on GitLab: https://gitlab.com/CHEF-KOCH

---

:large_blue_circle:  **PROOF 1**: See second post re https://github.com/CHEF-KOCH/Firefox-prefs.js-cleaner

See https://www.ghacks.net/2017/12/29/automation-comes-to-the-ghacks-user-js-configuration-for-firefox/#comment-4325581

My reply if it gets published
> https://github.com/ghacksuserjs/ghacks-user.js/pull/311 - the first PR by claustromaniac 18 days ago, and that was after a week of talking about it and sample scripts
> 
> https://github.com/ghacksuserjs/ghacks-user.js/commit/fbd9f344af3cb6d0fb320de57497278887cc4b83 - v1.0 commit 15 days ago
> 
> https://github.com/CHEF-KOCH/Firefox-prefs.js-cleaner/commit/1f77f541bdd5cfbda52eecf9c7d3435c771fe4a9 - your version committed 14 days ago. Except for a few word changes such as the author, it's IDENTICAL
> 
> I call BS. This is not the first time you've been caught passing off other people's hard work as your own. Where is this reddit link that predates claustromaniac's work? I can show you the changes and discussions that claustromaniac and earthlng had and how the script changed to its final version. Yours is a single commit and IDENTICAL .. come on, where's your proof

CHEF-KOCH has not provided any proof to back up his denial

:exclamation: **FACT**: CHEF-KOCH plagiarizes and passes off other people's work as his own

---

:large_blue_circle:  **PROOF 2**:  His [ASN-blocking](https://github.com/CHEF-KOCH/ASN-blocking) repo is a direct ripoff of [ASN_IPFire_Script](https://notabug.org/maloe/ASN_IPFire_Script) , and he has been [called out](https://github.com/CHEF-KOCH/ASN-blocking/pull/1) on it, but refuses to acknowledge or do anything about it.

:exclamation: **FACT**: CHEF-KOCH plagiarizes and passes off other people's work as his own

---

[...]

PR #402, opened 2018-04-18
patches for batch scripts

claustromaniac:
The most significant change here is that I borrowed my own technique from the merge function in the updater v4 and adapted it to give the prefsCleaner a speed bump. The rest is mostly semantics.

To be honest, I considered rewriting the prefsCleaner in VBScript instead, since it's more straightforward than the updater, but the lack of console output by default still kinda puts me off. So, I just went with this.

It is worth noting that, while it wasn't my goal, with these changes the prefsCleaner should also have better tolerance for special characters within preference names (which counters [the one downside that v1.2 brought along](https://github.com/ghacksuserjs/ghacks-user.js/pull/321#issuecomment-354394222)).

### DIFF
--- a/prefsCleaner.bat
+++ b/prefsCleaner.bat
@@ -33,7 +33,7 @@ COPY /B /V /Y prefs.js "prefs-backup-%date:/=-%_%_time::=.%.js"
 CALL :message "Cleaning prefs.js..."
 CALL :cleanup
 CLS
-CALL :message "All done^!"
+CALL :message "All done!"
 TIMEOUT 5 >nul
 ENDLOCAL
 EXIT /B
@@ -56,7 +56,7 @@ IF NOT ERRORLEVEL 1 (
 	CLS
 	CALL :message "Firefox is still running."
 	ECHO   If you're not currently using this profile you can continue, otherwise
-	CALL :message "close Firefox first^!"
+	CALL :message "close Firefox first!"
 	ECHO:
 	PAUSE
 	CLS
@@ -67,10 +67,12 @@ GOTO :EOF
 REM ######### Cleanup Function ##########
 :cleanup
 (
-	FOR /F tokens^=2^ delims^=^'^" %%G IN ('FINDSTR /R /C:"^[^'\"]*user_pref[^;]*\)[ 	]*;" "user.js"') DO (SET "[%%G]=1")
+	FOR /F tokens^=2^ delims^=^'^" %%G IN ('FINDSTR /R ^^[^^\^"^']*user_pref user.js') DO (
+		IF NOT ""=="%%G" (SET "[%%G]=1")
+	)
 	FOR /F "tokens=1,* delims=:" %%G IN ('FINDSTR /N "^" prefs.js') DO (
-		FOR /F tokens^=2^ delims^=^" %%I IN ("%%H") DO (
-			IF NOT DEFINED [%%I] (
+		FOR /F tokens^=1^,2^ delims^=^" %%I IN ("%%H") DO (
+			IF NOT DEFINED [%%J] (
 				ECHO:%%H
 			)
 		)
--- a/prefsCleaner.bat
+++ b/prefsCleaner.bat
@@ -65,18 +65,16 @@ IF NOT ERRORLEVEL 1 (
 GOTO :EOF
 REM ######### Cleanup Function ##########
 :cleanup
+FOR /F tokens^=2^ delims^=^'^" %%G IN ('FINDSTR /R /C:"^[^\"']*user_pref[       ]*\([   ]*[\"'][^\"']*[\"'][    ]*," user.js') DO (
+	IF NOT ""=="%%G" (SET "[%%G]=1")
+)
 (
-	FOR /F tokens^=2^ delims^=^'^" %%G IN ('FINDSTR /R /C:"^[^\"']*user_pref[ 	]*\([ 	]*[\"'][^\"']*[\"'][ 	]*," user.js') DO (
-		IF NOT ""=="%%G" (SET "[%%G]=1")
-	)
 	FOR /F "tokens=1,* delims=:" %%G IN ('FINDSTR /N "^" prefs.js') DO (
 		IF ""=="%%H" (
 			ECHO:
 		) ELSE (
-			FOR /F tokens^=1^,2^ delims^=^" %%I IN ("%%H") DO (
-				IF NOT DEFINED [%%J] (
-					ECHO:%%H
-				)
+			FOR /F tokens^=1^,2^ delims^=^"^' %%I IN ("%%H") DO (
+				IF NOT DEFINED [%%J] (ECHO:%%H)
 			)
 		)
 	)
--- a/prefsCleaner.bat
+++ b/prefsCleaner.bat
@@ -1,18 +1,17 @@
-@ECHO OFF
+@ECHO OFF & SETLOCAL DisableDelayedExpansion
 TITLE prefs.js cleaner
 
 REM ### prefs.js cleaner for Windows
 REM ## author: @claustromaniac
-REM ## version: 1.2
+REM ## version: 2.0
 
-SETLOCAL EnableDelayedExpansion
 :begin
 ECHO:
 ECHO:
 ECHO                 ########################################
 ECHO                 ####  prefs.js cleaner for Windows  ####
 ECHO                 ####        by claustromaniac       ####
-ECHO                 ####              v1.2              ####
+ECHO                 ####              v2.0              ####
 ECHO                 ########################################
 ECHO:
 CALL :message "This script should be run from your Firefox profile directory."
@@ -29,12 +28,14 @@ IF NOT EXIST "user.js" (CALL :abort "user.js not found in the current directory.
 IF NOT EXIST "prefs.js" (CALL :abort "prefs.js not found in the current directory." 30)
 CALL :FFcheck
 CALL :message "Backing up prefs.js..."
-COPY /B /V /Y prefs.js "prefs-backup-!date:/=-!_!time::=.!.js"
+SET "_time=%time: =0%"
+COPY /B /V /Y prefs.js "prefs-backup-%date:/=-%_%_time::=.%.js"
 CALL :message "Cleaning prefs.js..."
 CALL :cleanup
 CLS
 CALL :message "All done^!"
 TIMEOUT 5 >nul
+ENDLOCAL
 EXIT /B
 
 REM ########## Abort Function ###########
@@ -44,11 +45,9 @@ TIMEOUT %~2 >nul
 EXIT
 REM ########## Message Function #########
 :message
-SETLOCAL DisableDelayedExpansion
 ECHO:
 ECHO:  %~1
 ECHO:
-ENDLOCAL
 GOTO :EOF
 REM ####### Firefox Check Function ######
 :FFcheck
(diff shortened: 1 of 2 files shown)


claustromaniac:
It is worth noting that, while it wasn't my goal, with these changes the prefsCleaner should also have better tolerance for special characters within preference names ([which is the one downside that v1.2 brought along](https://github.com/ghacksuserjs/ghacks-user.js/pull/321)).

earthlng:
Holy shit! 3 Pull Requests?!? that's a first :)

earthlng:
I'll have to wrap my head around the weird and hard to understand MSDOS logic again which might take a while

claustromaniac:
> Holy shit! 3 Pull Requests?!? that's a first :)

That's what happens when you go MIA the few days that I have nothing better to do. I have to admit I was wondering what would be your reaction when you saw all this lol The thought of submiting even more pull requests just to see what would happen was tempting.

Just kidding, of course. In fact, I can give you a breakdown of this if you want, just to prove that I don't really enjoy annoying you by making you reverse-engineer everything I dump here.

earthlng:
Haha, yeah I was slightly shocked.

A breakdown would certainly help, thanks :+1: 

I haven't looked at the new function code yet but I tested both v1.2 and v2 against the same file and the output was (nearly) identical (ie. empty lines were preserved in v1.2 but that's not an issue).
And DAMN is it fast now! awesome mate!

a couple things I noticed so far:
- It's great that we don't need Delayed Expansion as much anymore! But is that an "AND" (&) in the 1st line? Could `@ECHO OFF` theoretically fail? If so, wouldn't it be maybe better to do it in 2 lines?
- the new time thing: I ran v2.0 at minute 3 and the minute part in the backup filename was `03`, as expected. Then I ran v1.2 at minute 6 and it also had the zero already in there. (?) Is that maybe because I use Win7, and Win10 or so somehow has empty spaces in the time? Or where would the spaces come from? Either way, I don't mind if we add that line, I'm just curious why you changed that.

earthlng:
> prove that I don't really enjoy annoying

just FYI, I do statement analysis as a hobby and the word "really" in there says a lot, mate ;)

claustromaniac:
```
:cleanup
(
	FOR /F tokens^=2^ delims^=^'^" %%G IN ('FINDSTR /R /C:"^[^\"']*user_pref[ 	]*\([ 	]*[\"'][^\"']*[\"'][ 	]*," user.js') DO (
		IF NOT ""=="%%G" (SET "[%%G]=1")
	)
	FOR /F "tokens=1,* delims=:" %%G IN ('FINDSTR /N "^" prefs.js') DO (
		FOR /F tokens^=1^,2^ delims^=^" %%I IN ("%%H") DO (
			IF NOT DEFINED [%%J] (
				ECHO:%%H
			)
		)
	)
)>tempcleanedprefs
```


`FOR /F tokens^=2^ delims^=^'^" %%G IN ('FINDSTR /R /C:"^[^\"']*user_pref[ 	]*\([ 	]*[\"'][^\"']*[\"'][ 	]*," user.js') DO (`
The FINDSTR returns lines with instances of `user_pref`. The only caveat to this weird-looking regex pattern is that it skips instaces of `user_pref` that have any quote or double-quote **before** `user_pref` (like` // "this" user_pref`). I had to do it like that to make sure the FOR only extracts prefnames (because of the tokenization thing). This simply means that lines skipped this way are not cleaned up (unless the same prefname turns up again in a valid way in the user.js).

`IF NOT ""=="%%G" (SET "[%%G]=1")`
Set var [*prefname*] to 1, so it is defined.

`FOR /F "tokens=1,* delims=:" %%G IN ('FINDSTR /N "^" prefs.js') DO (`
Parse prefs.js. This one's the same as before.

`FOR /F tokens^=1^,2^ delims^=^" %%I IN ("%%H") DO (`
Extract the prefname in that prefs.js line. I removed all the checks for syntax from it because prefs.js is not meant to be edited by hand anyway, and this algorithm will only cleanup preferences that were defined as variables above anyway.

```
IF NOT DEFINED [%%J] (
		ECHO:%%H
```
If [*prefname*] is not defined, echo *line*. The [ ] are there so it will print the line even if it happens to be empty, otherwise the IF DEFINED syntax would go bananas in that scenario.

> But is that an "AND" (&) in the 1st line? Could @ECHO OFF theoretically fail? If so, wouldn't it be maybe better to do it in 2 lines?

I'm not aware of any way it could fail. I just got into the habit of declaring the first (global) SETLOCAL like that so as to save one line. If you look at the code it could even do without the SETLOCAL - ENDLOCAL clause, but I'd rather leave it there to respect the environment in which the script is run (when execution ends the ENDLOCAL will return any variables to their previous value if they had one before the script started running).

As for the _new time thing_, I'm also on Win 7, but somehow it seems the formatting can include spaces sometimes. It may have to do with language packs (I have one here, even though I don't need it).

> just FYI, I do statement analysis as a hobby and the word "really" in there says a lot, mate ;)

I guess I should've said "prove that I don't _actually_ enjoy annoying you". Does your statement-analysis hobby have a special clause for foreign english speakers somewhere, by any chance? :)

claustromaniac:
> empty lines were preserved in v1.2

That was not meant to change. I'll have a second look.

earthlng:
Lol, you little perfectionist :)

I want to mention the new version in the next changelog and wondered what the timing improvement is so that we can mention that too.
On my fairly old machine and in a idle state, with a prefs.js with 769 prefs and a user.js with 630 prefs (including section 9999) the cleanup function in v1.2 takes 3 minutes 37 seconds. In v2.0 it takes **less than a second!!*** HOLY COW man, great job!

claustromaniac:
I know lol I even had to remove the last CLS (https://github.com/ghacksuserjs/ghacks-user.js/pull/402/commits/b0fb59f735b23466824d66878a50fcc76d5a72ca) so I could get to read the console output. 

This started out as an experiment but it turned out well, and I thought it was enough of an improvement to be worth the little downside about the lines containing quotes before user_pref (that I mentioned previously).

I'm likely gonna push more changes soon (the stuff we discussed with @overdodactyl). If you want to merge this in the meantime or not is up to you.

BTW, when did :jeans: give you a promotion? LOL congrats...I guess! :laughing:

Thorin-Oakenpants:
:jeans: discovered that an org can have multiple owners, and after the :hankey: 's attempt to get me banned (and also cuz I'm like 94 and could drop dead any day), this is a prudent move

Obligatory :hankey: link: #323

earthlng:
> BTW, when did :jeans: give you a promotion?

Well, :jeans: is such an asshole sometimes + I guess he fears he could get permanently banned next time :grin: 

> LOL congrats...I guess! :laughing: 

yeah thanks. Promoted from 1 unpaid position to another. Feels bad man :)

> If you want to merge this in the meantime or not is up to you.

If I get to it, I will. It's mostly the cleanup function that takes a while to properly review. Feel free to add more commits if you're ready before I am. And thanks for the breakdown, I'll get back to you about that asap

Thorin-Oakenpants:
I am rather young for a dwarf

earthlng:
re: 1st line - MSDOS is weird. According to [this guy](https://stackoverflow.com/a/8332012) you can do something like `exit & start notepad.exe` ... exit and then do something else?!? weird but whatever

> The only caveat to this weird-looking regex pattern is that it skips instaces of `user_pref` that have any quote or double-quote **before** `user_pref` (like` // "this" user_pref`).

that's fine

> Set var [prefname] to 1, so it is defined.

nice! I was wondering if we may need to worry about prefs like `user_pref("ComSpec",` until I saw that you already wrap the prefname in `[ ]`.

> I removed all the checks for syntax from it because prefs.js is not meant to be edited by hand

this limits its usefulness to strip all user.js prefs from another user.js to see what the 1st user.js is missing (fe. if the 2nd user.js uses single quotes). But it's good enough for our purposes and I can see why you removed it.

> it could even do without the SETLOCAL - ENDLOCAL clause, but I'd rather leave it there to respect the environment in which the script is run

I agree. It's also nice that it clears the environment from all the vars set for the prefs

> Also, this ensures Delayed Expansion is off while running this script, regardless of how it was before.

perfect :+1: 

> As for the *new time thing* ... It may have to do with language packs

Ok, whatever it was - if you saw spaces this is a good fix

> Does your statement-analysis hobby have a special clause for foreign english speakers somewhere, by any chance?

Yes that could be 1 reason. Or ... the speaker could use it as an excuse after being confronted :)

earthlng:
> **When this first loop is over, the script defined a variable for each prefname ...**

1. can we move this loop outside of the `(...)>tempcleanedprefs` for the unlikely event that it spills some output that would end up in prefs.js?

2. `FOR /F tokens^=1^,2^ delims^=^" %%I IN ("%%H") DO (` - we don't use/need token 1

3. `IF NOT DEFINED [%%J] (` - ECHO on the same line? If you agree with (2) also change to `[%%I]`

claustromaniac:
Good timing, I came back a little while ago.

> re: 1st line - MSDOS is weird. According to this guy you can do something like exit & start notepad.exe ... exit and then do something else?!? weird but whatever

Oh I just remembered that I read sometime that, because of that weird behaviour, & is also useful to get data past ENDLOCAL like this:

`ENDLOCAL & SET "_var=%_othervar%"`

> this limits its usefulness to strip all user.js prefs from another user.js to see what the 1st user.js is missing (fe. if the 2nd user.js uses single quotes). But it's good enough for our purposes and I can see why you removed it.

I was mostly talking about the checks to see if the lines in the target file (prefs.js) begin with user_pref and such. I took those away because they were no longer necessary. Making it good for doing what you describe is actually pretty simple, just need to change this line:

> OR /F tokens^=1^,2^ **delims^=^"** %%I IN ("%%H") DO (

...to this:

> OR /F tokens^=1^,2^ **delims^=^"^'** %%I IN ("%%H") DO (

...and then it will catch those too.

> Yes that could be 1 reason. Or ... the speaker could use it as an excuse after being confronted :)

Aw man.. that hurts!

1. Sure thing.
2 & 3. We totally need token 1. If it's not there and token 2 happens to be empty, the FOR loop decides nothing needs to happen and just skips the body of the loop for that iteration (the same reason the script was swallowing empty lines before the latest commit).

earthlng:
> ...and then it will catch those too.

ah nice, I didn't realize it was *that* easy. IDK if we want to support that though. Up to you. God, I love delegating decision-making :)

> We totally need token 1

collaboration is a beautiful thing

claustromaniac:
> God, I love delegating decision-making :)

Not trying to spoil the moment but I wouldn't do that often if I were you. It could prove *dangerous* (I do tend to fuck up).

earthlng:
nah man, don't sell yourself short, you're great. Thanks for the making the last few changes as discussed - this is ready to go then. I'll commit it right away. Great stuff. Thanks mate! :kissing_heart:

Thorin-Oakenpants:
@claustromaniac great work man!! The speed improvement reminds me of some of the early days in SQL (i.e early days of SQL, not early days of me). Data was not denormalized, and we're talking at the time some tens of millions of transactional records - and the one of the analytical data reports required things like averages (items/value/etc) - one value required to do this was days in the month. The inbuilt MS function was slow. A query would take 10 minutes. I eventually wrote and tested about 6 or 7 other ways of doing it, and got down to approx 10 secs. In reality, the thing needed to be data warehoused and denormalized, which is different story.

What you just did here with the speed reminds me of that. You're like a young ME!! :laughing:

Issue #340, opened 2018-01-13
prefsCleaner.bat in Linux [solved: use scratchpad scripts]


1xPdd:
Hey, sorry for my ignorance--can anyone suggest what I would need to do in order to get an equivalent cleaning under Linux? I've run all the scratchpad scripts but my scripting is not good enough to rewrite the Windows Batch file into bash, so I imagine I still have plenty of legacy settings hanging around. Thanks.

claustromaniac:
I may be wrong, but I believe you don't need to run `prefsCleaner.bat` if you already ran all the scratchpad scripts, at least as long as you don't have a custom `user.js` file with preferences from different sources or something like that. 

If you used to use a different `user.js` file or just liked to fiddle a lot with settings yourself, you should probably edit `prefs.js` manually to make a more thorough cleanup. For example, you can make a backup of `prefs.js` (or better, a backup of the entire profile) and then proceed to remove all entries in it that don't have anything to do with the UI or with extensions that you still use, and launch Firefox again to have all settings in `user.js` reapplied. 

I did that recently in an old profile and in my case it didn't take as long as it sounds it could take.

claustromaniac:
By the way, I might be willing to port the script to bash if someone is willing to test it for me, but It could take a while since I would need to do some research and I still have to get used to this shitty netbook's shitty keyboard, because my PC decided to die on me the other day.

1xPdd:
> but I believe you don't need to run prefsCleaner.bat if you already ran all the scratchpad scripts

Ahh, thanks for the clarification. I had thought one needed to do both.

> proceed to remove all entries in it that don't have anything to do with the UI or with extensions that you still use, and launch Firefox again to have all settings in user.js reapplied.

Just so I follow you, you mean, in the **about:config**, clicking reset on everything that is bolded?

>  I might be willing to port the script to bash if someone is willing to test it for me

I can help test, though if I'm the only Linux user here, definitely don't suffer a shitty netbook keyboard just for me.

Thorin-Oakenpants:
> Just so I follow you, you mean, in the about:config, clicking reset on everything that is bolded?

No. Don NOT do that - not everything that is **user set** or **modified** (i.e bolded) is that way because of the user.js or end-user changes - a LOT are internal firefox settings

Thorin-Oakenpants:
[...]

PR #405, opened 2018-04-20
prefsCleaner.sh (for Linux/Mac)

claustromaniac:
[Long overdue](https://github.com/ghacksuserjs/ghacks-user.js/issues/340#issuecomment-357446813) port of the prefsCleaner.bat for Unix-like systems. I was bored and had the time for it so.. why not?

This is my very first shell script ever, and I have only tested it with a minimal Cygwin install, so all feedback is very welcome. I did my best aiming for compatibility and reliability but, being a beginner and all, it would surprise me if this couldn't be improved still.

Bear in mind that I couldn't let go of bash-only syntax completely, so this won't work in other shells.

@1xPdd, I hope you're still up for testing.

[Download link](https://raw.githubusercontent.com/claustromaniac/ghacks-user.js/548f0f7e9840de3425bb798127b104bd11298bfc/prefsCleaner.sh) for the lazy ones.

### DIFF
--- a/prefsCleaner.sh
+++ b/prefsCleaner.sh
@@ -2,7 +2,7 @@
 
 ## prefs.js cleaner for Linux/Mac
 ## author: @claustromaniac
-## version: 1.0b7
+## version: 1.0
 
 ## special thanks to @overdodactyl and @earthlng for a few snippets that I stol..*cough* borrowed from the updater.sh
 
@@ -38,21 +38,19 @@ fClean() {
 	prefs="@@"
 	prefexp="user_pref[ 	]*\([ 	]*[\"']([^\"']*)[\"'][ 	]*,"
 	while read -r line; do
-		if [[ "$line" =~ $prefexp ]]; then
-			if [[ $prefs != *"@@${BASH_REMATCH[1]}@@"* ]]; then
+		if [[ "$line" =~ $prefexp && $prefs != *"@@${BASH_REMATCH[1]}@@"* ]]; then
 			prefs="${prefs}${BASH_REMATCH[1]}@@"
-			fi
 		fi
-	done <<< "`grep -E "$prefexp" user.js`"
+	done <<< "`grep -E \"$prefexp\" user.js`"
 
 	while IFS='' read -r line || [[ -n "$line" ]]; do
 		if [[ "$line" =~ ^$prefexp ]]; then
 			if [[ $prefs != *"@@${BASH_REMATCH[1]}@@"* ]]; then
 				echo "$line"
 			fi
-			continue
+		else
+			echo "$line"
 		fi
-		echo "$line"
 	done < "$1" > prefs.js
 }
 
@@ -60,7 +58,7 @@ echo -e "\n\n"
 echo "                   ╔══════════════════════════╗"
 echo "                   ║     prefs.js cleaner     ║"
 echo "                   ║    by claustromaniac     ║"
-echo "                   ║          v1.0b7          ║"
+echo "                   ║           v1.0           ║"
 echo "                   ╚══════════════════════════╝"
 echo -e "\nThis script should be run from your Firefox profile directory.\n"
 echo "It will remove any entries from prefs.js that also exist in user.js."
@@ -76,9 +74,7 @@ select option in Start Help Exit; do
 			fi
 
 			fFF_check
-			## create backup folder if it doesn't exist
-			mkdir -p userjs_backups;
-			bakfile="userjs_backups/prefs.js.backup.$(date +"%Y-%m-%d_%H%M")"
+			bakfile="prefs.js.backup.$(date +"%Y-%m-%d_%H%M")"
 			mv prefs.js "${bakfile}" || fQuit 1 "Operation aborted.\nReason: Could not create backup file $bakfile"
 			echo -e "\nprefs.js backed up: $bakfile"
 			echo "Cleaning prefs.js..."
--- a/prefsCleaner.sh
+++ b/prefsCleaner.sh
@@ -2,7 +2,7 @@
 
 ## prefs.js cleaner for Linux/Mac
 ## author: @claustromaniac
-## version: 1.0b6
+## version: 1.0b7
 
 ## special thanks to @overdodactyl and @earthlng for a few snippets that I stol..*cough* borrowed from the updater.sh
 
@@ -37,13 +37,14 @@ fClean() {
 	# the magic happens here
 	prefs="@@"
 	prefexp="user_pref[ 	]*\([ 	]*[\"']([^\"']*)[\"'][ 	]*,"
-	while read -r line || [[ -n "$line" ]]; do
+	grep -E "${prefexp}" user.js >/tmp/userjs_updater_temp.dat
+	while read line; do
 		if [[ "$line" =~ $prefexp ]]; then
 			if [[ $prefs != *"@@${BASH_REMATCH[1]}@@"* ]]; then
 			prefs="${prefs}${BASH_REMATCH[1]}@@"
 			fi
 		fi
-	done <<< $(grep -E "${prefexp}" user.js)
+	done < /tmp/userjs_updater_temp.dat
 
 	while IFS='' read -r line || [[ -n "$line" ]]; do
 		if [[ "$line" =~ ^$prefexp ]]; then
@@ -60,7 +61,7 @@ echo -e "\n\n"
 echo "                   ╔══════════════════════════╗"
 echo "                   ║     prefs.js cleaner     ║"
 echo "                   ║    by claustromaniac     ║"
-echo "                   ║          v1.0b6          ║"
+echo "                   ║          v1.0b7          ║"
 echo "                   ╚══════════════════════════╝"
 echo -e "\nThis script should be run from your Firefox profile directory.\n"
 echo "It will remove any entries from prefs.js that also exist in user.js."
@@ -79,10 +80,7 @@ select option in Start Help Exit; do
 			## create backup folder if it doesn't exist
 			mkdir -p userjs_backups;
 			bakfile="userjs_backups/prefs.js.backup.$(date +"%Y-%m-%d_%H%M")"
-			mv prefs.js "${bakfile}"
-			if [ ! $? ]; then
-				fQuit 1 "Operation aborted.\nReason: Could not create backup file $bakfile"
-			fi
+			mv prefs.js "${bakfile}" || fQuit 1 "Operation aborted.\nReason: Could not create backup file $bakfile"
 			echo -e "\nprefs.js backed up: $bakfile"
 			echo "Cleaning prefs.js..."
 			fClean "$bakfile"
--- a/prefsCleaner.sh
+++ b/prefsCleaner.sh
@@ -2,7 +2,7 @@
 
 ## prefs.js cleaner for Linux/Mac
 ## author: @claustromaniac
-## version: 1.0b4
+## version: 1.0b5
 
 ## special thanks to @overdodactyl and @earthlng for a few snippets that I stol..*cough* borrowed from the updater.sh
 
@@ -51,9 +51,9 @@ fClean() {
       if [[ $prefs != *"@@${BASH_REMATCH[1]}@@"* ]]; then
         echo "$line"
       fi
-    else
-      echo "$line"
+      continue
     fi
+    echo "$line"
   done < "$1" > prefs.js
 }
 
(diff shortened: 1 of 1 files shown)


Thorin-Oakenpants:
Make sure you let CHEF-KOCH know, so he can plagiarize it

overdodactyl:
Really appreciate this @claustromaniac! 

I did a quick test (nothing extensive) and it seems to work great!

My only comment/suggestion after a quick run would be about the shebang.  Instead of `#!/usr/bin/env` I think you might want to use `#!/usr/bin/env bash`.

When using the first, I noticed two "problems" on macOS (neither of huge concern, admittedly):

* You have to specify `sh` in the terminal shell when running.  In other words, you can't just drag/drop the file as you normally can.  Without including `sh`, the script won't run.  

* When you do use `sh`, printing is a little off (`-e`'s show up from echo commands):

```
-e 
                   ╔══════════════════════════╗
                   ║     prefs.js cleaner     ║
                   ║    by claustromaniac     ║
                   ║          v1.0b           ║
                   ╚══════════════════════════╝
-e 
This script should be run from your Firefox profile directory.

It will remove any entries from prefs.js that also exist in user.js.
This will allow inactive preferences to be reset to their default values.
-e 
This Firefox profile shouldn't be in use during the process.
```

claustromaniac:
Thanks @overdodactyl. I had my share of doubts about the shebang. I'll change it right away :)

overdodactyl:
Little off topic here (I can open a new issue if that's preferable), but what's your guy's opinion on integrating the cleaners and updaters through some means?  Couple possibilities would be to combine into one script, another would be for the updater to call/execute the cleaner script after it updates to the latest user.js.

claustromaniac:
For me it would be nice if the updater could detect the presence of the cleaner and call it. Maybe with a prompt after the updating is over.

claustromaniac:
Or better yet, a prompt before the updating even begins.

claustromaniac:
> Might be nice to show the prompt even if it isn't present...if that's the case, the script would download the cleaner script, make it executable and eventually run it.

If someone doesn't want to use the cleaner for some reason they would get that prompt continuously, unless we managed to make it happen only once somehow (like with a config file or something). I think I would rather add a message in the updater informing users that it can now launch the cleaner if it's present. It could even be a read-this-while-you-wait sort of message, to make it even less intrusive.

> If you think there might be frequent updates to the cleaner script, the updater could also run a check for version number and install the latest one if necessary.

Frankly, I tried my best to make both this and the windows version of the cleaner as complete and robust as I could from the beginning so I wouldn't have to push many updates in the future, but I don't think that's a bad idea at all. It could be a worthy addition, or at least it wouldn't hurt.

overdodactyl:
>If someone doesn't want to use the cleaner for some reason they would get that prompt continuously, 

Would definitely be a bother to be asked each time, that's a good point.  I do wonder if there are many use cases where one wouldn't want to use the cleaner though?

claustromaniac:
>  I do wonder if there are many use cases where one wouldn't want to use the cleaner though?

You and me, both. But if such use cases exist, I won't want to be pushy :)

claustromaniac:
> I like the idea but why would before be better than after?

It wouldn't be better by itself. I was thinking that if we added a prompt like "Would you like to run the prefsCleaner afterwards?" and I added a parameter to the cleaners so they can run unattended, the users would then be allowed to walk away after all those prompts and leave the thing doing its business.

earthlng:
Oh okay, yeah that makes sense.

earthlng:
- `fClean $bakfile` and the `$1` in `done < $1 > prefs.js` should probably be wrapped in quotes, just in case
- what's the difference between `done <` and `done <<<` ?
- not sure but I think the way you look for string-in-string is kinda risky. There are no separators between prefs, right? For example looking for `a.b.c` would match if there's a different pref `a.b.c.d` AFAICS, no?
- after having added separators, shouldn't `*${BASH_REMATCH[1]}*` also be wrapped in quotes, just in case? not sure exactly how the quotes need to be set but I'd imagine something like `*@@"${BASH_REMATCH[1]}"@@*`. So you would start with `prefs="@@"` and also append it in `prefs=${prefs}${BASH_REMATCH[1]}@@`. IDK if @ is a good separator though. You get the idea, I'm sure. These are all just example and I have no idea if the format is correct so please don't copy it without testing :)

earthlng:
https://stackoverflow.com/questions/229551/string-contains-in-bash

> The `=~` operator already searches the whole string for a match; the `.*`'s here are extraneous.

and

> Test if it does **NOT** contain a string: `if [[ ! "abc" =~ "d" ]]`

this way you don't need the asterisks around `${BASH_REMATCH[1]}` which should make it easier to set the quotes correctly and is also easier to read/understand IMO. Up to you ;)

claustromaniac:
> $bakfile in fClean $bakfile and the $1 in done < $1 > prefs.js should probably be wrapped in quotes, just in case

Yes, in double quotes, so $1 expands. :+1:

> can you refresh my memory what the difference is between done < and done <<< ?

`<` is redirection like in the CMD interpreter. `<<<` is a here string. There's plenty of documentation on that but, the short version as I undestand it is, the former redirects files to stdin whereas the latter redirects strings (the output of `grep` in this case).

> not sure but I think the way you look for string-in-string is kinda risky. There are no separators between prefs, right? For example looking for a.b.c would match if there's a different pref a.b.c.d AFAICS, no?

> after having added separators, shouldn't *${BASH_REMATCH[1]}* also be wrapped in quotes, just in case?

As keen as always, I see. That was a total oversight on my part. Thanks :) I'll look into it.

> Test if it does NOT contain a string: `if [[ ! "abc" =~ "d" ]]`

The problem I saw with that is the prefnames include special characters (most notably DOT), that would have to be escaped (I think). I initially considered using arrays for that, but I figured the overhead would be worse since I would have to loop over the whole array once for each line in prefs.js. Also, the code wouldn't look as clean that way IMO. IDK.

earthlng:
> Yes, in double quotes, so $1 expands.

and to work correctly if there happen to be spaces in the filename

> The problem I saw with that is the prefnames include special characters

good point, I didn't think about that

prefsCleaner.sh
```diff
@@ -0,0 +1,107 @@
...
+  exit $1
+}
+
+fFF_check() {
+  # there are many ways to see if firefox is running or not, some more reliable than others
+  # this isn't elegant and might not be future-proof but should at least be compatible with any environment
+  while [ -e webappsstore.sqlite-shm ]
+  do
+    echo -e "\nThis Firefox profile seems to be in use. Close Firefox and try again.\n"
+    read -p "Press any key to continue."
```
claustromaniac: Not fully convinced about this part. It's all or nothing. Thinking ahead, it might be better to do the check just once (not looping) in case `webappsstore.sqlite-shm` stops being a temporal file someday.

earthlng:
The script doesn't work in my linux vm. Haven't you tested it? I have to say I'm a bit disappointed :)

The problem is that the 1st `while` sees the full grep output as 1 line (= it only resets the 1st pref it finds).
I imagine the regex array has all prefnames because of it and you could either loop over that, or just pipe the grep output into a temp file and delete it again afterwards.

And does it really need the `|| [[ -n "$line" ]]` in the 1st while loop? I don't think so.

This works for me and produces the same prefs.js as the Windows version when ran against the same prefs.js and user.js:
```bash
  grep -E "${prefexp}" user.js >/tmp/userjs_updater_temp.dat
  while read line; do
    ## blah blah
  done < /tmp/userjs_updater_temp.dat
```

earthlng:
`webappsstore.sqlite-shm` only exists if something happened to write to `localStorage` but the absence of it doesn't necessarily mean a running FF isn't using this profile. It's better than nothing I guess.

prefsCleaner.sh
```diff
@@ -70,22 +70,23 @@ select option in Start Help Exit; do
...
 
 			fFF_check
 			## create backup folder if it doesn't exist
 			mkdir -p userjs_backups;
 			bakfile="userjs_backups/prefs.js.backup.$(date +"%Y-%m-%d_%H%M")"
-			mv prefs.js "${bakfile}" && echo -e "\nprefs.js backed up: $bakfile"
+			mv prefs.js "${bakfile}"
+			if [ ! $? ]; then
+				fQuit 1 "Operation aborted.\nReason: Could not create backup file $bakfile"
+			fi
```
earthlng: `mv prefs.js "${bakfile}" || fQuit 1 "Operation aborted.\nReason: Could not create backup file $bakfile"`

?

claustromaniac:
>  Haven't you tested it?

I have, thoroughly. Not a single issue on my end :confused:

> does it really need the || [[ -n "$line" ]] in the 1st while loop?

Hmm. I guess not.

prefsCleaner.sh
claustromaniac: Yeah, I originally wrote it the other way around (if it doesn't fail do stuff). The if clause was a remnant of it. Didn't think it through.

claustromaniac:
Your linux vm is _weird_.

I don't see any reason why the output of grep should be a single line until written into a file, but whatever. I don't mind writing a temp file if that means better compatibility.

earthlng:
Yeah sorry about that. The VM is a very old and never-updated snapshot of an already running archlinux. :man_facepalming: 

But yes I tested all kinds of things and nothing except what I wrote above worked.

claustromaniac:
Alright, I trust you. I'll just archive this X file and move on then.

earthlng:
maybe it's the `<<<`, IDK. I also tried to explicitly set IFS to `\n` but that didn't work either.

claustromaniac:
@overdodactyl I'm using the same subfolder as the updater for the backups (even though the name is **userjs**_backups). What do you think? Would you rather have them in separate folders?

earthlng:
Woohoo! wrapping it in double-quotes works! 
```bash
done <<< "`grep -E "$prefexp" user.js`"
```
works with and without `-r`, too. IDK, is the `-r` necessary?

earthlng:
Damn I'm sooo sorry man, now you already wasted even more time making changes for me! SOOOORRY

claustromaniac:
Nah, don't be sorry. I'd say it's great that you had that issue with the script and not random users, since you could find a solution. Thanks :+1:

earthlng:
I prefer `else` over `continue` and think it's stupid to backup prefs.js files in **userjs**_backups, sorry :)

If you're happy with that, I'm ready to commit it.

earthlng:
Thanks @claustromaniac !

claustromaniac:
Glad to contribute. Thanks for all the input, as always.

earthlng:
np, thanks for your patience

1xPdd:
Hi @claustromaniac  you caught me while real life was giving me a beating. But since I asked for this, I wanted to thank you. I just now ran it on my main profile and, as far as I can tell, it worked great. So thanks. Cheers.

claustromaniac:
@1xPdd,

I hope _real life_ is treating you better now, and I'm glad the script works well for you. Be sure to let me know if you ever have any issues with it or if you think it can be improved. I'll try to make some of the changes discussed here as soon as I manage to escape my current procrastination cycle, so don't be shy.

atomGit:
hello @claustromaniac :)

```bash
fFF_check() {
	# there are many ways to see if firefox is running or not, some more reliable than others
	# this isn't elegant and might not be future-proof but should at least be compatible with any environment
	while [ -e webappsstore.sqlite-shm ]; do
		echo -e "\nThis Firefox profile seems to be in use. Close Firefox and try again.\n"
		read -p "Press any key to continue."
	done
}
```
i don't think this is working - i never see this message if i run the script while FF is running

would something like this be better perhaps? ...
[Bash script to check running process - Stack Overflow](https://stackoverflow.com/questions/2903354/bash-script-to-check-running-process)

rusty-snake:
Confirming:
```
$ ls webappsstore.sqlite*
webappsstore.sqlite  webappsstore.sqlite-wal
```

Alternative I would suggest to use `lock` instead of `webappsstore.sqlite-shm`.

atomGit:
in that case something like this maybe ???...

```bash
iFile=`ls "$sFirefoxDir" | grep '^lock$' | wc -l`
```

earthlng:
I don't have `lock` on Windows. `parent.lock` yes, but I'm not sure if that gets reliably deleted on close.
Maybe `favicons.sqlite-shm` would be a better candidate? Do you guys have that?

rusty-snake:
> Does `lock` get reliably deleted on close?

Yes

> Maybe `favicons.sqlite-shm` would be a better candidate? Do you guys have that?

No

atomGit:
i don't have any *shm files in the profile folder on linux

earthlng:
@rusty-snake are you on linux too, or Mac?

rusty-snake:
Linux

earthlng:
okay, done. Thanks guys

PR #406, opened 2018-04-20
Auto update and save backups to folder

overdodactyl:
* updated shebang

* The script now compares its version number to the one online.  If there is a newer version of `updater.sh` online, it is downloaded and run.

* Backup files are now saved to the directory `userjs_backups` instead of causing more bloat in the profile directory.

### DIFF
--- a/updater.sh
+++ b/updater.sh
@@ -2,11 +2,11 @@
 
 ### ghacks-user.js updater for Mac/Linux
 ## author: @overdodactyl
-## version: 1.3
+## version: 1.1
 
 ghacksjs="https://raw.githubusercontent.com/ghacksuserjs/ghacks-user.js/master/user.js"
 updater="https://raw.githubusercontent.com/ghacksuserjs/ghacks-user.js/master/updater.sh"
-
+update_pref=${1:--ask}
 
 currdir=$(pwd)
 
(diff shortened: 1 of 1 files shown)


earthlng:
I don't like that it does the updating without any prompts. I mean, downloading a file, making it executable and then running it isn't something everyone would be willing to do. The Windows version only does that when the user explicitly runs it with a command line parameter.

What I think would be nice is:
- add support for a command line parameter, possibly 2 (see below)
- if the parameter isn't set, ask (before you do the curl, etc) if the user wants to check for an update to the script. If an update is available maybe ask again if they want to download and run the new version. And in that case pass the 2nd parameter (-dontupdate or something like that) so the new version doesn't check and prompt again.
- if the other parameter is set (fe. -updatescript or whatever you like), skip the prompts and just check for update and download+run it

or at least something along those lines.

claustromaniac:
@earthlng So, basically... default to ask, and two parameters to skip the prompts (one that dowloads and one that doesn't). Did I get that right?

I'm butting in because I might want to bring the windows version closer to this one if @overdodactyl is willing to improve on it. Not saying they should be the same (I mean I know I kinda went overboard with the windows version), but I think it would be nice to reach some common ground where possible and not have a massive wiki page on updaters lol.

I guess a dedicated issue for these things could be of use after all (so I don't have to butt in :zipper_mouth_face:).

overdodactyl:
@earthlng I agree - probably not the best to download and execute a file with explicit say so.

@claustromaniac - I just started working on making the suggested changes.  I'm good with earthlings suggestions if that sounds good with you.

>default to ask, and two parameters to skip the prompts (one that dowloads and one that doesn't). Did I get that right?

I interpreted it to be the same.

It's a bummer we have to write different scripts for Windows and *nux.  Someone helped me convert all my ShadowFox scripts to GoLang which compiles to native executables for each OS. Would be nice here, but not sure it's worth it given how far down the rabbit hole we are haha

overdodactyl:
@earthlng, on second thought, do you think it's necessary to ask the user if they want to check for an update?  To me it seems like this will just be adding an extra prompt a user has to answer without much benefit.  There won't be new updates all that frequently I imagine.  Maybe just do a silent check and ask the user if they want to update when one is available (unless the -dontupdate arg is passed)?

Seemingly, the only benefit to asking the user first would be saving a connection to GitHub, but presumably the user is ok with that anyway since they're updating the user.js

overdodactyl:
Current method used (not difficult to change if you don't like it)

* `updater.sh -donotupdate`
    * Run current updater.sh
* `updater.sh -update`
    * Check for update
    * If update available, backup current and download/execute new
    * No update available, run current updater.sh
* `updater.sh`
    * Check for update
    * If update available, ask user if it should be updated
        * Yes
            * backup current and download/execute new
        * No
            * Run current updater.sh

earthlng:
support case-insens parameters?

bash 4.0: `"${update_pref,,}"`
or more universal: `"$(echo $update_pref | tr '[A-Z]' '[a-z]')"`

I think everyone should have bash v4+ by now, right?

earthlng:
LGTM but I haven't tested it. Have you tested it?

overdodactyl:
>support case-insens parameters?

Good idea! Added.

>I think everyone should have bash v4+ by now, right?

Unfortunately macOS still only comes preinstalled with 3, so it's probably best to stay with compatible with that if possible.

>Have you tested it?

I have, but never bad to have another person give it a once over if you have the time :)

claustromaniac:
I installed a couple of packages I was missing (curl and bc), to help testing.

It seems to work just fine on my end.

overdodactyl:
Thanks for testing it @claustromaniac!

claustromaniac:
Just returning the favor :)

claustromaniac:
After some more testing and looking at the code a bit more, I have one observation: the auto update part is only reliable as long as the name of the updater is `updater.sh`. To have consistent results when the user renames the file, I believe the script should use something like:

`path_to_script="$(dirname "${sfp}")/$0"`

Although, I'm not sure how that should be when reading through symlinks instead. I'm still a newbie :sweat_smile:

overdodactyl:
Hmm good point...I didn't really think about people renaming the file.  

IIRC, `script_filename="$(basename $0)"` should work, but I'll have to test to double check...just going on memory.  

On a similar not, I've been thinking it might be a good idea to change the name of the updater.  It's pretty nondescript and when thrown in w/ the profile directory it might be nice to be `userjs_updater.sh` or something.  Another plus would it would always be located right next to the `user.js` file

claustromaniac:
I got mixed up. Still haven't made my proper research on those variables. I found some quick and dirty "shortcuts" [here](https://stackoverflow.com/questions/192319/how-do-i-know-the-script-file-name-in-a-bash-script).

overdodactyl:
Alright, I *think* that last update fixes the problem.  I used the suggested

`"$(basename "$(test -L "$0" && readlink "$0" || echo "$0")")"`

claustromaniac:
wot :no_mouth: 

Something went wrong with that last commit. It looks exactly like v1.2.

overdodactyl:
Shoot.  I must have run the updater with a lower version number to check the updating portion then committed the downloaded version.  

The erroneous commit has been removed (poof! gone from history) and replaced.

earthlng:
do we really want to support renamed scriptfiles? I thought readlink doesn't exist in MacOS?

re: dependencies - my test VM doesn't have `bc`.
As long as we don't go higher than version x.9, this works without `bc`:
`[ "$current_version" \< "$online_version" ]` = ASCII comparison.
or `[[ "$current_version" < "$online_version" ]]` if you don't like the escaped `<`

`40.0` > `4.0` :+1: 
but `4.10` < `4.9` !

overdodactyl:
I'd be ok removing the last commit regarding the renamed script files.  @claustromaniac, thoughts?

What do you guys think about the idea of renaming `updater` to `userjs_updater` or something similar? Or just keep as-is?

>I thought readlink doesn't exist in MacOS?

Good point...I forgot that won't work on the default mac setup.  We should probably avoid that.

Thanks for the `bc` heads up, I'll switch that to your suggestion

claustromaniac:
> do we really want to support renamed scriptfiles?

If compatibility would be the trade-off I'd say definitely no. Forgive me for even suggesting it (remember I'm the newbie here).

> What do you guys think about the idea of renaming updater to userjs_updater or something similar?

 I like it.

earthlng:
Wanna finish this? I made a couple of changes. If you're happy with that I can merge it.

earthlng:
merged it and updated the wiki: https://github.com/ghacksuserjs/ghacks-user.js/wiki/3.3-Updater-Scripts

thanks @overdodactyl for the update !
thanks @claustromaniac for your input and helping with testing !

overdodactyl:
@earthlng, my apologies for dropping the ball on this - got a little busy with other things and forget to wrap things up here.  Thanks for finishing it!

PR #407, opened 2018-04-24
updater for Windows v4.5

earthlng:
supports commenting-out active user-prefs with the merge function.

It needs a line like this in an override file:
```js
// comment-out "security.tls.version.max" = 1202 - wait until FF enables TLS1.3 for everyone by default
```
such a line has to start at column 0 with a double forward-slash followed by a single space, lowercase `comment` `minus` `out`, single space and the prefname in double quotes. Anything after that is optional and can be used for a reminder comment or whatever.

But maybe we should make the format a bit more uncommon, like with 4 slashes or so.

### DIFF
--- a/updater.bat
+++ b/updater.bat
@@ -213,6 +213,7 @@ FOR /F tokens^=2^,^*^ delims^=^" %%G IN ('FINDSTR /B /R /C:"\/\/ comment-out .*"
 								SETLOCAL EnableDelayedExpansion
 								FOR /F "delims=" %%L IN ("!%%K!") DO (
 									ENDLOCAL & ECHO:user_pref("%%K"%%L
+									SET "%%K="
 								)
 							)
 						)
--- a/updater.bat
+++ b/updater.bat
@@ -3,7 +3,7 @@ TITLE ghacks user.js updater
 
 REM ## ghacks-user.js updater for Windows
 REM ## author: @claustromaniac
-REM ## version: 4.4
+REM ## version: 4.5
 REM ## instructions: https://github.com/ghacksuserjs/ghacks-user.js/wiki/3.3-Updater-Scripts
 
 SET _myname=%~n0
@@ -75,7 +75,7 @@ ECHO:
 ECHO:                ########################################
 ECHO:                ####  user.js Updater for Windows   ####
 ECHO:                ####       by claustromaniac        ####
-ECHO:                ####             v4.4               ####
+ECHO:                ####             v4.5               ####
 ECHO:                ########################################
 ECHO:
 SET /A "_line=0"
@@ -191,8 +191,9 @@ GOTO :EOF
 REM ############ Merge function ############
 :merge
 SETLOCAL DisableDelayedExpansion
+FOR /F tokens^=2^,^*^ delims^=^'^" %%G IN ('FINDSTR /B /R /C:"user_pref.*\)[    ]*;" "%~1"') DO (IF NOT "%%H"=="" (SET "%%G=%%H"))
+FOR /F tokens^=2^,^*^ delims^=^" %%G IN ('FINDSTR /B /R /C:"\/\/ comment-out .*" "%~1"') DO (SET "__unset__%%G=1")
 (
-	FOR /F tokens^=2^,^*^ delims^=^'^" %%G IN ('FINDSTR /B /R /C:"user_pref.*\)[ 	]*;" "%~1"') DO (IF NOT "%%H"=="" (SET "%%G=%%H"))
 	FOR /F "tokens=1,* delims=:" %%I IN ('FINDSTR /N "^" "%~1"') DO (
 		SET "_temp=%%J"
 		SETLOCAL EnableDelayedExpansion
@@ -205,11 +206,14 @@ SETLOCAL DisableDelayedExpansion
 				ENDLOCAL
 				FOR /F tokens^=2^ delims^=^'^" %%K IN ("%%J") DO (
 					IF NOT "_user.js.parrot"=="%%K" (
-						IF DEFINED %%K (
-							SETLOCAL EnableDelayedExpansion
-							FOR /F "delims=" %%L IN ("!%%K!") DO (
-								ENDLOCAL & ECHO:user_pref("%%K"%%L
-								SET "%%K="
+						IF DEFINED __unset__%%K (
+							ECHO://%%J
+						) ELSE (
+							IF DEFINED %%K (
+								SETLOCAL EnableDelayedExpansion
+								FOR /F "delims=" %%L IN ("!%%K!") DO (
+									ENDLOCAL & ECHO:user_pref("%%K"%%L
+								)
 							)
 						)
 					) ELSE (ECHO:%%J)
--- a/updater.bat
+++ b/updater.bat
@@ -191,8 +191,8 @@ GOTO :EOF
 REM ############ Merge function ############
 :merge
 SETLOCAL DisableDelayedExpansion
-FOR /F tokens^=2^,^*^ delims^=^'^" %%G IN ('FINDSTR /B /R /C:"user_pref.*\)[    ]*;" "%~1"') DO (IF NOT "%%H"=="" (SET "%%G=%%H"))
-FOR /F tokens^=2^,^*^ delims^=^" %%G IN ('FINDSTR /B /R /C:"\/\/ comment-out .*" "%~1"') DO (SET "__unset__%%G=1")
+FOR /F tokens^=2^,^*^ delims^=^'^" %%G IN ('FINDSTR /B /R /C:"user_pref[ 	]*\([ 	]*[\"'][^\"'][^\"']*[\"'][ 	]*,.*\)[ 	]*;" "%~1"') DO (SET "[%%G]=%%H")
+FOR /F tokens^=2^,^*^ delims^=^" %%G IN ('FINDSTR /B /R /C:"\/\/\/\/ --- comment-out --- \"[^\"][^\"]*\".*" "%~1"') DO (SET "__unset__%%G=1")
 (
 	FOR /F "tokens=1,* delims=:" %%I IN ('FINDSTR /N "^" "%~1"') DO (
 		SET "_temp=%%J"
@@ -209,11 +209,11 @@ FOR /F tokens^=2^,^*^ delims^=^" %%G IN ('FINDSTR /B /R /C:"\/\/ comment-out .*"
 						IF DEFINED __unset__%%K (
 							ECHO://%%J
 						) ELSE (
-							IF DEFINED %%K (
+							IF DEFINED [%%K] (
 								SETLOCAL EnableDelayedExpansion
-								FOR /F "delims=" %%L IN ("!%%K!") DO (
+								FOR /F "delims=" %%L IN ("![%%K]!") DO (
 									ENDLOCAL & ECHO:user_pref("%%K"%%L
-									SET "%%K="
+									SET "[%%K]="
 								)
 							)
 						)
--- a/updater.bat
+++ b/updater.bat
@@ -191,8 +191,8 @@ GOTO :EOF
 REM ############ Merge function ############
 :merge
 SETLOCAL DisableDelayedExpansion
-FOR /F tokens^=2^,^*^ delims^=^'^" %%G IN ('FINDSTR /B /R /C:"user_pref[ 	]*\([ 	]*[\"'][^\"'][^\"']*[\"'][ 	]*,.*\)[ 	]*;" "%~1"') DO (SET "[%%G]=%%H")
-FOR /F tokens^=2^,^*^ delims^=^" %%G IN ('FINDSTR /B /R /C:"\/\/\/\/ --- comment-out --- \"[^\"][^\"]*\".*" "%~1"') DO (SET "__unset__%%G=1")
+FOR /F tokens^=2^,^*^ delims^=^'^" %%G IN ('FINDSTR /R /C:"^user_pref[ 	]*\([ 	]*[\"'][a-z].*[\"'][ 	]*,.*\)[ 	]*;" "%~1"') DO (SET "[%%G]=%%H")
+FOR /F tokens^=2^,^*^ delims^=^" %%G IN ('FINDSTR /R /C:"^//// --- comment-out --- \"[a-z].*\".*" "%~1"') DO (SET "__unset__%%G=1")
 (
 	FOR /F "tokens=1,* delims=:" %%I IN ('FINDSTR /N "^" "%~1"') DO (
 		SET "_temp=%%J"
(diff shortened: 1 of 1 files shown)


earthlng:
> - support commenting-out active user-prefs with the merge function

... with a line like this in an override file:
```js
// comment-out "security.tls.version.max" = 1202 - wait until FF enables TLS1.3 for everyone by default
```

updater.bat
```diff
@@ -205,11 +206,14 @@ SETLOCAL DisableDelayedExpansion
 				ENDLOCAL
 				FOR /F tokens^=2^ delims^=^'^" %%K IN ("%%J") DO (
 					IF NOT "_user.js.parrot"=="%%K" (
-						IF DEFINED %%K (
-							SETLOCAL EnableDelayedExpansion
-							FOR /F "delims=" %%L IN ("!%%K!") DO (
-								ENDLOCAL & ECHO:user_pref("%%K"%%L
-								SET "%%K="
```
earthlng: I removed `SET "%%K="` because a user.js could have multiple prefs with the same name and without this change, only the 1st instance would get replaced I think, right?

earthlng:
what's perhaps a bit unexpected for users is that it doesn't matter where in the override file such a line exists, at the top, the middle or the bottom - it will always comment out all prefs with that name even if that pref is **below** the comment-out command (unless a user-pref is "hidden" from the merge function when it doesn't start at column 0). 
This is also true with multiple overrides files ie it doesn't matter whether you name a dedicated to-comment-out override file `aaa-to-comment-out.js` or `zzz-to-comment-out.js`. **If** one of your other override files has a non-hidden pref that matches the name of a to-comment-out pref, it **will** be commented out.
for example:
```js
/* my overrides */
// comment-out "security.tls.version.max" = 1202 - wait until FF enables TLS1.3 for everyone by default
user_pref("security.tls.version.max", 3); // 1202
```
user.js after merge:
```js
/* 1202: control TLS versions with min and max
 * ... ***/
//user_pref("security.tls.version.max", 4); // 4 = allow up to and including TLS 1.3

/* my overrides */
// comment-out "security.tls.version.max" = 1202 - wait until FF enables TLS1.3 for everyone by default
//user_pref("security.tls.version.max", 3); // 1202
```
- it will comment out the original line in the user.js **and** retain its value, comment, etc
- will append the comment-out command line as a reminder (we could also drop it but I prefer to keep it)
- **AND** will also comment out the user-pref line that is **below** the comment-out command

IMO that's the way it should be because it doesn't make sense to comment-out a pref and then set it again.

updater.bat
```diff
@@ -191,8 +191,9 @@ GOTO :EOF
 REM ############ Merge function ############
 :merge
 SETLOCAL DisableDelayedExpansion
+FOR /F tokens^=2^,^*^ delims^=^'^" %%G IN ('FINDSTR /B /R /C:"user_pref.*\)[    ]*;" "%~1"') DO (IF NOT "%%H"=="" (SET "%%G=%%H"))
+FOR /F tokens^=2^,^*^ delims^=^" %%G IN ('FINDSTR /B /R /C:"\/\/ comment-out .*" "%~1"') DO (SET "__unset__%%G=1")
```
claustromaniac: For my personal use, I would probably relax a bit the format requirements (allowing multiple spaces/tabs, single quotes and such) but I understand why you chose not to. This next bit does not look pretty, but provides better validation of the syntax (not sure why I didn't write it like this from the beginning):

```
FOR /F tokens^=2^,^*^ delims^=^'^" %%G IN ('FINDSTR /B /R /C:"user_pref[    ]*\([    ]*[\"'][^\"'][^\"']*[\"'],.*\)[    ]*;" "%~1"') DO (SET "[%%G]=%%H")
FOR /F tokens^=2^,^*^ delims^=^" %%G IN ('FINDSTR /B /R /C:"\/\/ comment-out [^\"]*[\"][^\"][^\"]*[\"].*" "%~1"') DO (SET "__unset__%%G=1")
```

Notice the brackets around the prefname in the first loop. Lines 212 and 216 would need brackets too in this case:

212: `IF DEFINED [%%K] (`
216: `SET "[%%K]="`

claustromaniac:
Wait!

Since we have our roles somewhat inverted this time, I should've said something more earthlng-esque like "do we really need [insert questionable bit here]?" or "[insert questionable bit here] is stupid".

Meh... too late now.

claustromaniac:
I noticed it would be fairly easy to adapt this to also allow the merge function to fully ignore user-defined preferences (the same way it ignores `_user.js.parrot`) just by adding 1 line. Example:

```
FOR /F tokens^=2^,^*^ delims^=^'^" %%G IN ('FINDSTR /B /R /C:"user_pref[    ]*\([    ]*[\"'][^\"'][^\"']*[\"'],.*\)[    ]*;" "%~1"') DO (SET "[%%G]=%%H")
FOR /F tokens^=2^,^*^ delims^=^" %%G IN ('FINDSTR /B /R /C:"\/\/\/\/ comment-out [\"][^\"][^\"]*[\"].*" "%~1"') DO (SET "__unset__%%G=1")
FOR /F tokens^=2^,^*^ delims^=^" %%G IN ('FINDSTR /B /R /C:"\/\/\/\/ ignore [\"][^\"][^\"]*[\"].*" "%~1"') DO (SET "__ignore__%%G=1")
(
	FOR /F "tokens=1,* delims=:" %%I IN ('FINDSTR /N "^" "%~1"') DO (
		SET "_temp=%%J"
		SETLOCAL EnableDelayedExpansion
		IF NOT "!_temp:~0,9!"=="user_pref" (
			ENDLOCAL & ECHO:%%J
		) ELSE (
			IF "!_temp:;=!"=="!_temp!" (
				ENDLOCAL & ECHO:%%J
			) ELSE (
				ENDLOCAL
				FOR /F tokens^=2^ delims^=^'^" %%K IN ("%%J") DO (
					IF NOT DEFINED __ignore__%%K (
						IF DEFINED __unset__%%K (
							ECHO://%%J
						) ELSE (
							IF DEFINED %%K (
								SETLOCAL EnableDelayedExpansion
								FOR /F "delims=" %%L IN ("!%%K!") DO (
									ENDLOCAL & ECHO:user_pref("%%K"%%L
									SET "%%K="
								)
							)
						)
					) ELSE (ECHO:%%J) 
```

Then we could add the following line near the top of user.js in this repo, maybe right above the first `_user.js.parrot` line:

`//// ignore "_user.js.parrot"`

updater.bat
earthlng: > (allowing multiple spaces/tabs, single quotes and such) 

that's just more to explain and document. It's easier to say "if you want this to work, it has to be exactly this 1 format". 

I want to make the format reasonably uncommon so it doesn't accidentally match a comment in someone's user.js.
How about `//// --- comment-out --- "prefname"` ? Any better ideas or something you like?

> Notice the brackets around the prefname in the first loop

good idea

BTW for some unknown reason the TABs in the regexes can somehow get replaced with spaces!
I already had to fix the prefsCleaner.bat. Looks like it was your [commit](https://github.com/ghacksuserjs/ghacks-user.js/commit/fdac8fba6a59d05800dd5a4cd0605fec764136ae) that fucked it up so it's not just on my end.
I noticed it in this PR on line 194. I think I cut and pasted that line within the github online editor and it turned out that way.
Just something to keep in mind when touching lines like that. Sucks that `\t` is not supported in FINDSTR regex.

claustromaniac:
Giving the merge function new superpowers made me think that it might be a decent idea to add a user-overrides.js template to the repo, with examples and such. It may also make things slightly more intuitive to newcomers.

updater.bat
earthlng: big No-No to this part: comment-out **[^\\"]*** - that's waaay too relaxed IMO. I like the rest of it though.
This is neat: `[^\"][^\"]*` - nice workaround for the not supported `+` :+1:

earthlng:
Damn mate, I swear I just had the exact same idea. I wonder if our periods are also synced-up already? :grin:

updater.bat
earthlng: > This one allows any number of spaces and tabs around the key word but requires 4 slashes

more of those shitty tabs? hell no. I like this: `\/\/\/\/ --- comment-out --- \"[^\"][^\"]*\".*`

claustromaniac: > Any better ideas or something you like?

Maybe...`//// merge-comment-out "prefname"`

> BTW for some unknown reason the TABs in the regexes can somehow get replaced with spaces!

Thanks for the heads up. Yeah, I noticed GitHub is rather nosy when it comes to those characters. I don't know if it's a sort of bug or just a side effect of the indenting options (that appear on top when editing the file).

> big No-No to this part

It _isn't_ more relaxed than your original approach though, if you look carefully. It's just about the same up to that point (not my favorite pattern anyway, so forget about it).

claustromaniac: Didn't get to see your previous comment before submitting mine. I'm OK with the last one you posted.

earthlng: > It isn't more relaxed than your original approach though

yeah my initial version was way too relaxed

> I'm OK with the last pattern you posted.

great, let's go with that then. should be uncommon enough

earthlng:
> allow the merge function to fully ignore user-defined preferences

it's already easy to let the merge function ignore/not-merge certain prefs. We list 3 different ways in the wiki. I also think this would maybe make it too confusing, especially for noobs. I mean what would users even want to ignore? We already tell people to use "our" debug prefname in their override files. And add to that, that we might not need those prefs for much longer. A bunch of syntax errors will be output on the console, I think starting in FF60 but not sure, and the parsing will continue, making the debug prefs a lot less useful already. You have to break the syntax pretty hard to stop the parsing. If they ever catch and handle those hard-breaks as well, the debug prefs will be completely useless and we will remove them. Other than debug prefs I don't see much use for what you're proposing.

> add the following line near the top of user.js in this repo

I'd rather define it at the top of the function. Assuming we want to support `ignore` at all.

earthlng:
I made the changes and tested it again but it doesn't work at all. Doesn't seem to like `[^\"'][^\"']*`. Not even `[^\"']*` works. It's weird because the prefsCleaner has a very similar regex and that works. I have no idea why it works in one and not the other. Yes one has the `/B` but even removing that and adding a `^` in front of `user_pref` doesn't work.
Should we just go with this?:
```dos
('FINDSTR /B /R /C:"user_pref[ 	]*\([ 	]*[\"'].*[\"'][ 	]*,.*\)[ 	]*;" "%~1"')
```
and for the `comment-out`:
```dos
('FINDSTR /B /R /C:"\/\/\/\/ --- comment-out --- \"[A-Za-z].*\".*" "%~1"')
```
doesn't really need the uppercase A-Z because all prefs start with lowercase a-z. We could do the same for the user_pref line which would automatically treat the parrots as a non-user-pref line and ignore it.

earthlng:
or like this:
```dos
/C:"user_pref[ 	]*\([ 	]*[\"'][a-z][A-Za-z_\.]*[\"'][ 	]*,.*\)[ 	]*;"
```

claustromaniac:
This works:
`FINDSTR /R /C:"^user_pref[ 	]*\([ 	]*[\""'][^\""'][^\""']*[\""'][ ]*,.*\)[ 	]*;" user.js`

I figured escaping the double quotes for the CMD interpreter before escaping them for the FINDSTR should do the trick... what puzzles me is why this was not an issue with the prefsCleaner. :confused:

claustromaniac:
Moreover, if you _do_ escape the double quotes in the prefsCleaner like that, the FINDSTR call fails with an error. Really weird.

earthlng:
IDK. are you sure this doesn't split it up into 2 or more search strings? for example when I do this:
`/C:"^user_pref[ 	]*\([ 	]*[\"'][a-z][^\""']*[\"'][ 	]*,.*\)[ 	]*;"`
I get a FINDSTR no search string error.

claustromaniac:
> are you sure this doesn't split it up into 2 or more search strings?

Given how arcane the CMD parser is, no. I'm not sure. BUT:
1. doubling double-quotes is supposedly the way to escape double quotes within duoble quotes for the CMD parser.
2. IIRC multiple search strings are normally defined using spaces.
3. I tested it, and the output is fine.

Still, this should be more reliable:
`FOR /F usebackq^ tokens^=2^,^*^ delims^=^'^" %%G IN (`FINDSTR /B /R /C:"user_pref[ 	]*\([ 	]*[\""'][^\""'][^\""']*[\""'][ 	]*,.*\)[ 	]*;" "%~1"`) DO (SET "[%%G]=%%H")`

I should've added the usebackq option in the prefsCleaner too, in fact.

claustromaniac:
Great, even GitHub is trolling me. How am I supposed to escape backquotes here?

earthlng:
so usebackq runs the FINDSTR in a subshell and the cmd doesn't need to be in single quotes anymore, right?
Ok I trust you know what you're doing so let's go with that. Can we ditch /B and use `^` instead? that's 2 bytes less :) And what about limiting it to prefs that start with `[a-z]`. That would save us another 2 lines.

For the `comment-out` I'd just change it to single-quotes to spare us all the double-quote trouble.

And then remains the question about the changed filename. Re-uploading it as userjs_updater.bat would hopefully also get rid of these annoying `^M` github markings on a couple of lines.

earthlng:
I ran this in a test.bat and it still fails:
`FINDSTR /R /C:"^user_pref[ 	]*\([ 	]*[\""'][a-z][^\""']*[\""'][ 	]*,.*\)[ 	]*;" "user.orig.js"`
I suspect yours only works because it has 4 such weird double-double-escapes instead of 3. That suggests to me that it really isn't a proper way to escape it. If it was it wouldn't require an even number of them.

earthlng:
this works for the comment-out but only without usebackq:
```dos
FOR /F tokens^=2^,^*^ delims^=^' %%G IN ('FINDSTR /R /C:"^//// --- comment-out --- '[a-z].*'.*" "%~1"') DO (SET "__unset__%%G=1")
```

claustromaniac:
> That suggests to me that it really isn't a proper way to escape it. If it was it wouldn't require an even number of them.

It suggests a different thing to me: we're probably strolling a minefield of bugs right now. I can't make sense of so many things I've encountered in my tests so far...

claustromaniac:
I read a ton of info online and learned a couple of useful things in the process, but nothing related to this particular issue. I guess we'll have to settle for a simpler pattern after all.

Frankly, all the weird shit I read made me want to throw the thing out the window and rewrite it in VBScript or something. :weary:

Might happen someday.

One of the useful things I learned: FINDSTR apparently accepts parameters in this format too `FINDSTR /RC:"regex" "file"`. That would save an extra couple of bytes.

earthlng:
Yeah I'm quite happy with the way it is atm. The only problem is that the `[a-z]` in the 1st line will drop user_pref lines that don't start with a-z in the prefname. No such pref exists in FF atm but people might have debug prefs like that. So I'm okay with removing that limitation and just leave the `.*` between the quotes.
And just in case some weirdo wants to comment out his weirdly named debug prefs, I'm okay with `'[^'][^']*'` instead of `'[a-z][^']*'`

earthlng:
Thanks @claustromaniac !

I updated the [wiki page](https://github.com/ghacksuserjs/ghacks-user.js/wiki/3.3-Updater-Scripts)

what do you think about supporting user_pref lines with spaces/tabs in front of them, too?

claustromaniac:
This was a nice addition, thank you.

> what do you think about supporting user_pref lines with spaces/tabs in front of them, too?

Sure, as long as there is always **at least one** easy way left to prevent lines from getting merged. I don't even remember why I stopped supporting leading spaces at some point. Probably laziness after making other changes.

Other things to remember that came up during this discussion:

- Adding an user-overrides.js template to the repo.
- Renaming the updater. 
- Weird shit, terrible jokes, etc.

PR #409, opened 2018-04-25
prefsCleaner.sh v1.1 - regex fix

claustromaniac:


earthlng:
wait ..... `+` works in FINDSTR?? that's not documented anywhere

earthlng:
btw I updated the wiki page. Let me know what you think.

what do you think about supporting user_pref lines with spaces/tabs in front of them, too?

PR #447, opened 2018-06-16
Improve updater.sh

ema-pe:
Hi, I've completely rewritten the script `updater.sh` because I think that the old one was [too verbose](http://www.linfo.org/rule_of_silence.html) and did many useless things (but I respect the original author).

Feature of the new version:

- Basic options (see down) handling;
- Use `wget` instead of `curl`, because wget is more widespread (on GNU/Linux is pratically preinstalled);
- It can install `user.js` version according to Firefox version installed on the computer (for example if I have Firefox 59.0, It will download `user.js` for that version and not the latest).

Options available:

-  `-u`, `--update `       First updates the updater script, then the user.js
- `--force-version`, `-f` Force to download the user.js according to the Firefox version
- `--backup`, `-b`        Make a copy of the old user.js before overwriting it
-  `--quiet`, `-q`         Print only errors
- `--verbose`           Print additional informations
- `--help`            Print this message
- `--version`           Print script version

I've tested this script on my computer with Debian 9, it should work on any GNU/Linux distributions. I don't have a Mac, so I ask help for testing it on this platform (but It should work flawlessy). This script needs also a minimal code review.

In short, thanks for all what are you doing, this project is very useful both for documenting `user.js` and improve the privacy on Web.

If this pull requests will be merged I also take care of updating the wiki.

### DIFF
--- a/updater.sh
+++ b/updater.sh
@@ -7,7 +7,7 @@
 #
 # Special requirements (beyond Bash): wget and mktemp.
 #
-# Version: 1.4
+# Version: 2.0
 #
 # Please read the wiki or run 'updater.sh --help' to get informations about this
 # script.
--- a/updater.sh
+++ b/updater.sh
@@ -16,12 +16,15 @@
 
 readonly VERSION="2.0"
 
+# First part of the URL to get updater.sh and user.js.
+readonly REPOSITORY="https://raw.githubusercontent.com/ghacksuserjs/ghacks-user.js"
+
 # Default values for flags.
 QUIET="false"
 VERBOSE="false"
 FORCE_VERSION="false"
-UPDATED="false"
 MAKE_BACKUP="false"
+UPDATED="false"
 
 # Prints a message to the standard error and exit with error code 1.
 error() {
@@ -59,7 +62,8 @@ update_installer() {
     error "Failed to download the updater script."
   fi
 
-  mv "$TMPFILE" "UPDATER.SH"
+  mv "$TMPFILE" "$PROGRAM"
+  chmod u+x "$PROGRAM"
 }
 
 # Prints to the standard output the help message.
@@ -94,14 +98,98 @@ show_version() {
   echo "$PROGRAM for ghacks-user.js version $VERSION"
 }
 
+# Runs the newer script with right command line options given to this script.
+run_newer_updater() {
+  log "Starting newer script..."
+
+  # It recreates the command line options.
+  if [[ "$QUIET" == "true" ]]; then
+    QUIET="--qiet"
+  else
+    QUIET=""
+  fi
+
+  if [[ "$VERBOSE" == "true" ]]; then
+    VERBOSE="--verbose"
+  else
+    VERBOSE=""
+  fi
+
+  if [[ "$FORCE_VERSION" == "true" ]]; then
+    FORCE_VERSION="--force-version"
+  else
+    FORCE_VERSION=""
+  fi
+
+  if [[ "$MAKE_BACKUP" == "true" ]]; then
+    MAKE_BACKUP="--backup"
+  else
+    MAKE_BACKUP=""
+  fi
+
+  "./$PROGRAM" "$QUIET" "$VERBOSE" "$FORCE_VERSION" "$MAKE"_"$BACKUP"
+  exit $?
+}
+
 # Updates the user.js.
 update_userjs() {
   # Run the recently downloader version of this script.
   if [[ "$UPDATED" == "true" ]]; then
-    source "$PROGRAM" $@
-    exit 0
+    run_newer_updater
   fi
-  echo "aggiornato"
+
+  local URL="$REPOSITORY/master/user.js"
+
+  # If '--force-version' is given, it will try to download the user.js version
+  # according to Firefox version installed on the computer.
+  if [[ "$FORCE_VERSION" == "true" ]]; then
+    if [[ $(command -v "firefox") ]]; then
+      local JS_VERS=$(firefox --version | grep -Eo "[[:digit:]]+.[[:digit:]]+")
+      local URL_VERSION="$REPOSITORY/$JS_VERS/user.js"
+
+      wget --quiet --spider "$URL_VERSION"
+      if [[ $? != "0" ]]; then
+        warn "user.js version $JS_VERS is not available"
+      else
+        log "user.js $JS_VERS is available"
+        URL="$URL_VERSION"
+      fi
+    else
+      warn "Firefox binary not found, so download the latest user.js version"
+    fi
+  fi
+
+  # Downloades the user.js file on a temporary file.
+  local DOWNLOADED_USERJS="$(mktemp)"
+  log "Downloading newer user.js to $DOWNLOADED_USERJS..."
+  wget --quiet --output-document "$DOWNLOADED_USERJS" "$URL"
+
+  if [[ $? != "0" ]]; then
+    error "Failed to download the newer user.js. Check internet connection?"
+  fi
+
+  # If '--backup' options is given, it will make a copy of the old user.js if it
+  # is present.
+  if [[ "$MAKE_BACKUP" == "true" ]]; then
+    if [[ -e "user.js" ]]; then
+      log "Copying old user.js to user.js.old"
+      mv "user.js" "user.js.old"
+    else
+      warn "Old user.js not found, so no backup"
+    fi
+  fi
+
+  # Finally move downloaded user.js as the newer user.js and append
+  # user-overrides.js if it is present.
+  log "Move $DOWNLOADED_USERJS as user.js"
+  mv "$DOWNLOADED_USERJS" "user.js"
+
+  if [[ -e "user-overrides.js" ]]; then
+    log "Append user-overrides.js to user.js"
+    cat "user-overrides.js" >> "user.js"
+  else
+    log "No user-overrides.js file found"
+  fi
 }
 
 # Check if a program is installed. If it is not installed prints an error.
--- a/updater.sh
+++ b/updater.sh
@@ -218,8 +218,15 @@ while :; do
     --help|-h) shift;             SHOW_HELP="true";;
     --version|-v) shift;          SHOW_VERSION="true";;
     --backup|-b) shift;           MAKE_BACKUP="true";;
+
+    # Deprecated options.
+    -donotupdate) shift;          warn "'-donotupdate' is a deprecated option";;
+    -update) shift;               warn "'-update' is a deprecated option";;
+
+    # Special cases.
     "")                           break;; # Default case: no more options.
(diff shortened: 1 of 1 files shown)


ema-pe:
Hi, I added backward compatibily for old options `-donotupdate` and `-update`. I also added a prefix for messages, based on the type (log, error or warning).

Thorin-Oakenpants:
@ema-pe : Hey, thanks for all the hard work, much appreciated :+1: . Sorry no-one's replied yet (the weekend I guess). This is not my area, so I'll leave it up to @earthlng & co.

aesthicc:
Works on my mac. 👍

earthlng:
Hi, thanks for this. I like your coding style, it's very clean. :+1: 

> the old one was too verbose

that's certainly something that can be improved

> did many useless things

you mean the "check if a new version of updater.sh is available" and/or the "please compare versions" part with `sed`? For the former I think it's nice that there's some feedback when a new version is available instead of just always downloading the latest version. For the latter, that's just how @overdodactyl decided to do it and I don't have a problem with that.

> Feature of the new version:
> - Basic options (see down) handling;

I like that

> - Use wget instead of curl, because wget is more widespread 

I think curl is pretty widespread and common, too, isn't it? maybe more so than wget, IDK.
Either way I don't care which one we use.

> - It can install user.js version according to Firefox version installed on the computer

nice. But some linux distros use re-branded FF's like Iceweasel or whatever it's called.
Since Linux and Mac don't have portable FF versions available (AFAIK), it'd be nice to parse the profiles.ini file to get the profile directory. If more than 1 profiles are detected it could prompt the user to select which one he wants to update. And maybe it's even possible to also detect from the profile path name whether Firefox or Iceweasel or whatnot are in use and use the correct binary name for the version check.

>  I added backward compatibily for old options `-donotupdate` and `-update`

very nice, it definitely needs that. In addition to the warnings it could set `UPDATE_INSTALLER` to true when the old `-update` parameter is used.

---

now to some of the problems with this.
1. you completely ignored the profile detection. Your script will probably only work correctly if you `cd` into the profile dir first. I mean we didn't include that ugly AF part just for shits and giggles. It should be possible to run the script from any working directory, with absolute or relative paths; and also with soft and hard links.
2. user.js backups should be default true
3. backups in the current script are placed in a subfolder and with a timestamp to allow for multiple backups
4. the updater itself was also backed up when a new version was detected+downloaded+installed - this allows for an easy way to see exactly what changed in the new version
5. not super important but the current version also has case-insens parameters (which I kinda like)
6. your `mv`'s don't automatically overwrite existing files

I don't really know how to proceed here. @overdodactyl what do you think?

aesthicc:
Having both a wget and curl version seems plausible if most GNU/Linux distros come with wget. I like the verboseness of the original, i also like the -f option on this one. I don't really mind using either.

earthlng:
> To be completely honest I would rather see the current updater script built upon/improved where users have issues with it as opposed to changing to a complete re-write.

me too

ema-pe:
Thanks @earthlng and @overdodactyl for the feedback and code review.

On most GNU/Linux distributions `wget` is pratically preinstalled. Look a these statistics about Debian installations of `wget` and `curl`:

* [`wget`](https://qa.debian.org/popcon.php?package=wget): 198491;
* [`curl`](https://qa.debian.org/popcon.php?package=curl): 92662.

I don't know that `curl` is preinstalled on Mac systems, but it is not difficult to ad the support for it.

> nice. But some linux distros use re-branded FF's like Iceweasel or whatever it's called.
Since Linux and Mac don't have portable FF versions available (AFAIK), it'd be nice to parse the profiles.ini file to get the profile directory. If more than 1 profiles are detected it could prompt the user to select which one he wants to update. And maybe it's even possible to also detect from the profile path name whether Firefox or Iceweasel or whatnot are in use and use the correct binary name for the version check.

@earthlng you're right. In this script I assume that people are using a normal Firefox version. A solution can be to manually add all known Firefox forks (Iceweasel, Waterfox, ettc.), they are not many.

> you completely ignored the profile detection. Your script will probably only work correctly if you cd into the profile dir first. I mean we didn't include that ugly AF part just for shits and giggles. It should be possible to run the script from any working directory, with absolute or relative paths; and also with soft and hard links.

Yes, this script only work if the user has placed correctly the script and run it from the right directory. I did this because I was thinking that the user should know how to use the script, but if it necessary I can add an automatic profile detection.

> user.js backups should be default true

You're right, I'll fix it.

> backups in the current script are placed in a subfolder and with a timestamp to allow for multiple backups
I was thinking a lot about this behaviour and there are three cases:

1. If the user want a up-to-date version of `user.js` he won't use `--force-version` option, so the script download the latest version every time. In this case it can be useful report that to the user.
2. If the user use always `--force-version` and it already has the `user.js` version according to his Firefox version. The script will download every time the same `user.js`.
3. If the user use always `--force-version` and it has a previous version of `user.js` (for example 59 -> 60). The script download the `user.js` version according to his Firefox verions. In this case it can be useful to report that to the user.
Ok for multiple backups, but why add a timestamp on the file name? It not enough see the creation date from file properties with any file browser?

> the updater itself was also backed up when a new version was 
detected+downloaded+installed - this allows for an easy way to see exactly what changed in the new version

You're right.

> not super important but the current version also has case-insens parameters (which I kinda like)

Usually *nix commands are case sensitive. But it should not be difficult to add this function to the script.

> your mv's don't automatically overwrite existing files

What `mv` are you talking about? I didn't undestand.

> Partially a matter of preference, but I like to know when/what's going on with a script. We could certainly improve on it as earthling mentioned.

@overdodactyl you can use `--verbose` option to get more informations, but by default it is better to suppress any informations. It is the standard *nix way.

> We had previous discussions about whether or not the script should automatically download the latest version of the updater, should be notified when a new one is available, just be a silent background check etc, and we landed on what it currently is...personally I like that portion of the script, and think it's in the users best interest.

I was thinking a lot about this behaviour and I ended with three cases:

1. If the user want a up-to-date version of `user.js` he won't use `--force-version` option, so the script download the latest version every time. In this case it can be useful notify that to the user.
2. If the user use always `--force-version` and it already has the `user.js` version according to his Firefox version. The script will download every time the same `user.js`.
3. If the user use always `--force-version` and it has a previous version of `user.js` (for example 59 -> 60). The script download the new `user.js`. In this case it can be useful to report that to the user.

> To bI was thinking a lot about this behaviour and there are three cases:

1. If the user want a up-to-date version of `user.js` he won't use `--force-version` option, so the script download the latest version every time. In this case it can be useful report that to the user.
2. If the user use always `--force-version` and it already has the `user.js` version according to his Firefox version. The script will download every time the same `user.js`.
3. If the user use always `--force-version` and it has a previous version of `user.js` (for example 59 -> 60). The script download the `user.js` version according to his Firefox verions. In this case it can be useful to report that to the user.e completely honest I would rather see the current updater script built upon/improved where users have issues with it as opposed to changing to a complete re-write.

Initally I wanted to improve directly your script, but I found a complete different style of scripting; so I decided to do a re-write.

**Conclusions and to-do list for the next days:**

- [ ] Use `curl` if `wget` is not available;
- [ ] Add other Firefox forks binary names for getting FF's version;
- [ ] Allow the script to parse case insensitive options;
- [ ] Make `user.js` backup as a default behaviour;
- [ ] Add multiple `user.js` backups (based on timestamps);
- [ ] Make `updater.sh` backup automatically (as default behaviour);
- [ ] Add profile detection (this task means that the script can operate on different profiles);
- [ ] Avoid overwrite the same `user.js` and `updater.sh` file every time they are updater even if there are not real updates;
- [ ] Improve verbosity when `--verbose` is given.

ema-pe:
@earthlng my script was only an initial draft. If you want to keep the old script I'm happy to improve it instead of a re-write.

overdodactyl:
Personally, my vote would still be to keep the current one and open a new issue with suggested changes/feature additions that we can discuss (I'd be glad to help implement them) and/or create a new PR with alterations to the original to address any desired changes.

ema-pe:
@overdodactyl I have not problem with your way.

earthlng:
> @earthlng my script was only an initial draft. If you want to keep the old script I'm happy to improve it instead of a re-write.

that would be great, thanks

> - [ ] Add profile detection (this task means that the script can operate on different profiles);

maybe something like this:
```bash
PROFILE_PATH=""

getProfilesIni () {
  local f1=~/.mozilla/firefox/profiles.ini
  local f2=~/Library/Application\ Support/Firefox/profiles.ini
  if [[ -f "$f1" ]]; then
    echo "$f1"
  elif [[ -f "$f2" ]]; then
    echo "$f2"
  else
    echo ""
  fi
}

readIniFile () {
  local dummyfile=$(mktemp)

  if [[ $(grep '^\[Profile' "$inifile" | wc -l) == "1" ]]; then ### only 1 profile found
    grep '^\[Profile' -A 4 "$inifile" | grep -v '^\[Profile' > $dummyfile
  else
    grep -E -v '^\[General\]|^StartWithLastProfile=|^IsRelative=' "$inifile"
    echo ""
    read -p "Please select the profile number ( 0 for Profile0, 1 for Profile1, etc ) : " -n 1 -r
    echo -e "\n"
    if [[ $REPLY =~ ^[0-9]$ ]]; then
      grep '^\[Profile'$REPLY -A 4 "$inifile" | grep -v '^\[Profile'$REPLY > $dummyfile
      if [[ "$?" != "0" ]]; then
        echo "[Error] Profile$REPLY does not exist!" && exit 1
      fi
    fi
  fi
  
  local profpath=$(grep '^Path=' $dummyfile)
  local pathisrel=$(grep '^IsRelative=' $dummyfile)

  rm "$dummyfile"

  if [[ ${pathisrel#*=} == "1" ]]; then
    PROFILE_PATH="$(dirname "$inifile")/${profpath#*=}"
  else
    PROFILE_PATH="${profpath#*=}"
  fi
  
}

readonly inifile="$(getProfilesIni)"

if [[ "$inifile" == "" ]]; then
  echo "profiles.ini not found! Using scriptdir"
  
  ## get the full path of this script (readlink for Linux, greadlink for Mac with coreutils installed)
  local sfp=$(readlink -f "${BASH_SOURCE[0]}" 2>/dev/null || greadlink -f "${BASH_SOURCE[0]}" 2>/dev/null)
  
  ## fallback for Macs without coreutils
  if [ -z "$sfp" ]; then sfp=${BASH_SOURCE[0]}; fi

  PROFILE_PATH="$(dirname "${sfp}")"
else
  readIniFile
fi

## change directory to the selected Firefox profile directory so that we can use relative paths fe. mv user.js user.js.old
cd "$PROFILE_PATH" || exit 1

## rest goes here ...
```
feel free to improve :)

ema-pe:
I'm starting to make single pull requests in the next days, improving the existend `updater.sh` instead of writing a new one.

PR #476, opened 2018-08-09
relax 0850a (location bar autocomplete)

earthlng:
IMO this is a pretty fundamental and awesome feature for most people. also see https://github.com/ghacksuserjs/ghacks-user.js/issues/438

### DIFF
--- a/user.js
+++ b/user.js
@@ -531,10 +531,10 @@ user_pref("browser.urlbar.speculativeConnect.enabled", false);
  *   - If *ANY* of the suggestion types are true, 'autocomplete' must also be true
  * [SETTING] Privacy & Security>Address Bar>When using the address bar, suggest
  * [WARNING] If all three suggestion types are false, search engine keywords are disabled ***/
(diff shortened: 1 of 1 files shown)


earthlng:
IMO we can also remove 'browser.urlbar.autoFill.typed' because 'browser.urlbar.autoFill' disables '.typed' as well and people who want autofill probably also want typed stuff autofilled. (removing it instead of moving to 9999 for the FF62 diffs)

Thorin-Oakenpants:
we're jumping the gun a little here - we have a couple of issues open - sharing overrides and a discussion/vote (only as an indicator, not binding) re cookies, etc

We haven't decided what we're doing with the master & extra stickies/js-examples re do we make the master hardened and provide relaxed items, or more relaxed and provide hardened items. And I would consider this item (location bar dropdown etc) as only a small part of a bigger set to do with shoulder surfers and to some degree local data (eg history, forms etc) - and we should relax all of those "together" IMO

PS: created a separate issue for autofill

Issue #439, opened 2018-05-26
vote: change cookies from block all to allow 1st party


Thorin-Oakenpants:
proposal
- `2701`: `network.cookie.cookieBehavior` : currently 2, block all: suggested 1, allow only first party
- `2800s`: change the two cookie clearing prefs to true

reason
- website breakage (see story below)
- makes the user.js more user friendly (1 less thing to unbreak a site, or get logins to work etc)
- its simple enough for "hardened" users to add to their overrides
- we expect users to be using an extension or some other mechanism that suits them - there are so many options eg uMatrix (allows in = no web breakage, but blocks it going out), or C-AD, or containers, or TC, cleaning mechanisms and so on. And users can still set site exceptions for block and session only if they want.

issues:
- none. It's just the default we present in the template. It's a choice between my more hardened setup (default deny and allow exceptions, and I don't clear on close because I like my 4 or 5 sites auto-logins etc, and I don't even need an extension, well not quite true: session cookies used only to make a site work are still usefully blocked in uM) vs a more relaxed approach which requires cleaning (not just cookies but localStorage and IDB - on close is fine, its more about in-session).

I don't think my strategy would be the most popular. This is a fairly simple step to take in making the user.js more "mainstream" which can be supplemented with `relaxed` and `hardened` overrides (in the making).

story:
- I just wasted 15 minutes trying to get imgur to display a godamn image - pissed around with uM and uBO and decentraleyes and even ClearURLs. Then did the troubleshooting trick in a new profile and it came down to allowing a cookie to be *sent* to me. So I can see people getting frustrated. Not that I will change my strategy just yet. 

edit: FYI here was the imgur link - https://imgur.com/a/PPsbbKV . Its an album. Never had a problem with imgur before. Now it seems I need to let them send me a cookie (at least for albums).

:+1: or :-1:

Thorin-Oakenpants:
OT probably: maybe its because cookies are worth so much - [medium article](https://medium.com/@thezedwards/the-cookie-exchange-how-gdpr-created-a-new-market-for-pseudonymized-cookies-592cbde9b44b?source=linkShare-e55d568b3a27-1527345283) - go on, read it. Maybe its because of GDPR and they want to make sure everyone gets a cookie. IDK. Interesting read, will put you to sleep

earthlng:
[...]

PR #477, opened 2018-08-09
change cookies from block all to allow 1st party

earthlng:
see https://github.com/ghacksuserjs/ghacks-user.js/issues/439


Thorin-Oakenpants:
I'll have to update the wiki as well, but yeah, I'm OK with this .. just let me get the timing right. I think a FYI one off issue might be handy as well

PR #484, opened 2018-08-12
fix -updatebatch epic fail

claustromaniac:
I was only testing you, @earthlng! ... Just testing you!

Yeah... right. :rofl:

Oh, the joys of programming...

Summarising, my previous changes made the script compare itself with itself.. Yeah, brilliant. That's why I shouldn't write code when I'm sleepy. It's one of those lessons I never truly learn.

I'm sorry to say I can't properly test this, though, because I keep getting error codes 502 and 503 when I try to download the raw from my fork. But it *should* be OK now.... it *should*!!!


earthlng:
Shhhh! nobody saw it ;)

PR #486, opened 2018-08-15
move dom.battery.enabled back to 2500 from 9999

earthlng:


earthlng:
also need to remove it from https://github.com/ghacksuserjs/ghacks-user.js/issues/123

earthlng:
maybe we can drop some of the links, IDK. Feel free to edit as you see fit before merging

PR #504, opened 2018-09-12
infos about default values

earthlng:


earthlng:
we can also remove 2661 IMO. It was added in FF51 with defaults false and never changed since.
Do we really need to enforce it or keep for the info? 2 prefs less (6 lines total) if we remove it

PR #523, opened 2018-10-23
removed, renamed or hidden in v63.0

earthlng:
- 0301a - do you want to add the `[NOTE] Firefox currently checks every 12 hrs  ...` to `0302a` ? The problem is it also checks for updates every time you open/reload about:preferences and in Menu>Help>About Firefox regardless of when the last check was.

- 0513 - removed because follow-on-search is no longer a deletable system addon

- 2703 - do we just remove `3=for n days` or add a [NOTE] that value 3 was removed in FF63 or something?

- `browser.ctrlTab.recentlyUsedOrder` replaces `browser.ctrlTab.previews` but it now defaults to true. No need to list the new one under 5000 IMO

### DIFF
--- a/user.js
+++ b/user.js
@@ -2126,7 +2126,7 @@ user_pref("services.blocklist.update_enabled", true);
 // 0503: disable "Savant" Shield study (FF61+)
    // [-] https://bugzilla.mozilla.org/1457226
 user_pref("shield.savant.enabled", false);
-// 1031: disable favicons in tabs and new bookmarks
+// 1031: disable favicons in tabs and new bookmarks - merged into browser.chrome.site_icons
    // [-] https://bugzilla.mozilla.org/1453751
    // user_pref("browser.chrome.favicons", false);
 // 2030: disable auto-play of HTML5 media - replaced by media.autoplay.default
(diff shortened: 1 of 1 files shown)


Thorin-Oakenpants:
- 0301a - do you want to add the [NOTE]
   * nah, lets remove it
- 0513 - removed
   * yeah, OK
- 2703 - do we just remove 3=for n days
   * see diffs issue, proposed changing the end to `(see 2704-deprecated in FF63+)`
- browser.ctrlTab.previews
   * agreed we don't need to add this

Everything else looks good

earthlng:
2703: yeah either that or something like `[NOTE] option 3 is no longer supported in FF63+` which might be a bit clearer. Whatever you prefer

Thorin-Oakenpants:
heh, already pushed it, but yours is clearer

earthlng:
or "value 3 is ..."?

Thorin-Oakenpants:
hold ya horsies .. am trying to keep this sucker on one line :)

Thorin-Oakenpants:
```
 * 0=until they expire (default), 2=until you close Firefox 3=for n days
 * [NOTE] value 3 is no longer supported in FF63+ (see 2704-deprecated)
```

would have to go to 2 lines I think. Could push (default) to the same line as the pref itself

earthlng:
```js
 * 0=until they expire (default), 2=until you close Firefox, 3=for n days (see 2704)
 * [NOTE] value 3 is no longer supported in FF63+
```

Thorin-Oakenpants:
^^ but 2704 is not in the deprecated section, so our convention is to add `-deprecated` (2 other matches). It's redundant for the number, but imparts info without actually searching

PR #524, opened 2018-10-25
Update README.md

KonoromiHimaries:
+ compressed

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,14 +1,14 @@
-### ![](https://github.com/ghacksuserjs/ghacks-user.js/blob/master/wikipiki/bullet01.png) user.js
+### ![][b] user.js
 A `user.js` is a configuration file that can control hundreds of Firefox settings. For a more technical breakdown and explanation, you can read more on the [overview](https://github.com/ghacksuserjs/ghacks-user.js/wiki/1.1-Overview) wiki page.
 
-### ![](https://github.com/ghacksuserjs/ghacks-user.js/blob/master/wikipiki/bullet01.png) ghacks user.js
+### ![][b] ghacks user.js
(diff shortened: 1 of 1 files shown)


earthlng:
found this:
https://help.github.com/articles/about-anonymized-image-urls/
https://blog.github.com/2014-01-28-proxying-user-images/

IDK what happens to the .io page but I'd just try as you can always revert if necessary.

> do we really need to worry about this?

IDK. might make the readme easier to read offline without the links

btw I didn't realize we embed a 3rd-party license svg. Can we get rid of that or replace with a svg (or preferably a png) hosted in this repo (or "camo" if possible)?

Thorin-Oakenpants:
ok, I'll merge it and see what happens. Will change the MIT to one on here afterwards

Thorin-Oakenpants:
done, also changed all images to use ones in our repo, including MIT Licence. About the only thing that's slightly iffy is the title on .io:

meh

earthlng:
title is ugly AF. And I'm not seeing the images on the .io page (?)

we should probably also use the raw links:
https://raw.githubusercontent.com/ghacksuserjs/ghacks-user.js/master/wikipiki/bullet01.png

Thorin-Oakenpants:
The images never showed before either (well, not recently, but I thought it did at one stage). I can't  exactly remember what happened. We can fiddle with it. When I used `:large_orange_diamond:` it either displayed the text instead on the io page, or it didn't work and by going with images it did. But obviously now it doesn't. So IDK. Title is obviously taken from the first part of the page (maybe that will go away if we use `:large_orange_diamond:` ?

Have yourself a play

earthlng:
the title in _config.yml doesn't seem to work but the images are now showing up

btw while we're at it: "The ghacks user.js is a template, which, as provided, aims to provide .." is pretty hacky with all those commas and the double "provide".
How about ""The ghacks user.js is a template which aims to provide .."

Thorin-Oakenpants:
- changed the svg to use raw, it broke the image, so changed it back, now it's still broken
- .io still has a stupid title
- if you want to fix the MIT image, go ahead

Issue #533, opened 2018-11-13
MOAR stuff to look at


Thorin-Oakenpants:
From @bogachenko 's PR ( #532 )

* NAH: Promotional links need to be removed?
   * https://hg.mozilla.org/integration/mozilla-inbound/rev/1a5fd94651fb
   * user_pref("identity.mobilepromo.android", "");
   * user_pref("identity.mobilepromo.ios", "");
* NAH: Device Storage API
   * user_pref("device.storage.enabled", false);
* NAH: File Handle API
   * https://developer.mozilla.org/en-US/docs/Web/API/File_Handle_API
   * user_pref("dom.fileHandle.enabled", false);
* NAH: disabled warning about:networking
   * user_pref("network.warnOnAboutNetworking", false);
* NAH: new hardware acceleration settings (disabled by default) added
   * user_pref("webgl.force-enabled", false);  // default false
   * user_pref("webgl.msaa-force", false); // default false
   * user_pref("layers.acceleration.force-enabled", false); // default false

From the Big E (revisit these)

* [x] security.tls.version.min - https://github.com/ghacksuserjs/ghacks-user.js/commit/0cc4007eda10a92dadbe8d3548a06a16c21a1d7b
* [x] full-screen-api.enabled - https://github.com/ghacksuserjs/ghacks-user.js/commit/b85e748b532086a28fc69524223bde1e1e6e8c76
* [x] permissions.default.geo - https://github.com/ghacksuserjs/ghacks-user.js/commit/4e42bad6a13ba8a840a61e5807ea0d3ddd162690

From :jeans: (revisit these)

* media.autoplay.default: make inactive and move to 5000s
   * not just yet. In 64 or 65 the default will be a prompt, move it then so there is no disruption to current users
* [x] permissions api & FP'ing - https://github.com/ghacksuserjs/ghacks-user.js/commit/c12eb0fdc6ccd898298ee7094aca5a05fc4d3b42
* [x] browser.eme.ui.enabled - https://github.com/ghacksuserjs/ghacks-user.js/commit/36b90cd5e6bc71a2c214da81060c4d8e2673f69d
* [x] browser.crashReports.unsubmittedCheck.autoSubmit - https://github.com/ghacksuserjs/ghacks-user.js/commit/299a03663fb70673d017f82058e4ba97910f6d54

bogachenko:
Device Storage API 
https://wiki.mozilla.org/WebAPI/DeviceStorageAPI
[...]

PR #538, opened 2018-11-18
update [SETTING] info

Thorin-Oakenpants:


Thorin-Oakenpants:
> the setting can still exist but it's tied to the autofill system addon

I checked it in FF64 nilla profile, with the xpi present

Thorin-Oakenpants:
`browser.crashReports.unsubmittedCheck.autoSubmit` => deprecated, will do separately (added it as an item in #533)

0517 - I put it back, wonder why it didn't show in 64 nilla

> 2701: Privacy & Security>Cookies and Site Data>**Accept/Block cookies and sit*e* data**

Why? It's exactly what I have it as, which is the combo box for "Type blocked"

earthlng:
> Why? It's exactly what I have it as, which is the combo box for "Type blocked"

because the values are 0-4 and 0 is "Accept cookies and site data" which is not in "Type blocked".
But it doesn't really matter because the Content Blocking UI will be totally different again in FF65

Thorin-Oakenpants:
will be changed again in 65. Figures! Anyway, lets not split hairs - I know some prefs are controlled/set by others, but if we went with yours, then it wouldn't match any UI setting. Will fire this thru then.

PR #540, opened 2018-11-18
abort if neither curl nor wget are available

### LINKED ISSUE #537: bash script
Hi, I have a question about your script. 
https://github.com/ghacksuserjs/ghacks-user.js/blob/master/updater.sh

There are 2 mentions of two packages, wget and curl. These are packages that load something (in our case, user.js).
In this script, they are selected to choose from, because there is an "if" condition. The first package is "curl", but if it is not available (not installed or broken?), Then "wget" comes into play. However, what if there are none. ALL. NOT What then? second "if" is not there?
_this is not about debian distributions (for which they are available by default), but about (say) archlinux._
_On it, as far as I remember, you need to manually install it (although I could be wrong, I did not use it for a long time)._

and so it is possible to supplement the conditions on the verification (presence or absence) of these packages? Is there a guru js here?

I am not a guru in js, and maybe I’ll repeat myself I don’t understand. Explain to me?

_I hope I wrote without errors_ :cry:

claustromaniac:
Additionally, hide the output of those checks.

fixes #537

### DIFF
--- a/updater.sh
+++ b/updater.sh
@@ -17,7 +17,7 @@ DOWNLOAD_TO_FILE="curl -O"
 
 # Use wget if curl is not available.
 if [[ -z $(command -v "curl") ]] > /dev/null 2>&1; then
-  if [[ $(command -v "wget") ]] > /dev/null 2>&1; then
+  if [[ $(command -v "wget") ]]; then
     DOWNLOAD_TO_STDOUT="wget --quiet --output-document=-"
     DOWNLOAD_TO_FILE="wget"
   else
--- a/updater.sh
+++ b/updater.sh
@@ -16,9 +16,14 @@ DOWNLOAD_TO_STDOUT="curl -s"
 DOWNLOAD_TO_FILE="curl -O"
 
 # Use wget if curl is not available.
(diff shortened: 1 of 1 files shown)


claustromaniac:
Why does GitHub encourage redundantly redundant redundancy?...

earthlng:
- `[[ $(command -v "wget") ]]` never writes anything to stdout or stderr, does it? I don't think we need the redirection to dev/null
- is `[[ $(command -v "wget") ]]` in every way identical to `[[ -n $(command -v "wget") ]]` ?
If you're not sure either then I think we should use `-n` (or `-z` again and reverse the if-else)

earthlng:
nvm about the 2nd point, `-n` is the default operation

bogachenko:
@claustromaniac you nice guy and did a good job, but  I like this #541  + loading on perl is easy to implement. `perl` is dependent on `base`

earthlng:
closing this in favor of @overdodactyl's new pull request.
Thanks @claustromaniac for immediately jumping into action and creating this PR though. Love you man! :kissing_cat:

PR #541, opened 2018-11-19
updater_overhaul

### LINKED ISSUE #537: bash script

overdodactyl:
In large part a restructuring/re-write of the last version.

* Uses `perl` as a last resort if `curl` and `wget`  are not available (fixes #537)
* Aborts and notifies user if none of the above are installed 
* Better use of functions
* When version numbers are checked, the contents are immediately saved to a temp dir.  This allows us to skip using wget/curl/perl a second time
* Improved messages for users
* Added various font colors for ease of use and aesthetics

Pinging @claustromaniac and @earthlng ... any feedback/opinions would be great!

### DIFF
--- a/updater.sh
+++ b/updater.sh
@@ -1,129 +1,204 @@
 #!/usr/bin/env bash
 
-### ghacks-user.js updater for Mac/Linux
-## author: @overdodactyl, @ema-pe
-## version: 1.4
+## ghacks-user.js updater for macOS and Linux
+
+## version: 1.5
+## Author: Pat Johnson (@overdodactyl)
+## Additional contributors: @earthlng, @ema-pe
 
 ## DON'T GO HIGHER THAN VERSION x.9 !! ( because of ASCII comparison in check_for_update() )
 
-ghacksjs="https://raw.githubusercontent.com/ghacksuserjs/ghacks-user.js/master/user.js"
-updater="https://raw.githubusercontent.com/ghacksuserjs/ghacks-user.js/master/updater.sh"
+#########################
+#    Base variables     #
+#########################
 update_pref=${1:--ask}
+RED='\033[0;31m'
+BLUE='\033[0;34m'
+BBLUE='\033[1;34m' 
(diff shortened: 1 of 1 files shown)


bogachenko:
it's awesome

bogachenko:
it's awesome

earthlng:
wow it looks great, both visually and code-wise.
Please review my commit and let me know if it makes sense and if you're happy with it.

I think you should also move `mkdir -p userjs_backups` into `backup_file()` because currently when the folder doesn't exist and `update_updater()` calls `backup_file()` the move command will probably fail (right?)

overdodactyl:
Commit looks great, thanks! I moved the location of `mkdir -p userjs_backups` and made a few more changes, but nothing that should impact functionality.

earthlng:
okay then, ready to merge. Thanks @overdodactyl

claustromaniac:
Snazinessssssssssssssssssss!

I second bogachenko: this is awesome. Way more elegant than anything I'd usually write TBH :laughing:

Thorin-Oakenpants:
> Screenshot of new printing:

Pants Romance :kiss: you know what I'm talking about

PR #543, opened 2018-11-19
Updater args

overdodactyl:
This PR focuses on making better use of command line arguments for `updater.sh` using `getopts` as well as adding additional options.

I deleted some past comments and will just update this one to prevent spamming people's emails. 

Result of `./updater.sh -h`:

```txt
Usage: ./updater.sh [-h] [-u] [-d] [-s] [-n] [-b] [-m] [-o OVERRIDE]

Optional Arguments:
	-h,		 Show this help message and exit.
	-u,		 Update updater.sh and execute silently.  Do not seek confirmation.
	-d,		 Do not look for updates to updater.sh.
	-s,		 Silently update user.js.  Do not seek confirmation.
	-b,		 Only keep one backup of each file.
	-o OVERRIDE,	 Filename or path to overrides file (if different than user-overrides.js).
			 If given a directory, all files inside will be appended recursively.
			 You can pass multiple files or directories by passing a comma separated list.
				 IMPORTANT: do not add spaces.  Ex: -o file1.js,file2.js,dir1
	-n,		 Do not append any overrides, even if user-overrides.js exists.
	-m,		 Minify resulting user.js file. i.e. remove all comments except header.
```

Args can be written separately or together. Examples:

```txt
./updater.sh -sdm
./updater.sh -sd -o relaxed.js
./updater.sh -s -d
```

Deprecated arguments will still work, but a message like the following will show up in yellow:

```
Warning: command line arguments have changed.
-donotupdate has been deprecated and may not work in the future.

Please view the new options using the -h argument.
```

This should prevent breakage for people who are used to the current form, but could help transition to the more flexible version in this PR.

### DIFF
--- a/updater.sh
+++ b/updater.sh
@@ -309,11 +309,10 @@ update_userjs () {
     cp user.js userjs_diffs/past_user.js
   fi
   backup_file user.js
-    if [ "$OVERRIDE" != "none" ]; then
-    IFS=',' read -ra FILES <<< "$OVERRIDE"
-    for i in "${FILES[@]}"; do
-        add_override "$i"
-    done
+  if [ "$OVERRIDE" != "none" ]; then
+    while IFS=',' read -ra FILE; do
+      add_override "$FILE"
+    done <<< "$OVERRIDE"
   fi
 }
 
--- a/updater.sh
+++ b/updater.sh
@@ -38,7 +38,7 @@ ff_profile="$(dirname "${sfp}")"
 #########################
 
 usage() {                                             
-  echo -e ${BLUE}"\nUsage: $0 [-h] [-u] [-d] [-s] [-n] [-b] [-o OVERRIDE]\n"${NC} 1>&2  # Echo usage string to standard error
+  echo -e ${BLUE}"\nUsage: $0 [-h] [-u] [-d] [-s] [-n] [-b] [-m] [-o OVERRIDE]\n"${NC} 1>&2  # Echo usage string to standard error
   echo -e "Optional Arguments:"
   echo -e "\t-h,\t\t Show this help message and exit."
   echo -e "\t-u,\t\t Update updater.sh and execute silently.  Do not seek confirmation."
@@ -50,6 +50,7 @@ usage() {
   echo -e "\t\t\t You can pass multiple files or directories by passing a comma separated list."
   echo -e "\t\t\t\t IMPORTANT: do not add spaces.  Ex: -o file1.js,file2.js,dir1"
   echo -e "\t-n,\t\t Do not append any overrides, even if user-overrides.js exists."
+  echo -e "\t-m,\t\t Minify resulting user.js file. i.e. remove all comments except header."
   echo -e
   exit 1
 }
@@ -66,7 +67,7 @@ UPDATE="check"
 CONFIRM="yes"
 OVERRIDE="user-overrides.js"
 BACKUP="multiple"
-
+MINIFY="false"
 
 if [ $# != 0 ]; then
   # Display usage if first arguement is -help or --help
@@ -79,7 +80,7 @@ if [ $# != 0 ]; then
     UPDATE="yes"
     legacy_argument $1
   else
-    while getopts ":hudso:nb" opt; do
+    while getopts ":hudso:nbm" opt; do
       case $opt in 
         h)
           usage
@@ -101,6 +102,9 @@ if [ $# != 0 ]; then
           ;;
         b)
           BACKUP="single"
+          ;;
+        m)
+          MINIFY="true"
           ;;
         \?)
           echo -e ${RED}"\nInvalid option: -$OPTARG"${NC} >&2
@@ -292,6 +296,19 @@ update_userjs () {
   fi
 }
 
+remove_comments () {
+  if [ $MINIFY = "true" ]; then
+    sed -n 1,8p user.js >> userjs_temps/no_comments.js # Add header
+    echo "******/" >> userjs_temps/no_comments.js # Add end of comment
+    # Remove comments and empty lines
+    sed -e 's|/\*|\n&|g;s|*/|&\n|g' -e '/\/\*/,/*\//d' -e 's/\s*\/\/ .*$//' -e '/^\s*$/d' -e '/^[[:space:]]*$/d' user.js >> userjs_temps/content.js
+    cat userjs_temps/content.js >> userjs_temps/no_comments.js
+    mv userjs_temps/no_comments.js user.js
+  else
+    return 0
+  fi
+}
+
 #########################
 #        Execute        #
 #########################
@@ -302,5 +319,6 @@ cd "$ff_profile"
 initiate
 update_updater
 confirmation && update_userjs
+remove_comments
 rm -rf userjs_temps
 cd "${currdir}"
--- a/updater.sh
+++ b/updater.sh
@@ -50,7 +50,7 @@ set_wd () {
     ff_profile="$SCRIPT_DIR"
   elif [ "$PROFILE_PATH" = "list" ]; then
     local firefox_dir=""
-    if [ -d $macdir ]; then
+    if [ -d "$macdir" ]; then
       firefox_dir=$macdir
     elif [ -d $nixdir ]; then
       firefox_dir=$nixdir
@@ -80,7 +80,6 @@ usage() {
   echo -e "\t-h,\t\t Show this help message and exit."
   echo -e "\t-p PROFILE,\t Path to your Firefox profile (if different than the dir of this script)"
   echo -e "\t-l, \t\t Choose your Firefox profile from a list"
-  echo -e "\t\t\t Note: This will not work with portable Firefox installations."
   echo -e "\t\t\t IMPORTANT: if the path include spaces, wrap the entire argument in quotes."
   echo -e "\t-u,\t\t Update updater.sh and execute silently.  Do not seek confirmation."
   echo -e "\t-d,\t\t Do not look for updates to updater.sh."
@@ -176,8 +175,7 @@ get_updater_version () {
 #   -update: Check for update, if available, execute without asking
 update_updater () {
   if [ $UPDATE = "no" ]; then
-    # User signified not to check for updates
-    return 0
+    return 0 # User signified not to check for updates
   fi
 
   declare -r tmpfile=$(download_file 'https://raw.githubusercontent.com/ghacksuserjs/ghacks-user.js/master/updater.sh')
@@ -188,13 +186,11 @@ update_updater () {
       read -p "" -n 1 -r
       echo -e "\n\n"
       if [[ $REPLY =~ ^[Nn]$ ]]; then
-        # Update available, but user chooses not to update
-        return 0
+        return 0 # Update available, but user chooses not to update
       fi
     fi
   else
-    # No update available
-    return 0
+    return 0 # No update available
   fi
   mv "${tmpfile}" "${SCRIPT_DIR}/updater.sh"
   chmod u+x "${SCRIPT_DIR}/updater.sh"
@@ -209,7 +205,7 @@ update_updater () {
 
 # Returns version number of a user.js file
 get_userjs_version () {
-  echo $(sed -n '4p' "$1")
+  echo "$(sed -n '4p' "$1")"
 }
 
 add_override () {
@@ -354,8 +350,9 @@ if [ $# != 0 ]; then
           ;;
         r)
           tfile=$(download_file 'https://raw.githubusercontent.com/ghacksuserjs/ghacks-user.js/master/user.js')
-          echo -e ${ORANGE}"Warning: user.js was saved to temporary file ${tfile}"${NC}
-          open "$tfile"
+          mv $tfile "${tfile}.js"
+          echo -e ${ORANGE}"Warning: user.js was saved to temporary file ${tfile}.js"${NC}
+          open "${tfile}.js"
           exit 1
           ;;
         \?)
--- a/updater.sh
+++ b/updater.sh
@@ -54,6 +54,10 @@ usage() {
   echo -e "\t\t\t\t\t Ex: -o \"override folder\" "
   echo -e "\t-n,\t\t Do not append any overrides, even if user-overrides.js exists."
   echo -e
+  echo -e "Deprecated Arguments:"
+  echo -e "\t-donotupdate,\t Use instead -d"
+  echo -e "\t-update,\t Use instead -u"
+  echo -e
   exit 1
 }
 
--- a/updater.sh
+++ b/updater.sh
@@ -12,7 +12,7 @@ readonly currdir=$(pwd)
 
 sfp=$(readlink -f "${BASH_SOURCE[0]}" 2>/dev/null || greadlink -f "${BASH_SOURCE[0]}" 2>/dev/null)
 if [ -z "$sfp" ]; then sfp=${BASH_SOURCE[0]}; fi
-readonly SCRIPT_DIR="$(dirname "${sfp}")"
+readonly SCRIPT_DIR=$(dirname "${sfp}")
 
 
 #########################
@@ -22,10 +22,10 @@ readonly SCRIPT_DIR="$(dirname "${sfp}")"
 # Colors used for printing
 RED='\033[0;31m'
 BLUE='\033[0;34m'
-BBLUE='\033[1;34m' 
+BBLUE='\033[1;34m'
 GREEN='\033[0;32m'
 ORANGE='\033[0;33m'
-CYAN='\033[0;36m'  
+CYAN='\033[0;36m'
 NC='\033[0m' # No Color
 
 # Argument defaults
@@ -43,13 +43,17 @@ PROFILE_PATH=false
 #########################
 
 set_wd () {
+  declare -r macdir=~/Library/Application\ Support/Firefox/Profiles/
+  declare -r nixdir=~/.mozilla/firefox/
+  local ff_profile
   if [ "$PROFILE_PATH" = false ]; then
     ff_profile="$SCRIPT_DIR"
   elif [ "$PROFILE_PATH" = "list" ]; then
-    if [ "$(uname)" == "Darwin" ]; then
-      firefox_dir=~/Library/Application\ Support/Firefox/Profiles/ 
-    elif [ "$(expr substr $(uname -s) 1 5)" == "Linux" ]; then
-      firefox_dir=~/.mozilla/firefox/
+    local firefox_dir=""
+    if [ -d $macdir ]; then
+      firefox_dir=$macdir
+    elif [ -d $nixdir ]; then
+      firefox_dir=$nixdir
     else
       echo -e ${RED}"Error: Sorry, -l is not supported for your OS"${NC}
       exit 1
@@ -70,7 +74,7 @@ set_wd () {
 #      Arguments       #
 #########################
 
-usage() {                                             
+usage() {
   echo -e ${BLUE}"\nUsage: $0 [-h] [-p PROFILE] [-u] [-d] [-s] [-n] [-b] [-c] [-v] [-r] [-o OVERRIDE]\n"${NC} 1>&2  # Echo usage string to standard error
   echo -e "Optional Arguments:"
   echo -e "\t-h,\t\t Show this help message and exit."
@@ -93,7 +97,7 @@ usage() {
   echo -e "\t\t\t\t\t Ex: -o \"override folder\" "
   echo -e "\t-n,\t\t Do not append any overrides, even if user-overrides.js exists."
   echo -e "\t-v,\t\t Open the resulting user.js file."
-  echo -e "\t-r,\t\t Only download user.js to the userjs_temps dir and open the file."
+  echo -e "\t-r,\t\t Only download user.js to a temporary file and open it."
   echo -e
   echo -e "Deprecated Arguments (they still work for now):"
   echo -e "\t-donotupdate,\t Use instead -d"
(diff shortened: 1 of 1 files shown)


overdodactyl:
In the latest commit, I've changed it so the deprecated arguments will still work, but the following message will show up in yellow:

```
Warning: command line arguments have changed.
-donotupdate has been deprecated and may not work in the future.

Please view the new options using the --help argument.
```

This should prevent breakage for people who are used to the current form, but could help transition to the more flexible version in this PR.

overdodactyl:
Just took a closer look at the Windows version.  Maybe I should try and emulate those options a little closer instead...

overdodactyl:
A bit more elegant of an option:

```bash
Usage: ./updater.sh [-h] [-u] [-d] [-s] [-n] [-o OVERRIDE]

Optional Arguments:
	-h,		 Show this help message and exit.
	-u,		 Update updater.sh and execute silently.  Do not seek confirmation.
	-d,		 Do not look for updates to updater.sh.
	-s,		 Silently update user.js.  Do not seek confirmation.
	-o OVERRIDE,	 Filename or path to overrides file (if different than user-overrides.js).
	-n,		 Do not append any overrides, evein if user-overrides.js exists.
```

earthlng:
great idea and nicely implemented! couple nits if you don't mind:
- I made a commit to keep old parameters case-insens
- what happens if I provide -o without an argument?
- backup-single: it shouldn't delete all old backups IMO. especially since $filename* is pretty relaxed.
I would do something like this:
```bash
mkdir -p userjs_backups
bakname="userjs_backups/${filename}.backup.$(date +"%Y-%m-%d_%H%M")"
if [ $BACKUP = "single" ]; then
  bakname="userjs_backups/${filename}.backup"
fi
```
users can delete their old backups themselves if they want. You might need to pass a param to `mv` to force overwrite (?), IDK.
- `./updater.sh "$@"` - I'm not sure if that works correctly. Don't you need to pass `-d`? I guess later params overwrite earlier ones?
- can you ELI5 `'s|/\*|\n&|g;s|*/|&\n|g'` and `'/\/\*/,/*\//d'`
- the 1st regex is a bit risky IMO. I would split them in 2 and make each 1 more restrictive:
`'s/^\s*\/\/.*$//'` and `'s/\);\s*\/\/.*$//'
- the `else` in remove_comments() is unnecessary but it's fine if you want to keep it

earthlng:
there may also be some problems with OVERRIDE if the paths contain spaces. Maybe wrap the param to `add_override` in quotes and also `FILES="${input}/*"` and `cat "$input" >> user.js` ?
I guess it won't support paths with commas in it? but it's probably not worth supporting that edge case

overdodactyl:
Thanks earthling!! I've fixed some of the problems in the latest version:

>what happens if I provide -o without an argument?

I added the following:

```bash
:)
          echo -e ${RED}"Error! Option -$OPTARG requires an argument."${NC} >&2
          exit 1
          ;;
```

which should prompt the error when an argument for `-o` is missing, and it should also handle similar scenarios in the future.

>backup-single: it shouldn't delete all old backups IMO....

Agreed...I implemented your suggestion :)

>I'm not sure if that works correctly. Don't you need to pass -d? I guess later params overwrite earlier ones?

I've changed it to `"$@ -d"`, which I think should work.  `$@` grabs the previously used arguments and `-d` will override any other arguments before it.

I have a little more work to do regarding the regex and spaces in path names.

Paths:

* `getups` seems to parse always parse at spaces, but we can get around this by using quotes.   For example, `-o "/sample folder/relaxed.js"` should now work.  If pointing just to a directory with a space, however, it still breaks.  i.e `-o "/sample folder/` won't work:

```bash
FILES="${input}/*"
    for f in $FILES
```

seems to parse at the space.

Regex:

To be honest, regular expressions aren't my forte, and most of those expressions came from stack overflow adaptations haha.  So I'll have to look a little closer into those.  I tried using `'s/\);\s*\/\/.*$//'` from your suggestion above, but get this error:  

```txt
sed: 1: "s/\);\s*\/\/.*$//
": RE error: parentheses not balanced
```

overdodactyl:
Directories with spaces should work with the latest version.  

Turns out for loops use `$IFS` to parse elements, which is set to a space by default.

claustromaniac:
`'s/[)];\s*\/\/.*$//'` works on my end.

Those regexes still can be improved, as in made less risky, but they can get pretty long and awful to read. I can offer help with that if that's not a problem.

overdodactyl:
Thanks @claustromaniac!! Always appreciate the help :)

Unfortunately that doesn't seem to work on my end; I remember seeing something about how macOS deals with spaces a little weird compared to Linux.  

Using the format you gave, the following does work however: `sed -e 's/^[[:space:]]*\/\/.*$//' -e 's/[)];[[:space:]]*\/\/.*$//'` (just replaces the `\s`with `[[:space:]]`).  Does that work for you as well or do I just need to keep both versions in there?

earthlng:
- `FILES="${input}"/*` - maybe limit it to .js files in folders? `-o file1.txt` still works

as for the regexes:
- `'/^[[:space:]]*$/d'` is already covered by `'/^\s*$/d'` or is that not the case in Mac for some reason?
- I really don't understand if or why `'/\/\*/,/*\//d'` works at all because the 2nd `*` is not escaped (???)
- `'s|/\*|\n&|g;s|*/|&\n|g'` seems unnecessary
- and while `'s/\s*\/\/ .*$//'` is somewhat safe because of the space after `//`, it also won't match comments users may have in their overrides that don't have the space.

if you don't mind keeping the end-of-line comments (for now) then this works for me in my linux VM:
`sed -e 's/^\s*\/\/.*$//' -e '/^\/\*/,/\*\//d' -e '/^\s*$/d'`
- the 1st cmd is more restrictive because it only matches at the start of lines. But it also matches comments without a space after `//`
- the 2nd is what I believe it should be and also only matches multi-line comments if they start at column 0

lastly the cat is not really necessary because you can append the sed output directly to no_comments.js (but if you prefer to keep it as is that's fine too)

ps I have mixed feelings about giving users the option to strip all comments from the user.js because it kind of defeats the purpose of a lot of what we're doing here BUT it's your script and if you want to do it I'll commit it

overdodactyl:
Thanks earthlng! 

The latest version restricts overrides to files ending in `.js`.

Appreciate all the help with the regex.  That being said, I say we just get rid of the minify option (it's gone in the latest commit); my only thinking there was that it might be a nice feature for "production" use after users have already looked over the file.  I didn't mean to introduce something that negates all the work you guys put in here, so my apologies :)

overdodactyl:
Do you think we should include the `-n` option?

It's essentially equivalent to:

`./updater.sh -o ""`

earthlng:
I would have just limited it to .js files for folders (`FILES="${input}"/*.js` ?) but it's up to you.

>  I didn't mean to introduce something that negates all the work you guys put in here, so my apologies :)

no worries mate :) I like the idea and can definitely see some good uses for it (mainly comparing new vs old) but I just think the user.js should stay intact (because of the numbers, sections, etc etc). What we could do though is to provide an option to write a diff file comparing just the active user_pref lines from the new user.js and the backup. (?)

> Deprecated Arguements:

"Deprecated Arguments (they still work for now):" maybe? or something like that

re `-n`: I would keep it. Less explaining to do

overdodactyl:
Made those changes!

For colliding arguments I'm tempted to say we can just rely on people using it correctly, but I'm not sure if that's always the wisest choice haha.  In the latest version, I moved `-o` after `-n`.  If people are explicitly writing an override file to be used, I think it's fair to assume they want it. As it is right now, `-d` takes precedence over `-u`.  That ones a little tougher to decide because if both are used, users will either never get an updated version (and think they are), or will be downloading and executing new files when they don't want to.

Is correct usage a fair assumption or do you think I should add in a catch if both are listed?

I really like the diff file idea though, let's add that!

overdodactyl:
Going back the regex ...

>'/^[[:space:]]*$/d' is already covered by '/^\s*$/d' or is that not the case in Mac for some reason?

I'm not able to get most of the regex's to work when they include `\s`, only with `[[:space:]]` instead.  I included both because I wasn't sure if `[[:space:]]` worked for Linux as well

earthlng:
> I moved -o after -n

it doesn't work like that :) the order of `case` is irrelevant because it processes 1 at a time and doesn't care if it hits the 3rd "case" or the 5th or whatever. What matters is the order of the argument given on the command line: `-o ... -n` is different from `-n -o ...`. The problem is with storing multiple arguments in the same variable.
IMO -n should take precedence over -o because explicit "denies" should always have higher priority.
If we want to do that, we have to store -n in a new variable (SKIPOVERRIDE or so) and then in `update_userjs ()`:
```bash
backup_file user.js
if [ "$SKIPOVERRIDE" = "yes" ]; then
  # optional message goes here
  return 0
fi
IFS=',' read -ra FILES <<< "$OVERRIDE"
# ...
```

re `\s` vs `[[:space:]]`: `[[:space:]]` is just the "normal" space whereas `\s` is every horizontal whitespace including normal space, tabs and I think a couple more

earthlng:
> For POSIX compliance, use the character class [[:space:]] instead of \s, since the latter is a GNU sed extension.

maybe it'd be better to use something that works the same for all nix's, like perl

claustromaniac:
@earthlng actually `[:space:]` is the POSIX `\s` according to:
https://en.wikipedia.org/wiki/Regular_expression#Character_classes
see also https://stackoverflow.com/questions/28256178/how-can-i-match-spaces-with-a-regexp-in-bash

It works fine on my end BTW. No need to keep separate expressions with `\s` for Linux users IMO.

overdodactyl:
>it doesn't work like that :)

Oh :/ Ya you're totally right, my bad haha.  And I like the idea of using SKIPOVERRIDE.

On my end then the following works:

`sed -e 's/^[[:space:]]*\/\/.*$//' -e '/^\/\*/,/\*\//d' -e '/^[[:space:]]*$/d'` to remove all non end-of-line comments

earthlng:
> On my end then the following works:

same here

overdodactyl:
For end-of-line comments, how does this look?

`'s/);[[:space:]]\/\/.*/);/'`

Should replace anything matching `);<any # of spaces>//` with `);`

updater.sh
```diff
@@ -182,15 +274,38 @@ get_userjs_version () {
...
 
+add_override () {
+  input=$1
+  if [ -f "$input" ]; then
+    echo "" >> user.js
+    cat "$input" >> user.js
+    echo -e "Status: ${GREEN}Override file appended:${NC} ${input}"
+  elif [ -d "$input" ]; then
+    FSAVEIFS=$IFS
+    IFS=$(echo -en "\n\b") # Set IFS
```
claustromaniac:
```suggestion
    IFS=$'\n\b' # Set IFS
```

earthlng:
^^ works for me. because we're not stripping the real user.js we don't have to be overly cautious. :+1:

updater.sh
```diff
@@ -182,15 +274,38 @@ get_userjs_version () {
...
 update_userjs () {
   backup_file user.js
-  if [ -e user-overrides.js ]; then
-    cat user-overrides.js >> user.js
-    echo -e "Status: ${GREEN}Your user-overrides.js customizations have been applied!${NC}"
+    if [ "$OVERRIDE" != "none" ]; then
+    IFS=',' read -ra FILES <<< "$OVERRIDE"
+    for i in "${FILES[@]}"; do
+        add_override "$i"
+    done
```
claustromaniac: Completely unnecessary alternative:
```shell
update_userjs () {
  backup_file user.js
  if [ "$OVERRIDE" != "none" ]; then
    while IFS=',' read -ra FILE; do
      add_override "$FILE"
    done <<< "$OVERRIDE"
  fi
}
```
It's more readable to *me*, but that's highly subjective.

updater.sh
overdodactyl: Thanks! Changed in 
d6401a3

overdodactyl:
There's probably a more elegant way to do it, but I've just added the the `-c` option, which creates a diff file.

updater.sh
overdodactyl: >It's more readable

Agreed! 4e8291d

overdodactyl:
Implemented the `SKIPOVERRIDE` suggestion from above, and decided to bump the version up to `2.0`

claustromaniac:
> There's probably a more elegant way to do it, but I've just added the the -c option, which creates a diff file.

I like how you implemented it. It is relatively easy to read and performs great. I considered some ways to make the regexes safer (like I said before) but I concluded that they're probably not worth it, because regex support is kinda limited without using Perl, and we would have to add *a lot* of complexity for very little benefit. The current approach seems reasonable to me :+1: 

> If you think I'm adding too many/useless options feel free to let me know :) I'd rather not keep stuff that you guys don't see a practical use for!

Without feedback it is hard to say. I like them, personally. They add flexibility, and the only significant (yet subjective) tradeoff is readability. If someone can split hairs over that, they surely can modify the script to their liking or rewrite it, or better yet: provide feedback. :stuck_out_tongue:

updater.sh
```diff
@@ -182,24 +275,138 @@ get_userjs_version () {
...
   fi
+  while IFS=',' read -ra FILE; do
+    add_override "$FILE"
+  done <<< "$OVERRIDE"
 }
 
+remove_comments () {
+  from_file=$1 
+  to_file=$2
+  sed -e 's/^[[:space:]]*\/\/.*$//' -e '/^\/\*/,/\*\//d' -e '/^[[:space:]]*$/d' -e 's/);[[:space:]]\/\/.*/);/' $from_file > $to_file
```
claustromaniac:
```suggestion
  sed -e 's/^[[:space:]]*\/\/.*$//' -e '/^\/\*/,/\*\//d' -e '/^[[:space:]]*$/d' -e 's/);[[:space:]]*\/\/.*/);/' $from_file > $to_file 
```

earthlng:
I don't have a lot of time today but let's see ...

- What I really don't like is the meddling with `"$currdir"` all of a sudden
- does Linux or Mac even have portable versions of Firefox?
- Is Darwin the name of the current MacOS version? If so, wouldn't it be better to just check if that weird mac folder exists? I'm pretty sure not a single non-MacOS 'nix will have that folder.
- if only 1 profile exists it would be nice to just auto-select that without asking
  - if [[ $(ls -d "$firefox_dir"* | wc -l) == "1" ]]; then ... else ...

claustromaniac:
I see this keeps getting better :)

Sorry I haven't provided much feedback so far, I still haven't been able to read the whole thing in one go. Here's something I noticed, though:
https://github.com/ghacksuserjs/ghacks-user.js/blob/ce0f86558f35de0a73e77764b2c6a2123d13899c/updater.sh#L288-L289

Consider using something like this to avoid writing empty files if `diff` fails or there are simply no differences:
```shell
    diff=$(diff -w -B -U 0 $past_nocomments $current_nocomments) 
    if [ ! -z "$diff" ]; then
      echo "$diff" > "$diffname"
    fi
```

overdodactyl:
Good idea!

I used this:

```bash
    diffname="userjs_diffs/diff_$(date +"%Y-%m-%d_%H%M").txt"
    diff=$(diff -w -B -U 0 $past_nocomments $current_nocomments) 
    if [ ! -z "$diff" ]; then
      echo "$diff" > "$diffname"
      echo -e "Status: ${GREEN}A diff file was created:${NC} ${PWD}/${diffname}"
    else
      echo -e "Warning: ${ORANGE}Your new user.js file appears to be identical.  No diff file was created."${NC}
    fi
```

so it now notifies the user if there were no diffs and lists the path to the diff file if it was created

earthlng:
> Thanks again @earthlng! Definitely a much improved version :)

glad you like it. Thanks for testing.

I think this is ready to land then? except maybe ...

https://unix.stackexchange.com/questions/253376/open-command-to-open-a-file-in-an-application

> The primary purpose of OS X's open command is to open a file in the associated application. The equivalent of that on modern non-OSX unices is xdg-open.

On non-OSX unices "the open command is a link to the openvt command and opens a binary in a new virtual console." ("which can only be done by root and isn't used very often in this century since most people only ever work in a graphical window environment.")

should we just add a MAC-ONLY tag to the description in usage() or what do you guys think?
we have 4 options:
- do nothing and let users figure it out for themselves
- add a note or a tag or something
- add more code to use open on OSX and xdg-open on other unices
- remove the 2 arguments/functions using open

earthlng:
and what about this?
> if only 1 profile exists it would be nice to just auto-select that without asking
`if [[ $(ls -d "$firefox_dir"* | wc -l) == "1" ]]; then ... else ...`

earthlng:
ah okay, thanks. Why all the new spaces in create_diff ?

> Or don't ask when using -l if only one exists?

this

earthlng:
> Maybe just do that if -p or -l was not set and only one profile exists? 

You can add something there too but I think it should also remain possible to use the script in a separate folder ie not the profile folder.

overdodactyl:
>Why all the new spaces in create_diff ?

Sorry, didn't realize I did that.  Should be fixed!

I modified it so it should auto select when using `-l` when there's only one profile.  I was having trouble with `[[ $(ls -d "$firefox_dir"* | wc -l) == "1" ]];`, so I tried a bit different of an approach.

updater.sh
```diff
@@ -58,12 +58,19 @@ set_wd () {
...
     fi
-    echo -e ${GREEN}"The following profiles were found:\n"${ORANGE}
-    ls -d "$firefox_dir"*
-    echo -e ${RED}"\nWhich profile would you like to update?"${NC}
-    read -p ""
-    echo -e ""
-    ff_profile=$REPLY
+    numdirs=("$firefox_dir"/*)
+    numdirs=${#numdirs[@]}
+    if [[ $numdirs == "1" ]]; then
```
claustromaniac: Alternative:
```shell
    numdirs=("$firefox_dir"/*)
    if [[ ${#numdirs[@]} == 1 ]]; then 
```
Works on my end anyway.

overdodactyl: Definitely better, thanks! 91501a4

earthlng: code-wise this works but have you guys tested it on the real folder? because in my linux vm it also counts the profiles.ini in that folder and therefore will never return 1

overdodactyl: It works on my end, but it sounds like the folder layout is a little different for macOS.  

profiles.ini is inside `.../Application Support/Firefox` and profiles are all profiles are located inside `.../Application Support/Firefox/Profiles`.

Maybe we just add a files variable inside the os check that's either 0 or 1, then in subtract in the if statement?  Or does that sound too fragile?

Something to the effect of:

```bash
local num_files=0
if [ -d "$macdir" ]; then
      firefox_dir=$macdir
elif [ -d $nixdir ]; then
      firefox_dir=$nixdir
      num_files=1
...
numdirs=("$firefox_dir"/*)
if [[ ${#numdirs[@]} == (1-$num_files) ]]; then...
```

(haven't tested that..)

claustromaniac: @earthlng True. I didn't test on Linux before. I get the same as you.

@overdodactyl What problems were you having with `[[ $(ls -d "$firefox_dir"* | wc -l) == "1" ]];` though? maybe it would be better to find a way to get that approach (or something similar) to work. IMO subtracting doesn't sound like a good approach because it relies on users not adding any other files by themselves and such.

earthlng: does this not work on mac?
`firefox_dir=~/Library/Application\ Support/Firefox/Profiles/; if [[ $(find $firefox_dir/ -maxdepth 1 -type d | wc -l) == "2" ]]; then echo only 1 profile found; fi`

== "2" instead of 1 because it also lists the /Profiles/ folder

earthlng: assuming you only have 1 profile

overdodactyl: >What problems were you having with [[ $(ls -d "$firefox_dir"* | wc -l) == "1" ]]

It actually works if we remove the double bracket: `[ $(ls -d "$firefox_dir"* | wc -l) == "1" ]`

>IMO subtracting doesn't sound like a good approach because it relies on users not adding any other files by themselves and such.

Agreed.  However this method does as well (just relies on users not adding folders).  Worst case is that they just have to enter it themselves though

overdodactyl: Does single brackets work for you guys as well or should I give that new version a go?

earthlng: I'd use the find command above for the if line and keep the ls for the listing if more than 1 profile exists

earthlng: `if [[ $(find "$firefox_dir" -maxdepth 1 -type d | wc -l) == "2" ]]; then` should work on linux and mac

overdodactyl: For mac it looks like that will work either by changing `==` to `-eq` or removing the double bracket (or both):

```bash
[[ $(find "$firefox_dir" -maxdepth 1 -type d | wc -l) -eq 2 ]]; then ..
[ $(find "$firefox_dir" -maxdepth 1 -type d | wc -l) == "2" ]; then ...
[ $(find "$firefox_dir" -maxdepth 1 -type d | wc -l) -eq 2 ]; then ..
```

overdodactyl:
I *think* that would work, but I'm not as familiar with bash on Windows and don't have a great way to test it at the moment :/

That could be a good addition though!

earthlng:
do any of these tools come with curl or wget? I'd really rather not make this any more complicated and complex than it already is. We already have an awesome updater batch for Windows thanks to a certain someone :) :kissing_cat:

earthlng:
`IMPORTANT: if the path include spaces, wrap the entire argument in quotes."` in usage should be under -p instead of -l, no?

updater.sh
overdodactyl: >Maybe we could use that?

Sounds like a good idea to me! Do you want to add it in or should I?

overdodactyl:
>IMPORTANT: if the path include spaces, wrap the entire argument in quotes." in usage should be under -p instead of -l, no?

Good catch...just fixed along with the single profile check

earthlng:
updated the wiki: https://github.com/ghacksuserjs/ghacks-user.js/wiki/3.3-Updater-Scripts

earthlng:
can I merge this now? we've been working on it for almost a week and I think it's in a good state now.

we can always create a new PR for more stuff or changes

overdodactyl:
Let's do it :)

earthlng:
Thanks a lot @overdodactyl & @claustromaniac !!

PR #544, opened 2018-11-20
2030: inactive media.autoplay.default

bogachenko:


bogachenko:
Oh yeah.. not disabled, not active!

Thorin-Oakenpants:
Yeah, we would do it something like that. But I'm going to close this, as Earthlng says it might be better to wait for the next change in FF64/65 (less hassle for end-users)

bogachenko:
Yes. I read

PR #549, opened 2018-11-23
Update updater.sh

earthlng:
my version based on the latest pull request by @overdodactyl

### DIFF
--- a/updater.sh
+++ b/updater.sh
@@ -131,14 +131,15 @@ fi
 download_file () {
   declare -r url=$1
   declare -r tf=$(mktemp)
+  local dlcmd=""
 
   if [ $DOWNLOAD_METHOD = "curl" ]; then
-    curl -o $tf "${url}"
+    dlcmd="curl -o $tf"
   else
-    wget -O $tf "${url}"
+    dlcmd="wget -O $tf" 
   fi
 
-  echo "$tf" # return the temp-filename
+  $dlcmd "${url}" >/dev/null && echo "$tf" || echo "" # return the temp-filename (or empty string on error)
 }
 
 
--- a/updater.sh
+++ b/updater.sh
@@ -2,126 +2,160 @@
 
 ## ghacks-user.js updater for macOS and Linux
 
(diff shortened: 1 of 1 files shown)


earthlng:
- I inlined a lot of functions because it was kind of a mess with functions spread all over the place. You had to scroll up and down and up again to find code that's only used once anyway.
There were other problems as well, like the backup_file() not only backed up a file but then also replaced it with a file that needed to be previously created by another function.
Or the view function ran even if a user cancelled the process f.e after seeing that there's no new version available. Showing the user his "old" user.js is probably not very useful and potentially confusing.
Similar thing with create_diff which expected a past_user.js but there was none if the user cancelled the "confirmation".
- changed the download function to use a mktemp file and return the path to the newly downloaded temp file. (= no need to cleanup any temp folders)
- got rid of the perl option again because I couldn't figure out how to download to a given temp file in the time I was willing to spend reading about that. Since nobody ever reported having a system without neither curl nor wget I say fuck it. But feel free to add it back in if you want to spend more time on this.
- the update_updater function was completely fucked. $currdir and script dir aren't necessarily the same thing. Because it's now only using the SCRIPT_DIR and a tmpfile, the function can be run before changing the working directory and thus we are also more likely to restore the correct original working dir after the script ends. :fingerscrossed: :jesus_needed:

I hope you don't mind that I ended up changing more than I probably should have. Take parts of it, all of it or nothing at all - I don't care, I'll commit whatever you want me to.

ps I just hacked away at it so there are probably still a bunch of problems, bugs, typos, etc. Definitely needs some review and testing

overdodactyl:
Changes look great! Thanks for all the work on it :) Sorry I'm just getting a chance to look - took most of yesterday off.  

There's a few things that are acting a little wonky on my end, but I think this it's definitely heading in a better direction.  

Do you mind making a commit with this version on #543 and we can continue working on it there?

earthlng:
> Do you mind making a commit with this version on #543 and we can continue working on it there?

done

PR #551, opened 2018-11-24
Tor-related warnings

### LINKED ISSUE #546: Firefox over Tor
Something I noticed some time ago while doing diffs. We have:
```js
user_pref("network.proxy.socks_remote_dns", true);
```
just in case users access the Tor network from Firefox, but we also have:
```js
user_pref("privacy.donottrackheader.enabled", true);
```
...which is default false in TBB, therefore it would make such FF users more unique. I am aware that there are other factors like PB mode, and even some extensions that can set the DNT header regardless of this pref, but maybe there should be a little note/warning there like:
```js
/* 1610: ALL: enable the DNT (Do Not Track) HTTP header
 * [SETTING] Privacy & Security>Content Blocking>Send websites a "Do Not Track"...
 * [NOTE] DNT is enforced with TP (see 0420) regardless of this pref 
 * [NOTE] Set to false if connecting over the Tor network ***/
user_pref("privacy.donottrackheader.enabled", true);
```
or, if you (we?) don't support Firefox over Tor, maybe it would help to add a little statement somewhere making that fact very clear and shortly explaining *why*. Maybe in `0704`, like:
```js
/* 0704: enforce the proxy server to do any DNS lookups when using SOCKS
 * e.g. in TOR, this stops your local DNS server from knowing your Tor destination
 * as a remote Tor node will handle the DNS request
 * [1] http://kb.mozillazine.org/Network.proxy.socks_remote_dns
 * [2] https://trac.torproject.org/projects/tor/wiki/doc/TorifyHOWTO/WebBrowsers 
 * [NOTE] We do *not* support connecting over Tor on Firefox because a number of settings and
characteristics differ from the TBB and they can make you more easily identifiable ***/
user_pref("network.proxy.socks_remote_dns", true);
```
Whatcha think?

claustromaniac:
Closes #546.

To squash, or not to squash, that is the question.

### DIFF
--- a/README.md
+++ b/README.md
@@ -6,6 +6,10 @@ The `ghacks user.js` is a **template** which aims to provide as much privacy and
 
 Everyone, experts included, should at least read the [implementation](https://github.com/ghacksuserjs/ghacks-user.js/wiki/1.3-Implementation) wiki page, as it contains important information regarding a few `ghacks user.js` settings.
 
+Note that we do *not* recommend connecting over Tor on Firefox. Use the [Tor Browser](https://www.torproject.org/projects/torbrowser.html.en) if your [threat model](https://www.torproject.org/about/torusers.html.en) calls for it, or for accessing hidden services. 
+
+Also be aware that this `user.js` is made specifically for Firefox. Using it as-is in other Gecko-based browsers can be counterproductive, especially in the Tor Browser.
+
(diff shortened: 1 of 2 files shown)


Thorin-Oakenpants:
Squash and merge... that is the answer

claustromaniac:
LOL that's a lie! I'm hungrrrrrrry! :crying_cat_face:

PR #554, opened 2018-11-25
Update updater.sh

earthlng:


updater.sh
```diff
@@ -102,76 +92,101 @@ usage() {
...
-  echo -e                "  ####                           ghacks user.js                           ####"
-  echo -e                "  ####       Hardening the Privacy and Security Settings of Firefox       ####"
-  echo -e                "  ####           Maintained by @Thorin-Oakenpants and @earthlng           ####"
-  echo -e                "  ####            Updater for macOS and Linux by @overdodactyl            ####"
-  echo -e                "  ####                                                                    ####"
-  echo -e                "  ############################################################################"${NC}
-  echo -e
-  echo -e
-  echo -e "Documentation for this script is available here: ${CYAN}https://github.com/ghacksuserjs/ghacks-user.js/wiki/3.3-Updater-Scripts${NC}\n"
+  if [[ $(grep '^\[Profile' "$inifile" | wc -l) == "1" ]]; then ### only 1 profile found
```
overdodactyl:
```suggestion
  if [ $(grep '^\[Profile' "$inifile" | wc -l) == "1" ]; then ### only 1 profile found
```

overdodactyl:
Looks great @earthlng!! 

Besides those two minor things, the only other thing I really noticed is the user input for `-l` won't work if you have 11 or more profiles...not sure if that's something we need to worry about or not

updater.sh
```diff
@@ -102,76 +92,101 @@ usage() {
...
+    grep '^\[Profile' -A 4 "$inifile" | grep -v '^\[Profile' > $tfile
+  else
+    grep -E -v '^\[General\]|^StartWithLastProfile=|^IsRelative=' "$inifile"
+    echo ''
+    read -p 'Select the profile number ( 0 for Profile0, 1 for Profile1, etc ) : ' -n 1 -r
+    echo -e "\n"
+    if [[ $REPLY =~ ^[0-9]$ ]]; then
```
claustromaniac:
```suggestion
    if [[ $REPLY =~ ^[0-9]+$ ]]; then
```

claustromaniac: Even if @earthlng doesn't care about folks with 11+ profiles, personally I'd still prefer to remove the `-n 1` because my fingers are clumsy sometimes.

updater.sh
```diff
@@ -102,76 +92,101 @@ usage() {
...
-  echo -e                "  ############################################################################"${NC}
-  echo -e
-  echo -e
-  echo -e "Documentation for this script is available here: ${CYAN}https://github.com/ghacksuserjs/ghacks-user.js/wiki/3.3-Updater-Scripts${NC}\n"
+  if [[ $(grep '^\[Profile' "$inifile" | wc -l) == "1" ]]; then ### only 1 profile found
+    grep '^\[Profile' -A 4 "$inifile" | grep -v '^\[Profile' > $tfile
+  else
+    grep -E -v '^\[General\]|^StartWithLastProfile=|^IsRelative=' "$inifile"
+    echo ''
+    read -p 'Select the profile number ( 0 for Profile0, 1 for Profile1, etc ) : ' -n 1 -r
```
claustromaniac:
```suggestion
    read -p 'Select the profile number ( 0 for Profile0, 1 for Profile1, etc ) : ' -r
```

updater.sh
```diff
@@ -102,76 +92,101 @@ usage() {
...
+    echo -e "\n"
+    if [[ $REPLY =~ ^[0-9]$ ]]; then
+      grep '^\[Profile'${REPLY} -A 4 "$inifile" | grep -v '^\[Profile'${REPLY} > $tfile
+      if [[ "$?" != "0" ]]; then
```
claustromaniac:
```suggestion
      if [ !$? ]; then
```

earthlng:
okay then, thanks guys! I kept the broken else but commented it out (as a reminder that there's a 3rd option ie -p) and changed the proposed output when the user-input for profile selection doesn't match our "safe" regex to "Invalid selection!" because it's generally better not to output random unsanitized user input. I committed everything else exactly as suggested

overdodactyl:
Looks great!

PR #555, opened 2018-11-26
Improve remove comment regex

overdodactyl:
* Split the regex into multiple lines for readability
* Added `tr  -s '[:space:]' > "$2`" to remove extra white spaces.  Without doing so, identical prefs will be counted as diffs if the number of spaces aren't the same
     * Ex: https://github.com/pyllyukko/user.js/blob/master/user.js

### DIFF
--- a/updater.sh
+++ b/updater.sh
@@ -2,7 +2,7 @@
 
 ## ghacks-user.js updater for macOS and Linux
 
-## version: 2.1
+## version: 2.2
 ## Author: Pat Johnson (@overdodactyl)
 ## Additional contributors: @earthlng, @ema-pe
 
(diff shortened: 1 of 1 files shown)


earthlng:
wasn't the idea to compare **ghacks** user.js' ? why should we make it support other user.js'

on a related note, how about we use an IMHO better output for the diffs ie side-by-side?
fe comparing v62 vs v63 betas would output this:
```
user_pref("browser.search.countryCode", "US");				      <
user_pref("services.blocklist.update_enabled", true);			      <
user_pref("privacy.trackingprotection.ui.enabled", true);		      <
user_pref("shield.savant.enabled", false);				      <
									      >	user_pref("browser.newtabpage.activity-stream.feeds.section.topstories", false);
									      >	user_pref("browser.newtabpage.activity-stream.section.highlights.includePocket", false);
									      >	user_pref("browser.newtabpage.activity-stream.showSponsored", false);
									      >	user_pref("network.gio.supported-protocols", "");
									      >	user_pref("dom.forms.datetime", false);
									      >	user_pref("security.ssl.require_safe_negotiation", true);
user_pref("gfx.downloadable_fonts.enabled", true);			      <
user_pref("media.autoplay.enabled", false);				      |	user_pref("media.autoplay.default", 1);
user_pref("extensions.webextensions.keepStorageOnUninstall", false);	      <
user_pref("extensions.webextensions.keepUuidOnUninstall", false);	      <
user_pref("browser.cache.offline.enable", false);			      <
```

overdodactyl:
>wasn't the idea to compare ghacks user.js' ? why should we make it support other user.js'

I was just using that as an example of a syntax that wouldn't work with our current version. Our comparison also takes into account overrides, so if a user writes their own tweaks and isn't consistent or adds spaces between versions, or even if an extra space slips into a ghacks pref it could cause problems.  Take it or leave it though, we don't need to add it.  Just thought it was a simple fix that could only help :)

>IMHO better output for the diffs ie side-by-side...

I agree, that looks better

earthlng:
> Without doing so, identical prefs will be counted as diffs if the number of spaces aren't the same

IMO it works exactly as it should. And it can help to find extra spaces that slipped in somehow.

I'll close this and make a new PR for the side-by-side diff

Issue #535, opened 2018-11-13
ToDo: diffs FF63-FF64


earthlng:
FF64 is [scheduled](https://wiki.mozilla.org/Release_Management/Calendar) for release 11th Dec

[FF64 release notes](https://www.mozilla.org/en-US/firefox/64.0/releasenotes/) [when ready]
[FF64 for developers](https://developer.mozilla.org/en-US/docs/Mozilla/Firefox/Releases/64)
[FF64 compatibility](https://www.fxsitecompat.com/en-CA/versions/64/)
[FF64 security advisories](https://www.mozilla.org/en-US/security/advisories/mfsa2018-29/)

140 diffs ( 63 new, 45 gone, 32 different )

### misc:
* [x] Onboarding no longer a system addon (now part of Activity Stream?) - https://github.com/ghacksuserjs/ghacks-user.js/commit/61be5ae563e0e7e3427885a16e6f920437797966
* [x] Pocket no longer a system addon - https://github.com/ghacksuserjs/ghacks-user.js/commit/ccdd4decf01dd3576e84c775d1a99fe3d7f5b27d
* [x] move `0370` to `0105b` - https://github.com/ghacksuserjs/ghacks-user.js/commit/0e1b0a4b6e572f88a7e456a75be8e88f46a9b559

Added here for visibility
* general.buildID.override [583181](https://bugzilla.mozilla.org/show_bug.cgi?id=583181)
   * buildID is a hardcoded value for websites now, except privileged domains (e.g AMO), so there is no need to spoof this anymore. If you do not use RFP, then just be like the other 99% of FF users. Leave this pref alone. Here's how it works
   * > **if** RFP is enabled -> return the new fixed timestamp `20181001000000`
     > **else** if `general.buildID.override` is set -> return that value
     > **else** if URL is not `https://*.mozilla.org/*` -> return the fixed timestamp
     > **else** return the real buildID for all `https://*.mozilla.org/*` URLs

### new in v64.0:

* [x] CFR (Contextual Feature Recommendation) - https://github.com/ghacksuserjs/ghacks-user.js/commit/31adbba774139f1f79cc90c4b48126b5bd3912fa
   * pref("browser.newtabpage.activity-stream.asrouter.userprefs.cfr", true);
[...]

PR #556, opened 2018-11-26
Update updater.sh

earthlng:
nicer diffs

### DIFF
--- a/updater.sh
+++ b/updater.sh
@@ -4,7 +4,7 @@
 
 ## version: 2.2
 ## Author: Pat Johnson (@overdodactyl)
-## Additional contributors: @earthlng, @ema-pe
+## Additional contributors: @earthlng, @ema-pe, @claustromaniac
 
 ## DON'T GO HIGHER THAN VERSION x.9 !! ( because of ASCII comparison in update_updater() )
 
--- a/updater.sh
+++ b/updater.sh
@@ -320,12 +320,10 @@ update_userjs () {
     diff=$(sdiff -s -w 1000 $past_nocomments $current_nocomments)
     if [ ! -z "$diff" ]; then
       local leasttabs=999999
-      local oldIFS=$IFS
-      while IFS= read -r line; do
-        ntabs=$(printf '%s\n' "$line" | tr -c -d "\t" | wc -c)
+      while read -r; do
+        ntabs=$(printf '%s\n' "$REPLY" | tr -c -d "\t" | wc -c)
         if [ $ntabs -lt $leasttabs ]; then leasttabs=$ntabs; fi
       done <<< "$diff"
-      IFS=$oldIFS
       if [ $leasttabs -lt 2 ]; then
         echo "$diff" > "$diffname"
       else
--- a/updater.sh
+++ b/updater.sh
@@ -328,7 +328,7 @@ update_userjs () {
         echo "$diff" > "$diffname"
       else
         # strip leasttabs-1 number of TABS once per line
-        sed -E 's/\t{'$((leasttabs-1))'}//1' <<< "$diff" > "$diffname"
+        sed -E "s/$(printf '\t'){$((leasttabs-1))}//1" <<< "$diff" > "$diffname"
       fi
       echo -e "Status: ${GREEN}A diff file was created:${NC} ${PWD}/${diffname}"
     else
--- a/updater.sh
+++ b/updater.sh
@@ -317,7 +317,7 @@ update_userjs () {
     remove_comments user.js $current_nocomments
 
     diffname="userjs_diffs/diff_$(date +"%Y-%m-%d_%H%M").txt"
-    diff=$(sdiff -s -w 1000 $past_nocomments $current_nocomments)
+    diff=$(diff -B -d -y -W 1000 --suppress-common-lines $past_nocomments $current_nocomments)
     if [ ! -z "$diff" ]; then
       local leasttabs=999999
       while read -r; do
--- a/updater.sh
+++ b/updater.sh
@@ -317,16 +317,21 @@ update_userjs () {
     remove_comments user.js $current_nocomments
 
     diffname="userjs_diffs/diff_$(date +"%Y-%m-%d_%H%M").txt"
-    diff=$(sdiff -s -w 500 $past_nocomments $current_nocomments)
+    diff=$(sdiff -s -w 1000 $past_nocomments $current_nocomments)
     if [ ! -z "$diff" ]; then
       local leasttabs=999999
-      while read line; do
-        ntabs=$(echo "$line" | tr -c -d "\t" | wc -c)
-        # somehow lines starting with TAB are counted as 1 so we need to catch that
-        if [ $ntabs -gt 1 ] && [ $ntabs -lt $leasttabs ]; then leasttabs=$ntabs; fi
+      local oldIFS=$IFS
+      while IFS= read -r line; do
+        ntabs=$(printf '%s\n' "$line" | tr -c -d "\t" | wc -c)
+        if [ $ntabs -lt $leasttabs ]; then leasttabs=$ntabs; fi
       done <<< "$diff"
-      # strip leasttabs-1 number of TABS once per line
-      sed -E 's/\t{'$((leasttabs-1))'}//1' <<< "$diff" > "$diffname"
+      IFS=$oldIFS
+      if [ $leasttabs -lt 2 ]; then
+        echo "$diff" > "$diffname"
+      else
+        # strip leasttabs-1 number of TABS once per line
+        sed -E 's/\t{'$((leasttabs-1))'}//1' <<< "$diff" > "$diffname"
+      fi
       echo -e "Status: ${GREEN}A diff file was created:${NC} ${PWD}/${diffname}"
     else
       echo -e "Warning: ${ORANGE}Your new user.js file appears to be identical.  No diff file was created.${NC}"
--- a/updater.sh
+++ b/updater.sh
@@ -2,7 +2,7 @@
 
 ## ghacks-user.js updater for macOS and Linux
 
-## version: 2.1
+## version: 2.2
 ## Author: Pat Johnson (@overdodactyl)
 ## Additional contributors: @earthlng, @ema-pe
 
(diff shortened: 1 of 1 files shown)


overdodactyl:
Code wise it looks good to me, but I won't have time to do much testing until later today.

Random side note: Given how much you've worked on this, you're more than welcome to move your name onto the "authors" line and include it in the "banner" as well.  We should also add @claustromaniac to the "contributors" line.

earthlng:
if we want to support insanely long user-pref lines we can increase the sdiff width from 500 to 1000 or so, but other than that it should work pretty well (at least it did in my testing)

claustromaniac:
I don't have a whole lot of experience with Bash. Is there any good reason not to use this in L320?
```bash
    diff=$(diff -B -d --suppress-common-lines -y -w $past_nocomments $current_nocomments)
```
It seems to have a much cleaner output than `sdiff`. Would using that help make this more readable? That being said, it works well for me as you wrote it.

> If you do not give `read` any arguments to use to hold the input (relying on the default variable `REPLY`), no whitespace is stripped and you can omit the modification to `IFS`.

Just FYI, changing `IFS` like this:
```bash
while IFS= read -r line; do
```
only applies to the scope of the `while` loop, meaning you don't need to save and restore IFS. Chepner's tip is cool though.

overdodactyl:
I guess my biggest concern is how much the readability seems to depend on your window size and text editor.  I started out looking in Sublime full screen:

screen shot 2018-11-26 at 1 09 35 pm
  
For some reason it adds a ton of extra vertical space between entires can tell by the "map" on the right).

Switched to TextEdit and it was much better in full screen:

screen shot 2018-11-26 at 1 08 46 pm

But shrinking the window size a little made it more difficult to read:

screen shot 2018-11-26 at 3 37 15 pm

Next I checked out Xcode (full screen):

screen shot 2018-11-26 at 3 54 48 pm

overdodactyl:
>It seems to have a much cleaner output than sdiff

It's better on my end, but Sublime and Xcode still doesn't seem to handle it great.

earthlng:
> It seems to have a much cleaner output than sdiff

how so? we can use diff instead of sdiff. I only used sdiff because I somehow missed --suppress-common-lines in the man page :man_facepalming: But I get identical output from both.

As for what @overdodactyl posted, that seems to be a problem with either the code in Mac not working as expected OR your text-editor is set to wrap long lines, I'd guess.
Does diff (/sdiff) in Mac not use TABS to align left and right if you pass a long width?

earthlng:
> IFS ... only applies to the scope of the while loop

thanks that's good to know. I was wondering about that but didn't care to look it up because that other guys solution works perfect here and is shorter

earthlng:
> diff=$(diff -B -d --suppress-common-lines -y -w $past_nocomments $current_nocomments)

not sure how good `-d` works but we can definitely add it.
`--suppress-common-lines -y` can replace `sdiff -s`
`-B` is redundant (isn't it?) because we're diffing pre-stripped files but we can add it anyway.
IDK about using -w though. What exactly does "ignore all white space" mean? Is "hello world" the same as "helloworld" then? Probably not but there's also -b and that's why I'm confused

claustromaniac:
I meant to say that using `diff` instead of `sdiff` might help make *the script* more readable (sorry about that). On my end `diff` outputs a lot less spaces, so I don't need to edit the output to read it somewhat comfortably. If that is not consistent across different OSes, I guess it doesn't matter anyway.

earthlng:
@overdodactyl add an `echo $leasttabs` after the while and if that's in the 50-60 range the problem can only be in `sed -E ...`. I can't help much more because I don't have a mac and the code works in Linux

earthlng:
@claustromaniac wait what? How the hell is
`diff=$(diff -y -W 1000 -B -d  -w --suppress-common-lines $past_nocomments $current_nocomments)`
more readable than
`diff=$(sdiff -s -w 1000 -B -d $past_nocomments $current_nocomments)` ? :smile:

earthlng:
> On my end diff outputs a lot less spaces, so I don't need to edit the output to be able to read it somewhat comfortably. If that is not consistent across different OSes, I guess it doesn't matter anyway.

so I guess it doesn't work on your end either then. It should look like this https://github.com/ghacksuserjs/ghacks-user.js/pull/555#issuecomment-441727759

claustromaniac:
> `-B` is redundant (isn't it?)

Indeed.

> wait what? How the hell is ... more readable than ...

Sure, that line is not more readable that way, but if `diff` outputs just enough spaces to separate the two columns reasonably, maybe you wouldn't need some of the other lines below that one. That's how ***THE SCRIPT*** would be more readable.

As for your other questions... I have a whole 10 minutes of empirical knowledge + shitty documentation. I'll see if I can find something to reply to those accurately.

claustromaniac:
Dude you're misunderstanding everything I said. Ugh. 

Try just 
```shell
diff -B -d --suppress-common-lines -y -w "file1" "file2"``` > temp.txt
vs
```
sdiff -s -w 1000 -B -d "file1" "file2" > temp.txt
```
in the TERMINAL, and see the difference.

earthlng:
> you wouldn't need some of the other lines below that one

Oh now I see what you mean. Yeah sure we can do that but then it either truncates the lines or there's unnecessary space in between. I found using a fixed width is not ideal and that's what the lines below are for but apparently it only works on my system

claustromaniac:
>  I found using a fixed width is not ideal and that's what the lines below are for but apparently it only works on my system

I wouldn't be so sure about that. :sweat_smile: Both dactyl and I said what you wrote works fine to begin with! I just suggested a *possible* way to make the script somewhat shorter ***maybe***. That's all, really.

overdodactyl:
I think using `-t`(`--expand-tabs`) in `diff` is the magic solution for me.  

It looks like `diff` uses x number of tabs, then equals out the spacing via spaces.  Each text editor of mine was treating tabs as a different number of spaces, so it didn't line up properly.

earthlng:
I feel you misunderstood my stupidly placed smiley. It wasn't meant in an offensive way, like laughing at you, at all! I just didn't get that you meant also getting read of the lines below to make THE SCRIPT more readable. Sorry about that

but yeah if we don't care about truncated lines then we can just add `-d --suppress-common-lines -y` and remove all the extra garbage. I'm totally fine with that.
I just wanted to see if I can simulate "auto-width" diffs instead of fixed width, that's all.

earthlng:
> Each text editor of mine was treating tabs as a different number of spaces, so it didn't line up properly.

oh okay so it's an editor issue and you'd have that problem whenever you use `-y` without `-t`.
We could work around that by letting `diff` use spaces instead of TABS but it wouldn't be quite as easy to simulate auto-width as it was with TABS.

claustromaniac:
> I feel you misunderstood my stupidly placed smiley. It wasn't meant in an offensive way, like laughing at you, at all!

Don't you worry, I know you're not such a jerk :laughing: Failing to make myself clear is frustrating, that's all. In fact, I wanted to apologize more for the confusion, but I abstained from doing so because the conversation already got needlessly long (by my fault :crying_cat_face:).

earthlng:
but looking at your screenshots that doesn't seem to be the only problem.
fe in the Xcode pic, IF the code was working as expected, there should only be one tab and a few spaces after `... crashReports.unsubmittedCheck.enabled", false);` but there are clearly more

earthlng:
how many tabs between aaa and bbb are left after this? `echo -e "aaa\t\t\t\t\t\tbbb" | sed -E 's/\t{4}//1'`

overdodactyl:
>how many tabs between aaa and bbb are left after this?

It looks like 5 full tabs, 1 partial tab (?), and 0 spaces:

screen shot 2018-11-26 at 5 52 51 pm

overdodactyl:
>but looking at your screenshots that doesn't seem to be the only problem

Ya, the `sdiff` version seemed to have additional problems as well.

earthlng:
how many with this? `TAB=$"\t"; echo -e "tab\t\t\t\t\t\ttab" | sed -E 's/'${TAB}'{4}//1'`

overdodactyl:
Same thing there

overdodactyl:
>Default on my linux VM is 130

earthlng:
I'll close this PR and if you want (potentially truncated or very wide) side-by-side diffs at all you can just change that one line and add the options you want.

Sorry for wasting your time guys

claustromaniac:
I'm really sorry I can't participate right now, because I have to do some stuff, but I couldn't help it: I have to say this...

> hallelujah!!

We should totally find a place to gather physically and register this repo as a church, à la [Our Lady of Perpetual Exemption.](https://en.wikipedia.org/wiki/Our_Lady_of_Perpetual_Exemption)

Who is with me?!!!

overdodactyl:
Not a waste at all @earthlng! I was thinking this would be a good way to go as well, sorry I wasn't of more help this PR!

Thorin-Oakenpants:
If we're a registered church... can we [Canonize](https://github.com/ghacksuserjs/ghacks-user.js/issues/535#issuecomment-441651697) shit?

claustromaniac:
Well, I just thought it would be fun to be legally allowed to ask for money and reply with letters like [this](https://c1.staticflickr.com/6/5814/24087531116_9919c9b24c_c.jpg). Then again, we may have to end up [closing](http://www.ourladyofperpetualexemption.com/) anyway.

That won't stop me from treating the `user.js` as our holy book from now though!!!

claustromaniac:
> If we're a registered church... can we [Canonize](https://github.com/ghacksuserjs/ghacks-user.js/issues/535#issuecomment-441651697) shit?

Good prefs that end up deprecated would be good candidates for canonization IMO :angel:

Thorin-Oakenpants:
OMG ... :angel: is now my new fav emoji

earthlng:
Well you have my approval if you want to start something :)

we could probably even get an endorsement from Trump because he thinks *everything* is GREAT.
Although a Trump endorsement could be counter-productive ... unless ... we can also get a Clinton endorsement! somebody got some spare couple hundred K's to "donate" to the Clinton Foundation?
GOD BLESS THE USA FOR PAY-2-PLAY!! hallelujah! USA-USA-USA

earthlng:
too soon?

Issue #545, opened 2018-11-20
ToDo: reassess [WARNING] and [SETUP] tags


Thorin-Oakenpants:
* [x] Note to self: we want a (breakage) tag on security.OCSP.require
* [x] update wiki implementation and troubleshooting pages

Thorin-Oakenpants:
@earthlng umm, I just realized looking at someone's overrides, that `privacy.userContext.enabled` (1702) is default false. Is there any reason now ESR60 is out, that we don't make all or most of 1700s active? I vaguely remember an earlier discussion about it, but can't remember why

Atavic:
*Tip* for *Warning* and *Choose* for *Setup*?

Thorin-Oakenpants:
[Earthlng said](https://github.com/ghacksuserjs/ghacks-user.js/issues/438#issuecomment-440748648)
> They're perhaps hesitant about enabling containers because a lot of people still don't even understand what Private Browsing does and doesn't do.

So there's no *technical* reason anymore that we can't make section 1700 active

Thorin-Oakenpants:
- I think we could restrict `[WARNING]` to items that have serious consequences (and this would probably mean they're all going to be inactive). I can see some existing tags can be changed to `[NOTE]`. eg 0850a
- `[SETUP]` could stay the same (stuff to look at when first getting the user.js), and could even supplement/mirror a relaxed sticky
- Those left could use a `[BREAKAGE]` tag, as long as the updated readme line says that it's not all inclusive (and refers to web content). Good example of this would be 1201

Note to self: we want a (breakage) tag on `security.OCSP.require`

Thorin-Oakenpants:
I'm going thru playing with tags and I think we should do away with [SETUP] completely

`[WEB BREAKAGE]`, `[FF BEHAVIOR]`, and `[WARNING]` should be enough. Setting up means uses can just search for the first two for all the common items to consider (and we wouldn't need to use multiple tags for the same item, which we currently do). The definition of what each tag means would be outlined in the readme numbered items at the top.

so things like session restore, clearing shit on close, locationbar dropdown/keywords not working etc is all about FF internally, UX, and chrome behavior. 

By breaking it into app changes vs web breakage, it should also be easier for end users to troubleshoot.

PS: so far got [warning] down to 8 items (I changed em to setup or breakage at the moment).
- warning: 8
- setup: 27
- breakage: 17

----

[...]

PR #562, opened 2018-11-30
review [TAGS], #545

Thorin-Oakenpants:


Thorin-Oakenpants:
1401 + 1406: in this PR currently have 
`[NOTE] Disabling fonts can uglify the web a fair bit`

I'm OK with changing this to `[BREAKAGE]`. While it doesn't really break anything (maybe some OSes/distros and language settings = some tofu), it would be good to include in the search parameters for end-users - thought?

Thorin-Oakenpants:
:bulb: idea from something @atomGit said. Use `[SETUP-BREAKAGE]` and `[SETUP-BEHAVIOR]`, so all (`[setup`) or either (`kage]`, `vior]`) can be searched on. Personally I don't think it's a huge deal, users will only need to do the initial setup once - two searches is nothing.

earthlng:
> [BEHAVIOR] changes how Firefox itself works, i.e. not directly website related

when I read this I expected there to be ~100 of those tags. Not really liking it. [SETUP] on a few select ones makes sense but BEHAVIOR on only a couple really doesn't IMO.

> [BEHAVIOR] If you want or use Session Restore, then this isn't for you.

I have both of those set and Session Restore works just fine. (?!?)

1202: we don't actively change security.tls.version.min anymore so IMO those 2 lines about breakage can go

1212: breakage due to hard-fail is very rare. Not worth a tag IMO

1222: can "break" Antivirus but not sites or FF itself. NOTE?

1263+1264: breakage is probably very rare at this point. Don't need a tag

can we just remove 2710? People can use a private window, TC or just clear offline data on shutdown (or manually)

4502 addresses something that a lot of people didn't or don't like about RFP. It should be highlighted with a tag

2403 under 9999: breakage and/or warning tags in very old deprecated prefs are not necessary.

Thorin-Oakenpants:
> when I read this I expected there to be ~100 of those tags. Not really liking it.
- I know. I initially had `[FF CHANGE]` but that sucks as well. I really want to differentiate between website breakage and UX/Chrome/workflow type items. We'll see if :cat2: can think of something (and we can make the readme line a bit clearer maybe). [setup] doesn't allow for a decent difference IMO (and I want to get away from setup to making this mostly match (if not completely match) a relaxed sticky.

>  breakage and/or warning tags in very old deprecated prefs are not necessary
- I want information to stay (in general), rather than prune a living document (if that makes sense) - we worked hard for all that :) That said, it's FF44, so yeah, fukit :+1: 

> session restore
- I don;t use sessions myself, so what pref(s) then disable session restore? just `0102`? can you check, in which case I can remove the two and stick one on 0102

earthlng:
> [DISCLAIMER] any breakage due to this user.js is the Russian's fault, sources say

Thorin-Oakenpants:
changes under "tweaks" & "moar tweaks" patches

> > [BEHAVIOR] If you want or use Session Restore, then this isn't for you.
> I have both of those set and Session Restore works just fine. (?!?)
- :heavy_check_mark: removed the bit in `0102`, it still has a behavior tag since it affects recently closed tabs
- :question: does 0102 need one then

> 1202: we don't actively change security.tls.version.min anymore so IMO those 2 lines...
- :heavy_check_mark:  done

> 1212: breakage due to hard-fail is very rare. Not worth a tag IMO
- :x: see #545 which was started all because some had issues with this pref. The thread is in another issue, but I  myself flip this in overrides due to breakage. It's handy for troubleshooting

> 1222: can "break" Antivirus but not sites or FF itself. NOTE?
- :question: ~~I'm confused. I thought if AV breaks the web content cannot be gotten, hence web breakage (that's my limited experience)~~ nvmd, I am thinking about `1260` when MitM can't get the certs. I am probably still confused
- :heavy_check_mark: DONE, changed to warning (because its a security thing, don;t want anyone at risk, hence not a note)

> 1263+1264: breakage is probably very rare at this point. Don't need a tag
- :heavy_check_mark: DONE

> can we just remove 2710? ...
- :x: But if we remove it, then people will say, what about this. Leaving it in to inform users.
- :question:  Change it to a warning instead? "if you disable this, enjoy a broken web"

> 4502 addresses something that a lot of people didn't or don't like about RFP...
- :heavy_check_mark: DONE, changed it to behavior

> 2403 under 9999: breakage and/or warning tags in very old deprecated prefs are not necessary.
- :heavy_check_mark: DONE, see earlier comment, saved a line

can you double check I got all that correctly? Or just re-review. Or something. I need :coffee:

Thorin-Oakenpants:
> I have both of those set and Session Restore works just fine

both? I only changed 1020. Did you mean the other one was 1023? Or did you think we should do something with the ones in 1021+1022?

earthlng:
I don't even know how you can close Firefox with 2 or more windows open

Thorin-Oakenpants:
> I don't even know how you can close Firefox with 2 or more windows open

It remembers multiple window sessions, the last x that are active (in other words the last three closed). As soon as the last window is closed, so is firefox and the x sessions are locked in.

earthlng:
> Are you sure about 1020.

yes. I have both these prefs set to 0 (same as our default) and my last closed window with all its tabs restores just fine. AFAIK the 1020 title is incorrect because I don't think you can really disable the session restore file. (except perhaps with PB-always-on mode) You can tell FF not to restore it on startup but stuff will still get written to it.

Thorin-Oakenpants:
so ignoring all the session restore stuff (see issue #563 ) which we can do next, I'm left with a couple of things (and we can always do more changes later

current PR
```js
/* 2710: disable DOM (Document Object Model) Storage
 * [BREAKAGE] This will break a LOT of sites' functionality AND extensions!
 * You are better off using an extension for more granular control ***/
   // user_pref("dom.storage.enabled", false);
```

I'll change it to a warning. Warning are to be heeded and used sparingly. We currently have 8 and only 6 are for inactive prefs to educate people. This fits that bill. That's one less thing for users to check when troubleshooting.

Thorin-Oakenpants:
moat .. that's it .. typos .. going to bed. Anyway, want :cat2: 's input before committing :nite:

crssi:
1222 should also be default if you use proxy as a fiddler with self-signed certs for decrypting SSL.
Just saying if you would like to mention in description.

Cheers

overdodactyl:
>`[BREAKAGE] known to cause websites to break, or to impact performance`

If possible, it might be nice to split these into `[BREAKAGE]` and `[PERFORMANCE]` or something like that...some people might have a higher tolerance for one than the other.

Thorin-Oakenpants:
I don't really want to fragment tags any more than we have to. Breakage count is now 18, of which only 1 mentions performance: `2508` (HWA)

`2421` (baseline ion, JIT) which from my experience also causes (very rare) actual breakage (eg unable to check checkboxes, or get comboboxes to drop down), and  and `2517` (media capabilities), are both inactive. So I changed them to [NOTE], and if people don't read what it says when they use it, then that's on them. Two less items for people to worry about

That leaves HWA and I feel it's an important one for users to look at

---

I just wanted to split those that affect web content from those that don't. I do not like the word setup (although both tags can be used for that), it's also a troubleshooting thing. And can mostly (or completely) match a relaxed sticky. A hardened sticky could mirror another tag (we're getting a bit complex here ahead of time, because most inactive breakage & some inactive warning = some/most hardening).

My vision for the relaxed sticky is two part:
- FF UX/Chrome etc (non web site related) = relaxing shoulder surfers stuff (urlbar stuff, forms, history, save to desktop), relaxing disk avoidance (cache, session restore, clearing data), internals (extension scopes), and other UX workflow items ("open with")
- actual web site stuff
- and the one outlier IMO, HWA as a perf one (which fits under web site stuff) .. As I type this I can think of other major perf prefs - HTTP2, AltSrv, SSL sessions - but I'm pretty sure we wouldn't add those to a relaxed sticky.

The only way I can accurately describe these is
- make ff BEHAVE like normal
- stop making web sites BREAK

So now I don't know what to do. You all seem to hate my visionary world beating unique take on problem solving

Thorin-Oakenpants:
> `[SETUP] you may want to decide this before first applying these settings`

But breakage is also part of setting up the user.js the first time (and warnings but changing them is bad bad bad, and even notes have significance). Both tags are part of a larger thing called setting up your user.js, which is what the preceding paragraph says. The above implies or at least repeats what has already been said and doesn't signify the difference between the two, at all. 

Come on guys (and gals).. vision people... look at the wood for the trees

Thorin-Oakenpants:
Either come up with a better word, or lets roll with it. :cat2: , go get your thesaurus

PS: I suggested [this](https://github.com/ghacksuserjs/ghacks-user.js/issues/545#issuecomment-443174058)

earthlng:
OT: @claustromaniac your Compare-UserJS is broken for me because I don't have .NET installed. Making shit worse for performance's sake? Do you work at mozilla by any chance? ;)
No seriously, it wasn't slow to begin with so why add another dependency? You also mention that PS *could* perhaps work on linux but I don't know if anyone would want to install .NET (if that's even possible) for a script that worked perfectly fine without it. Just my 2 cents

earthlng:
> they should be expert enough to figure it out themselves? We'll see what E says

I agree. We have a **HTTPS** section and people who need to install a custom cert for things like fiddler should have no problem identifying the pref(s) they need to change

Thorin-Oakenpants:
OT: @claustromaniac Compare also broke for me. If E hadn't piped up, I would have forgotten to tell you. Win7 and it requires no spaces in the path (pretty sure that's what did it) or .bat gives an error.

claustromaniac:
> PS: I suggested [this](url)

That LGTM if you are willing to go with it.

> your Compare-UserJS is broken for me because I don't have .NET installed. 

Maybe install .NET? PowerShell is *built* on .NET. How TF can you even run PowerShell without it?

The [Linux packages](https://github.com/PowerShell/PowerShell#get-powershell) come with bundled dependencies IIRC, and my script uses some archaic classes that existed long before [.NET Core](https://en.wikipedia.org/wiki/.NET_Core) was a thing. The only reason I don't say it works on Linux with certainty is because I didn't test it myself, and so far no one else has informed me if it does or not.

What kind of error is it throwing, if any?

> it requires no spaces in the path (pretty sure that's what did it) or .bat gives an error

Is that true for the last version? I thought I fixed that some time ago (at least it works for me). [Look here](https://github.com/claustromaniac/Compare-UserJS/commit/18493c0eab2d9ffe4757d3980bcd1e1b2e5b8136). Maybe that commit is what broke it for you? if that's the case IDK what to do, because I have that same issue without that change.

Thorin-Oakenpants:
https://github.com/claustromaniac/Compare-UserJS/issues/10

earthlng:
> Maybe install .NET?

and patch that shit every month?! yeah no thanks

> What kind of error is it throwing, if any?

something about [Collections](https://github.com/claustromaniac/Compare-UserJS/blob/master/Compare-UserJS.ps1#L210) not found

Thorin-Oakenpants:
just waiting on the Big E ....bump @earthlng

PR #564, opened 2018-11-30
fix profile check

overdodactyl:
I must not have tested after this change was made:

https://github.com/ghacksuserjs/ghacks-user.js/pull/554#discussion_r236102460

With how it is now, macOS never detects a profile, so I reverted it back and bumped the version number.

PR #565, opened 2018-11-30
updater.sh: add -e option for ESR users

overdodactyl:
(This PR should be looked into after #564)

I'm not sure if we want to include this or not, but I thought it might make a nice addition...

From the wiki:

>Several sections in the user.js have one-char switches to enable/disable the entire section (e.g. deprecated items by ESR version). You will have to remember to do these manually if required.

I've added `-e`, which activates the ESR prefs when used.  It matches based on `/* ESR##.x still uses all`, which I think should be specific enough...that phrase would have to stay consistent in future versions of the `user.js` though (I did a quick glance through some old versions and it seems to have stuck around).

### DIFF
--- a/updater.sh
+++ b/updater.sh
@@ -2,7 +2,7 @@
 
 ## ghacks-user.js updater for macOS and Linux
 
-## version: 2.2
+## version: 2.4
 ## Author: Pat Johnson (@overdodactyl)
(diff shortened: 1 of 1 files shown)


earthlng:
Very nice. Thanks!

Issue #574, opened 2018-12-05
Interactive real-time user.js: Stage 1 [Data Collection]


overdodactyl:
We're trying to build an online, interactive, real-time user.js as an extension of this repository.

The prototype is here: https://nwmviewer.shinyapps.io/ghacks_userjs/

This issue will focus on (1) what data we want to include, and (2) ensuring the quality and accuracy of the data being pulled.  

Feedback and suggestions for what people would find useful to have are always welcome.

Right now what we have, for active prefs:

- [x] ID (text)
- [x] Section (text)
- [x] Preference Name (text)
- [x] Value (text)
- [x] Description (text)
- [x] UI Setting (icon)
- [x] Hidden Pref (icon)
- [x] Links (~icon)
- [x] Tags (icon)
- [x] Notes (icon)
- [x] Tests (icon)

Things we could consider adding data/icons for:

- [ ] Inactive prefs
- [ ] Firefox version specifics (i.e. FF55+, FF57+ ...)
- [ ] OS specifics (e.g. Windows only)

Current issues:

- [ ] Multi-line notes
- [ ] Tags that are indicated by section header only
- [ ] Settings that don't have a user_pref (see 0902/0903)

Thorin-Oakenpants:
how do we define descriptions?

consider
```js
/* 1021: disable storing extra session data [SETUP-CHROME]
 * extra session data contains contents of forms, scrollbar positions, cookies and POST data
 * define on which sites to save extra session data:
 * 0=everywhere, 1=unencrypted sites, 2=nowhere ***/
```

ideally the description should be the first line (minus anything we pull out separately). this keeps it short and simple. That bring up the questions of 1) what about subsequent lines that aren't a note 2) do we have any numbered items which would break this (yes we do, eg 1023)

line1 is borked
```js
/* 1023: set the minimum interval between session save operations - increasing it
 * can help on older machines and some websites, as well as reducing writes, see [1].. etc
``

which then brings up the question of Notes. Notes (and warnings and settings are always after the description/info lines) and are just part of the full description IMO. The "tagging" of them in the user.js just gives them more weight. Since they're always at the end, these tagged items are not a problem.

If we cleanup all first lines for numbered items, then subsequent lines would could be pulled thru as a new data field eg "extrainfo" (and subsequently used how we want in the layout)

Thorin-Oakenpants:
[...]

PR #576, opened 2018-12-07
#574 cleanups

Thorin-Oakenpants:
- FF version tags are now square brackets (allows for [FF52-63] etc, as long as you parse for [FF. Any other references to FF versions if needed, use round brackets (e.g in descriptions, header sections etc)
- FF version tags, where possible, are now after each pref. This removes confusion, and allows for extra new prefs to be inserted under that number. The bulk of these were numbered items with a single pref. Where the version tag was in the pref description but we had multiple prefs, I left those mostly alone, because I cannot be sure they all adhere to the same version release. At some stage it would be nice to enforce ALL version tags to each pref (but that would be a lot of work!).
- Default tags are now all `[DEFAULT*]` and can contain extra info, eg [DEFAULT: true in FF61+]. This means we can ignore tagging things such as the Font information in 1402 etc. The purpose of these is to pick up where we enforce defaults, but we include the value, because of OS differences.
- Hidden tags are now all `[HIDDEN PREF]
- all [OS tags are now all after individual prefs
- Got sidetracked on versions and defaults, but there are some other edits in here. Started to make sure numbered first lines are self contained and the wording doesn't overflow into the next line. Moved or removed a few comments bits from end of pref lines, i.e. `// text` where possible - e.g some setting info up into the man text
- I also did this throughout the entire file (e.g deprecated also got tidied up a bit), even though this is for overdodactyl awesome project which would, when finished, stop at the beginning of the personal section
- Need eyeballs on this to double check I didn't do any typos.

### DIFF
--- a/user.js
+++ b/user.js
@@ -178,9 +178,9 @@ user_pref("app.update.auto", false);
 /* 0302b: disable auto update installing for extensions (after the check in 0301b)
  * [SETTING] about:addons>Extensions>[cog-wheel-icon]>Update Add-ons Automatically (toggle) ***/
 user_pref("extensions.update.autoUpdateDefault", false);
-/* 0303: disable background update service [WINDOWS]
+/* 0303: disable background update service
  * [SETTING] General>Firefox Updates>Use a background service to install updates ***/
-user_pref("app.update.service.enabled", false);
+user_pref("app.update.service.enabled", false); // [WINDOWS]
 /* 0304: disable background update staging ***/
 user_pref("app.update.staging.enabled", false);
 /* 0305: enforce update information is displayed
@@ -444,9 +444,9 @@ user_pref("network.http.speculative-parallel-limit", 0);
  * [2] http://kb.mozillazine.org/Browser.send_pings.require_same_host ***/
 user_pref("browser.send_pings", false);
 user_pref("browser.send_pings.require_same_host", true);
-/* 0607: disable links launching Windows Store on Windows 8/8.1/10 [WINDOWS]
+/* 0607: disable links launching Windows Store on Windows 8/8.1/10
  * [1] https://www.ghacks.net/2016/03/25/block-firefox-chrome-windows-store/ ***/
-user_pref("network.protocol-handler.external.ms-windows-store", false);
+user_pref("network.protocol-handler.external.ms-windows-store", false); // [WINDOWS]
 /* 0608: disable predictor / prefetching ***/
 user_pref("network.predictor.enable-prefetch", false); // [FF48+]
 
@@ -600,13 +600,13 @@ user_pref("browser.formfill.enable", false);
  * This can leak your locale if not en-US
  * [1] https://trac.torproject.org/projects/tor/ticket/21787 ***/
 user_pref("dom.forms.datetime", false);
-/* 0870: disable Windows jumplist [WINDOWS] ***/
-user_pref("browser.taskbar.lists.enabled", false);
-user_pref("browser.taskbar.lists.frequent.enabled", false);
-user_pref("browser.taskbar.lists.recent.enabled", false);
-user_pref("browser.taskbar.lists.tasks.enabled", false);
-/* 0871: disable Windows taskbar preview [WINDOWS] ***/
-user_pref("browser.taskbar.previews.enable", false);
+/* 0870: disable Windows jumplist ***/
+user_pref("browser.taskbar.lists.enabled", false); // [WINDOWS]
+user_pref("browser.taskbar.lists.frequent.enabled", false); // [WINDOWS]
+user_pref("browser.taskbar.lists.recent.enabled", false); // [WINDOWS]
+user_pref("browser.taskbar.lists.tasks.enabled", false); // [WINDOWS]
+/* 0871: disable Windows taskbar preview ***/
+user_pref("browser.taskbar.previews.enable", false); // [WINDOWS]
 
 /*** [SECTION 0900]: PASSWORDS ***/
 user_pref("_user.js.parrot", "0900 syntax error: the parrot's expired!");
@@ -712,9 +712,9 @@ user_pref("browser.sessionstore.resume_from_crash", false);
  * This longer interval *may* affect history but we cannot replicate any history not recorded
  * [1] https://bugzilla.mozilla.org/1304389 ***/
 user_pref("browser.sessionstore.interval", 30000);
-/* 1024: disable automatic Firefox start and session restore after reboot [WINDOWS]
+/* 1024: disable automatic Firefox start and session restore after reboot
  * [1] https://bugzilla.mozilla.org/603903 ***/
-user_pref("toolkit.winRegisterApplicationRestart", false); // [FF62+]
+user_pref("toolkit.winRegisterApplicationRestart", false); // [FF62+] [WINDOWS]
 /** FAVICONS ***/
 /* 1030: disable favicons in shortcuts
  * URL shortcuts use a cached randomly named .ico file which is stored in your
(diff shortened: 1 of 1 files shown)


Thorin-Oakenpants:
I have to say the compare in github on this one sucks. You would be better off downloading it and doing a compare locally with your favorite IDE

overdodactyl:
It's looking good to me!  If we're prettying certain on going this route (which I think we are), I can switch the prototype to grabbing this file instead so I can start working with the changes.

Thorin-Oakenpants:
Yup, got for it. I won't commit it until at least E and :cat2: sign off. And you might pick up some new issues

Thorin-Oakenpants:
> I still read it, but it wasn't as easy as I thought it would be

Yeah. I said as much. It seems to jumble a few lines out of order

Thorin-Oakenpants:
That makes all `[FF` info all assigned to user_pref lines

earthlng:
I really don't like the removal of FF version (and OS info, etc) from the title lines but I guess it's too late now

Thorin-Oakenpants:
^^ no, it's not too late (it kinda hurt me a little as well). Where all multi-prefs under the number are the same, we could duplicate it back in the top line (it's probably all of them). I'm fine with that. Because @overdodactyl 's code will only parse for them in the pref lines, it's no big deal.

Happy to do that before we commit - just let me know (as long as we remember to remove it if the multi-prefs differ with version)

PS: am on beer # 8

Thorin-Oakenpants:
Or we could MAKE @overdodactyl just propagate it and stick his code in to compare etc

Thorin-Oakenpants:
I've just been looking at this for the last 2 hours, and in order to come to some sort of consensus, or informed decision, did some number crunching, in a pretty spreadsheet.

Considerations
- duplication (I'm not a fan of duplication)
- usability (I find it easier to spot the FF version info at the end of the first line), it seems more elegant
- we can make code do anything, overdodactyl already did the [FF` versions this way before
   * and is looking at doing the same for propagating [SETTING etc to children prefs
   * and if we put them back, ONLY if all multi-prefs have the same version, then this removes the need to compare, so the code from before is actually simplified a bit, just not to the same extent
- the user.js comes first, the cool real-time whizzy-ma-jiggy comes second (but we'll help if we can, this has already led to some content cleanup)
- consistency of where shit goes for end users

Here's what I found (up to but excluding 5000s)
```
 101 numbers containing 143 [FF tags
---
  71 numbers with a single pref
      71 tags, movestotopline=77, duplicityreduction= 0

   7 numbers where some multi-prefs don't have a tag so they can't move
       8 tags, movestotopline= 0, duplicityreduction= 0

   9 numbers where tags are variable so they can't move
      32 tags, movestotopline= 0, duplicityreduction= 0

  14 numbers where all multi-prefs have a tag and the tag is the same
      32 tags, movestotopline=14, duplicityreduction=18

 TOT (check)
 ===
 101 numbers
     143 tags, movestotopline=91, duplicityreduction=18 tags removed 
```

Although this would also include deprecated, for some reason I didn't bother to analyze that. **91 numbered items**, that's a lot of items to improve "readability" (although I could counter that as end users we would get used to it), and it also removes duplication: **18** tags would be saved (duplication currently stands at 12½%, do we want that? if every pref had a version that would be 60 odd extra tags). Also, we would save characters when we can do away with the ` //` (just sayin).

So, knowing that @overdodactyl can still achieve an end result regardless of what we do with these (although we would still keep the square brackets, and it cleans up having a 1st line tag with other release tags in the user_pref line - i.e if the 1st line has one, apply to all children)

It really comes down to consistency IMO. Either we always stick them against each pref, or we keep our smarter hybrid system.

I'm leaning towards our hybrid (it's more elegant). I know E wants status quo. Can I get some opinions? @overdodactyl @claustromaniac

earthlng:
`[SETTING] Privacy & Security>SecurityDeceptive Content and Software Protection` - 3 times

Thorin-Oakenpants:
that's because there are three numbered items, previously they all had the same string "(under Options>Privacy & Security)" in the first line.

It's just cleaner on a new line, and it's a setting (so tag it). Do we really need to list the full thing - it gets duplicitous and long

eg " * [SETTING] Privacy & Security>SecurityDeceptive Content and Software Protection>Warn you about unwanted and uncommon software" would be the longest one, at 128 chars and they only repeat what the description says

earthlng:
it's missing a `>`, that's all :)

Thorin-Oakenpants:
Do you mean to include it in the `/** SAFE BROWSING` header **instead of** where they are? The 0410->0416's are more than just the three prefs we're talking about (but the setting does relate to what the header talks about 100%). How would you word it, "for 0410, 0411, 0412, go to blah blah and read the bit in quotes so understand what I mean"?

 I think if we are going to give setting info, it should be with the pref where possible. It's kinda cool as one of the filters on the prototype, to be able to map everything to UI

Thorin-Oakenpants:
We can do it in this pref, or I can make a note to do it later - what about 2002, one of the prefs is "obsolete"? I can't remember the story around it. When the FF51 pref was added did we keep the old one for ESR45 users?

```
user_pref("media.peerconnection.ice.default_address_only", true); // [FF42-FF50]
```

Thorin-Oakenpants:
I've got a copy here with the changes reversed where appropriate. But it does show that in the past we have not followed this, eg FFv in 1st line and then variable ones for some of the children prefs.

And then it brings up another issue. Look at 0510 disable pocket: it says FF39+, bu that's the old pref that was deprecated. The new pref started in FF46. If the FF version is applied to each pref, then it removes all ambiguity AND prevents future fuckups

overdodactyl:
>and if we put them back, ONLY if all multi-prefs have the same version

This is the biggest thing IMO.

Code wise it wouldn't make a big difference where tags are located (in-line or between `/* .... ***/`), so long as the info in the header applies to every subsequent preference.  Generally speaking, I think this is just good practice regardless to remove any confusion for the `user.js` file itself.

Thorin-Oakenpants:
> but I wouldn't mind some duplication

18 items dude .. 

¯\_(ツ)_/¯ IDK .. lets toss a coin

Thorin-Oakenpants:
> so long as the info in the header applies to every subsequent preference

AGREED, and this is where I picked up on a couple that didn't. :heavy_check_mark: will triple check

Will reverse back to our hybrid system.

Thorin-Oakenpants:
OK, quadruple checked it locally. Even the github compare is much cleaner (90% of the changes are just tag format changes, eg square brackets). But it does show that we had inconsistencies in where we put tags, eg in deprecated we even had a pref line that should have been in the 1st line.

Awaiting reviewers

overdodactyl:
It's looking good to me.  Only thing that caught my eye was:

0510: disable Pocket (FF39+) --> 0510: disable Pocket [FF46+]

Was the version change intentional?

Thorin-Oakenpants:
scroll up (and I said so in the commit) :)

> And then it brings up another issue. Look at 0510 disable pocket: it says FF39+, bu that's the old pref that was deprecated. The new pref started in FF46.

deprecated FF46
```
// 0510: disable "Pocket" - replaced by extensions.pocket.*
   // [-] https://bugzilla.mozilla.org/1215694
```

actually, I should add [FF39+] to the old one

Thorin-Oakenpants:
you've got a real **thing** for emojis there :cat2: 

bump: waiting on big E's review... when he can, no big deal. Just can't really do any other commits otherwise it gets messy reconciling differences

earthlng:
LGTM too. Thanks Pantski

Thorin-Oakenpants:
Hellalulah!!

Issue #575, opened 2018-12-06
1020: SR max_undo* & 1022: SR after crash : make inactive or remove


Thorin-Oakenpants:
Background:
- Session Restore (SR) cannot be "disabled" in normal window mode (in PB mode it is not used)
- It is tied to history (I will test what effect disabling history has)
- The two prefs only control how much to restore, not what is collected in `recovery.jasonlz4`
   * I have tested this, with both set to 0, It still recorded info on multiple tabs and multiple windows
   * side note: it keeps a history per tab in there as well (I assume it obeys 0804)

I always mistakenly thought the prefs meant local persistent data would be restricted, but that is clearly not the case. 

I can do some tests re history settings, and this may be the way to control or limit the collection of it (local persistent data) - see ToDo below. But actually limiting what is restored is pointless - if someone has control of your computer, they already have the info should they dig - e.g computer forensics. And that was the *only* reason I ever had these prefs as active. There is a very very slight edge of shoulder surfing when opening FF (or resuming from a crash), but if the user wants to use SR, they want to use SR and do not care - these "active" prefs are just yet another thing for people's overrides and extra hoops to jump through.

Suggestions
- make 1020 + 1023 inactive
- remove them as they have zero to do with privacy/security

ToDo:
- Check if disabling history stops any data collection in `recovery.jasonlz4` and this info could be added to the relevant pref

Personally, I am in favor of removing them.

claustromaniac:
It seems @earthlng was the one who [initially suggested](https://github.com/ghacksuserjs/ghacks-user.js/issues/51#issuecomment-287061254) adding `browser.sessionstore.interval`. I'll quote: 

> for the sake of SSD's we should probably enable this with a reasonably high value. (I have 60000, but even 30sec is much better than the default 15sec)
> https://www.servethehome.com/firefox-is-eating-your-ssd-here-is-how-to-fix-it/

It is true that this is not related to privacy or security, but it is useful (for that, at least). How about moving it to `5000`?

Thorin-Oakenpants:
my mistake for typing `1023` .. I meant for resume from crash (1022), my bad - will edit title and OP

claustromaniac:
and my mistake for not paying more attention. But mine is justified because I'm a foreign cat. (yeah, right)

earthlng:
> The two prefs only control how much to restore, not what is collected in

something that's not stored cannot be restored. The 2 prefs control the number of **closed** tabs/windows to STORE.

[...]

PR #583, opened 2018-12-10
Update user.js

earthlng:
things not in the PR:

- 0911 - remove? or replace with `network.auth.subresource-http-auth-allow` (would need to add it to troubleshooter as well)
- 1022 - does it even do what we expect it to do? remove?
- 1700 - SETUP-CHROME and maybe add a recommendation for TC?
- 2002 - add test page?
- 2302 - [FF32+] - added in 32, enabled in 44 (still disabled in ESR60!)

### DIFF
--- a/user.js
+++ b/user.js
@@ -982,8 +982,8 @@ user_pref("privacy.userContext.enabled", true);
 /* 1703: enable a private container for thumbnail loads [FF51+] ***/
 user_pref("privacy.usercontext.about_newtab_segregation.enabled", true); // [DEFAULT: true in FF61+]
 /* 1704: set behaviour on "+ Tab" button to display container menu [FF53+] [SETUP-CHROME]
- * 0=disables long press, 1=when clicked, the menu is shown, 2=the menu is shown on long press
- * [NOTE] The menu does not contain a non-container tab option
+ * 0=no menu (default), 1=show when clicked, 2=show on long press
+ * [NOTE] The menu does not contain a non-container tab option (use Ctrl+T to open non-container tab)
  * [1] https://bugzilla.mozilla.org/1328756 ***/
 user_pref("privacy.userContext.longPressBehavior", 2);
 
--- a/user.js
+++ b/user.js
@@ -981,7 +981,7 @@ user_pref("privacy.userContext.ui.enabled", true);
 user_pref("privacy.userContext.enabled", true);
 /* 1703: enable a private container for thumbnail loads [FF51+] ***/
 user_pref("privacy.usercontext.about_newtab_segregation.enabled", true); // [DEFAULT: true in FF61+]
-/* 1704: set long press behaviour on "+ Tab" button to display container menu [FF53+]
+/* 1704: set long press behaviour on "+ Tab" button to display container menu [FF53+] [SETUP-CHROME]
  * 0=disables long press, 1=when clicked, the menu is shown
  * 2=the menu is shown after X milliseconds
  * [NOTE] The menu does not contain a non-container tab option
--- a/user.js
+++ b/user.js
@@ -50,7 +50,7 @@
      0700: HTTP* / TCP/IP / DNS / PROXY / SOCKS etc
      0800: LOCATION BAR / SEARCH BAR / SUGGESTIONS / HISTORY / FORMS
      0900: PASSWORDS
-     1000: CACHE
+     1000: CACHE / SESSION (RE)STORE / FAVICONS
      1200: HTTPS (SSL/TLS / OCSP / CERTS / HSTS / HPKP / CIPHERS)
      1400: FONTS
      1600: HEADERS / REFERERS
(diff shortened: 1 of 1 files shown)


earthlng:
just a couple things I'd like to change ...

:jeans:, :cat2:

Thorin-Oakenpants:
see https://github.com/ghacksuserjs/ghacks-user.js/issues/575 OP .. 1022 = inactive or remove (I'm fine with removing it)

earthlng:
the only reason why it could make sense to disable it IMO (assuming it even does disable session restore after crash!) would be if a tab manages to take down the whole browser you probably don't want to restore that tab

Thorin-Oakenpants:
1260: that's web breakage - if sha-1 is restricted, pages won't load .. web breaks. Just like not allowing DRM breaks video, video will not load. I know we can argue semantics here.

`user_pref("devtools.webide.adaptersAddonURL", "");` - why add a deprecated in FF57 pref? I don't mind, just wondering why. And is that in the `[-]` reference link?

The rest looks fine on first glance. I still need to go thru [SETUP* and make sure we have things in the right places. end of 1st line if ok, but in body if the next comment specifically is about that. So yeah, those moves look ok as well

earthlng:
> Useful if repeated crashes on restore are caused by a malicious tab

that implies that it does what we *think* it does. You're the one with the scratchpad to crash FF ... :)

Thorin-Oakenpants:
> 1700 - SETUP-CHROME and maybe add a recommendation for TC

No need to add setup tag (and I am going to get those all out of headers). Recommendation for TC, IDK - it hasn't really been updated for like 6 months from memory - mostly lint and things. The issues are piling up.

Thorin-Oakenpants:
Can I do a little editing to undo a couple of things and push this thru. I'd like to address the others one by one - i think my comment about not wanting make a note/todo list before was ambiguous. I meant for a single change it wasn't worth opening an issue.

earthlng:
> If 1260 causes a problem, what does the end user experience?

AFAIK that pref alone does nothing if you don't also set a proxy, fe fiddler. So most people will never have a problem with that pref and people who use fiddler or similar should have no problem knowing where to look to make it work. (1200 = HTTPS, CERTS, etc)
IMO we can just drop that tag all together so we don't have to argue WEB vs CHROME :)

earthlng:
> Can I do a little editing to undo a couple of things and push this thru

sure

earthlng:
> re: 1700 - "No need to add setup tag"

I mean currently due to needing to long-press to even have the option to use containers show up on `+`, some (most?) people won't even notice that we enabled containers. That's why I think it needs a setup tag of some kind.

earthlng:
I'll update the deprecated sticky

Thorin-Oakenpants:
> I mean currently due to needing to long-press to even have the option to use containers show up on +, some (most?) people won't even notice that we enabled containers. That's why I think it needs a setup tag of some kind.

Oh, i get ya. We can add one to the long press pref if you want. As for noticing it's on - users might notice them in right click contexts, and it will show in options as well. That should suffice

earthlng:
maybe change
`1704: set long press behaviour on "+ Tab" button to display container menu`
to
`1704: set behaviour on "+ Tab" button to display container menu`
and `2=the menu is shown on long press` ?

earthlng:
`0=no menu (default), 1=show when clicked, 2=show on long press`
and
`[NOTE] The menu does not contain a non-container tab option (use Ctrl+T to open non-container tab)`

claustromaniac:
- 0911 - remove? or replace with... :cat2:: replace!  (with `user_pref("network.auth.subresource-http-auth-allow", 0);` or `1`? I'd go with `0`, but IDK I'M NEW AT THIS SHIT!!! :scream_cat: \*freaking out\*).

> Recommendation for TC, IDK - it hasn't really been updated for like 6 months from memory - mostly lint and things. The issues are piling up.

who cares about a bunch of minor issues and feature requests not getting solved as long as the shit works? especially when we talk about such a useful extension! My own shitty extensions are heading the same way, except they're not nearly as useful!!! :sweat_smile:  It is already recommended on the wiki though, so it doesn't matter anyway.

Thorin-Oakenpants:
[NOTE] kinda lends weight (while our active value is the best so as to not spring nasty surprises - if FF ever flip containers on, I bet they go for 2 as well). I like that we added a setup tag to draw attention to it (the setting as well as the fact we turned containers on). The other prefs make no difference.

And, I want to get [setup tags against prefs. They don;t work with the real-time thingy if they are stuck in section headers. That doesn't mean we can't *also* have them in headers, but it creates duplicity. At the end of the day, I want users in the js and on the whizzy-do dactyl site, to be able to jump from pref to pref and make changes or troubleshoot. The less in section headers the better

Thorin-Oakenpants:
OK, I need to DL a copy and compare with my local master, then I'll start a new issue, or commandeer one, and add in all the things we haven't addressed from here

earthlng:
re: 0911

`pref("network.auth.subresource-img-cross-origin-http-auth-allow", true);` only prevents IMAGES from triggering a login prompt.
It was added in FF55 and is default false since FF59 = no need to enforce it anymore. We probably added it before it was set to false

`pref("network.auth.subresource-http-auth-allow", 2);` was added in FF41 and supports the following values:
```js
// Sub-resources HTTP-authentication:
//   0 - don't allow sub-resources to open HTTP authentication credentials dialogs
//   1 - allow sub-resources to open HTTP authentication credentials dialogs,
//       but don't allow it for cross-origin sub-resources
//   2 - allow the cross-origin authentication as well.
```

AFAIK HTTP Auth is pretty rare nowadays and 0 should probably be fine. 1 = less breakage

Thorin-Oakenpants:
> who cares about a bunch of minor issues and feature requests

Yeah I'm fine with it being recommended. Was just querying its status (I don't use it). From the outside looking in, I was just querying it's status. For all I know, it's stable AF. I'll defer to you and :earth_americas: since you gals use it. It's on the task list FYI

PR #586, opened 2018-12-11
move 0370 to 0105b

earthlng:
The link we had in 0370 is not very useful for end users.
https://abouthome-snippets-service.readthedocs.io/ is better

I did some digging into what exactly snippets are and how it works ...

Snippets = information about mozilla and firefox, like tips + tricks and things they want to highlight or advertise, etc. (~4MB of data with CSS, HTML + JS)
If you want to see the data, load [this](https://snippets.cdn.mozilla.net/5/Firefox/63.0/20181018182531/WINNT_x86_64-msvc/en-US/release/Windows_NT%206.1/default/default/) and look at the response.

"Updates from Mozilla and Firefox", as it says in the UI, doesn't sound that bad but it includes [metrics](https://abouthome-snippets-service.readthedocs.io/en/latest/data_collection.html#metrics), geolocation stuff and runs JS code received from a server (AKA remote code execution).

When I saw that it allows RCE I tried to inject code via the URL pref but "data" URIs don't work anymore because Activity Stream is now protected with a CSP.

Should we add something like `Runs code received from a server (aka Remote Code Execution) and sends information back to a metrics server`? Because "Updates from Mozilla and Firefox" makes it sound kind of tempting and a lot less shitty than it really is, IMHO.

ps: do we really want all those `// has setting` ?

### DIFF
--- a/user.js
+++ b/user.js
@@ -105,13 +105,15 @@ user_pref("browser.newtab.preload", false);
 user_pref("browser.newtabpage.activity-stream.feeds.telemetry", false);
 user_pref("browser.newtabpage.activity-stream.telemetry", false);
 user_pref("browser.newtabpage.activity-stream.telemetry.ping.endpoint", "");
-/* 0105b: disable AS Snippets ***/
+/* 0105b: disable AS Snippets
+ * [1] https://abouthome-snippets-service.readthedocs.io/ ***/
 user_pref("browser.newtabpage.activity-stream.disableSnippets", true);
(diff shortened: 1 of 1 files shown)


Thorin-Oakenpants:
> do we really want all those // has setting ?

0105 has a generic [SETTING] tag, but the UI only really shows/hides things and the bulk of the prefs have no UI presence. So it was an attempt I guess at indicating that. I'm OK with dropping it. AS is a morphing beast and the less upkeep the better

PR #587, opened 2018-12-11
all Deprecations + new ADB extension prefs

earthlng:


Thorin-Oakenpants:
I need to do a 63 final first.

PR #589, opened 2018-12-12
disable Telemetry Coverage

Thorin-Oakenpants:


earthlng:
you messed up the tag and it's missing the hidden pref tag.
IMO we only need the opt-out pref because .enabled is default false and it's not like super bad *if* it would somehow send that one-time anonymous ping despite having opted out.

PR #590, opened 2018-12-12
taking out the garbage

earthlng:
- don't need these peerconnection prefs other than the main one
- the chance that somebody ends up in WebIDE despite the menu being hidden with .enabled is already slim to none. But if they do then autoinstallADBExtension stops the download of the ADB extension. And even if it didn't - just uninstall it again. NP! the super paranoid and mozilla haters can add the URL in their overrides if they want.

### DIFF
--- a/user.js
+++ b/user.js
@@ -1019,6 +1019,7 @@ user_pref("_user.js.parrot", "2000 syntax error: the parrot's snuffed it!");
 /* 2001: disable WebRTC (Web Real-Time Communication)
  * [1] https://www.privacytools.io/#webrtc ***/
 user_pref("media.peerconnection.enabled", false);
+user_pref("media.navigator.video.enabled", false); // video capability for WebRTC
 /* 2002: limit WebRTC IP leaks if using WebRTC
  * [TEST] https://browserleaks.com/webrtc
  * [1] https://bugzilla.mozilla.org/buglist.cgi?bug_id=1189041,1297416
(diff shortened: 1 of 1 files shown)


Thorin-Oakenpants:
Just to be clear, you want `devtools.remote.adb.extensionURL` removed which you added in https://github.com/ghacksuserjs/ghacks-user.js/commit/61be5ae563e0e7e3427885a16e6f920437797966 just a day ago?

earthlng:
yeah I know, kinda stupid but that's me sometimes :)

Thorin-Oakenpants:
`media.navigator.video.enabled` I think should be left in .. there's a tor ticket about this (I'll see if I can find it again). But the others, I trust your code reading skills - we'll have to add em to removed scratchpad too

Thorin-Oakenpants:
OK, I'll edit Diffs and shove that pref back to ignore

Thorin-Oakenpants:
I think this is the ticket: https://trac.torproject.org/projects/tor/ticket/16328 . FWIW: TBB 8.* doesn't have this pref (at least not visible), but they do have media.peerconnection.enabled

Thorin-Oakenpants:
I'll leave `media.navigator.video.enabled` in, but make a note about in the misc issue

earthlng:
that ticket doesn't mention `media.navigator.video.enabled` at all ... ?

Thorin-Oakenpants:
True, I'm getting zoned out with pain (I've been thru childbirth, just the once .. and this is up there) .. it mentions `media.navigator.enabled` .. sorry mate

I'll just do a direct commit. And we have `media.navigator.enabled`=false, I assume that's the real master switch

PR #595, opened 2018-12-13
disable System Add-on updates

earthlng:
`https://github.com/pyllyukko/user.js/issues/419`

### DIFF
--- a/user.js
+++ b/user.js
@@ -385,10 +385,9 @@ user_pref("network.allow-experiments", false);
 user_pref("app.normandy.enabled", false);
 user_pref("app.normandy.api_url", "");
 user_pref("app.shield.optoutstudies.enabled", false);
-/* 0505: disable System Add-on updates
- * [NOTE] In FF61 and lower, you will not get any System Add-on updates except when you update Firefox ***/
-   // user_pref("extensions.systemAddon.update.enabled", false); // [FF62+]
-   // user_pref("extensions.systemAddon.update.url", "");
+/* 0505: disable System Add-on updates ***/
+user_pref("extensions.systemAddon.update.enabled", false); // [FF62+]
+user_pref("extensions.systemAddon.update.url", ""); // [FF44+]
 /* 0506: disable PingCentre telemetry (used in several System Add-ons) [FF57+]
  * Currently blocked by 'datareporting.healthreport.uploadEnabled' (see 0333) ***/
 user_pref("browser.ping-centre.telemetry", false);


Thorin-Oakenpants:
@earthlng I think we should put this back in to 335. I keep saying it .. future-proofing. Because code fuckups have happened in the past (I forget the exact details now, it was health reports), and will happen again - and despite efforts to combine almost everything under the telemetry master switch (eg experiments), they keep going outside this - pingsender, coverage, what next.

As long as blanking an URL doesn't affect updates for essential services such as blocklists, TP, SB etc , then I say fuck it - blank it. I'm adding this in... pref("toolkit.coverage.endpoint.base", "");

earthlng:
they've collected coverage data for at least 3 months already. During all that time the URL was hardcoded. FYI the opt-out pref before FF64 apparently was `toolkit.telemetry.coverage.opt-out`

The best way to stop this and all similar shit like that in the future is to disable system addon updates.

Thorin-Oakenpants:
^^ just playing devil's advocate ... but sure . icnsistencies :beer: i didn't delve too deep

earthlng:
add `toolkit.telemetry.coverage.opt-out` ?

Thorin-Oakenpants:
If you want to. It was in my original PR. I think killing the endpoint should be enough which I did in https://github.com/ghacksuserjs/ghacks-user.js/commit/f6ea20a8b09b128a0b7d2b8b3067a4334622d5a8, but why not. Just do a commit. Up to you, but yeah .. I'm starting to think the more fallbacks we have the better, because clearly some mechanisms are being deliberately bypassed now (I do get why this needs to be outside normal telemetry, but the odd code fuckup from years ago, which can still happen is one thing. But Telemetry as whole, is now being actively bypassed and in some cases mispresented IMO)

earthlng:
> It was in my original PR.

no it wasn't: https://github.com/ghacksuserjs/ghacks-user.js/pull/589/commits/03f51f174149e28106e635b926ea256502c86b9e

As I said, clearing the URL doesn't do anything when the system addon uses a hardcoded URL.

`https://github.com/mozilla/one-off-system-add-ons/pull/131` ...

> This change makes the Telemetry Coverage extension work as far back as Firefox 45.

ie if they're pushing this to ESR and older, it will apparently use a hardcoded URL but should honor `toolkit.telemetry.coverage.opt-out`.

earthlng:
https://github.com/ghacksuserjs/ghacks-user.js/pull/600

Thorin-Oakenpants:
> no it wasn't: https://github.com/ghacksuserjs/ghacks-user.js/pull/589/commits/03f51f174149e28106e635b926ea256502c86b9e

soz, so many fucking changes and merry-go-rounding

PR #597, opened 2018-12-14
google analytics

bogachenko:
need?

### DIFF
--- a/user.js
+++ b/user.js
@@ -1283,6 +1283,10 @@ user_pref("mathml.disabled", true);
  * [1] https://trac.torproject.org/projects/tor/ticket/10089
  * [2] http://kb.mozillazine.org/Middlemouse.contentLoadURL ***/
 user_pref("middlemouse.contentLoadURL", false);
+/* Preferences for the new performance panel This pref configures the base URL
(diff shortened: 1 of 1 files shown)


bogachenko:
@Thorin-Oakenpants @earthlng @claustromaniac bump

Atavic:
[Source](https://github.com/devtools-html/perf.html/tree/624f71bce5469cf4f8b2be720e929ba69fa6bfdc/res)

Gecko Profiler is a performance tool to record profiles, perf-html.io visualizes this performance data. Both are [Developer Tools](https://github.com/devtools-html) and work on [Nightly by default](https://github.com/devtools-html/Gecko-Profiler-Addon#development) or from [console code entry](https://github.com/devtools-html/perf.html/blob/master/docs-developer/gecko-profile-format.md#running-the-gecko-profiler-independently-from-the-add-on-and-client). The file generated is called *profile.sps.json*

claustromaniac:
From the snippet you posted that seems to be ignored when DNT is enabled (which we enforce). Not sure if :jeans: or :earth_africa: will want to add it anyway.

Issue #585, opened 2018-12-10
random stuff from the Big E


Thorin-Oakenpants:
:small_red_triangle_down: stuff :earth_africa: brought up

* [ ] remove 0906, 0907, 0908, 0910 - see [comment](https://github.com/ghacksuserjs/ghacks-user.js/issues/585#issuecomment-447243866) below
   * at this stage, `nah` .. see [reply](https://github.com/ghacksuserjs/ghacks-user.js/issues/585#issuecomment-447846238) to comment below, but feel free to keep discussing it
* [x] `0911` remove `network.auth.subresource-img-cross-origin-http-auth-allow`
   * default false in FF59+ anyway - https://github.com/ghacksuserjs/ghacks-user.js/commit/88b747ef36185c460b87a5ca5cd7d964a2dd597e
   * [x]  add to removed prefs scratchpad - https://github.com/ghacksuserjs/ghacks-user.js/commit/f0fbfd3086311548b8863a2a8d0e50593b76b306
* [x]  Add `network.auth.subresource-http-auth-allow` - https://github.com/ghacksuserjs/ghacks-user.js/commit/5bd5f6b28e801b8437e2574fad35f52365a6b593
   * [x] add it to the troubleshooter - https://github.com/ghacksuserjs/ghacks-user.js/commit/5b0952f60a5e9ed0fd62499187c6e62829fcd587
* `1022` does it even do what we expect it to do? remove?
     * this is covered under #575 
* [x] `1700` - add a recommendation for TC - https://github.com/ghacksuserjs/ghacks-user.js/commit/d5ece0f6f44896461a22c0dce46427aac77fcd19
* [x]  `2002 `add test page - https://github.com/ghacksuserjs/ghacks-user.js/commit/778dc89bb6126212d9c897f6e7add88ca5e6a96e
* [x] `2302` add version tag - added in 32, enabled in 44 (still disabled in ESR60!) - https://github.com/ghacksuserjs/ghacks-user.js/commit/23733097a95d9db37028d0ec8c7991bd0bf0a04b

That's right, the three wise "men" now all have emoji's :jeans: :earth_asia: :cat2: 

PS: I don't think I missed anything, but feel free to add anything else minor. We can use this as a miscellaneous checklist.

---

:cat2: said
[...]

PR #602, opened 2018-12-17
0912: HTTP Auth sub-resources #585

Thorin-Oakenpants:
if you're happy squash & merge


earthlng:
if you're happy squash & merge

PR #603, opened 2018-12-17
0900s: add some default tags, #585

Thorin-Oakenpants:
- although we *may* remove some of these prefs (but I doubt it, but persuade me in the open issue).
- I also don't want to pollute the js with tags on every fucking line. You started adding these for your diffs, right?. I'm not sure exactly how adding them to some prefs but not others helps you?
- that said, it is a nice way to show that we're *enforcing* something and does help end users, as well as the shiny overdactyl project

### DIFF
--- a/user.js
+++ b/user.js
@@ -634,17 +634,17 @@ user_pref("signon.autofillForms", false);
 user_pref("signon.storeWhenAutocompleteOff", true); // [DEFAULT: true]
 /* 0907: display warnings for logins on non-secure (non HTTPS) pages
  * [1] https://bugzilla.mozilla.org/1217156 ***/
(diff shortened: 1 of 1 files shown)


Thorin-Oakenpants:
OK, lets rethink this ... we should make sure `[default` is for active prefs only (maybe an exception)? and only stick them in for important shit... I'll close this, we stick have the issue open to re-address it

Thorin-Oakenpants:
I think we'll skip this in favor of a future step in cleaning up, improving tags as part of improving the shiny real time dactyl endeavor - and the project would say that they're selective (for emphasis on importance). Once we can start to filter the page, we can see how effective it is, and what's missing

Issue #528, opened 2018-10-31
ToDo: AS overhaul


Thorin-Oakenpants:
To do after we get 63-beta out

- move it out of system addons section (no need to cater for ESR, we have an archived v60 for that)
- expand/check on all the prefs (although old, see #396 )

claustromaniac:
FWIW I have this on FF 64. Some of them are absolutely not necessary, but whatever.
<details><summary> clickey </summary><p>



```js
// ACTIVITY STREAM
user_pref("browser.newtabpage.activity-stream.asrouter.devtoolsEnabled", false);
user_pref("browser.newtabpage.activity-stream.default.sites", "");
user_pref("browser.newtabpage.activity-stream.disableSnippets", true);
user_pref("browser.newtabpage.activity-stream.feeds.aboutpreferences", false); // FF61+
user_pref("browser.newtabpage.activity-stream.feeds.asrouterfeed", false);
user_pref("browser.newtabpage.activity-stream.feeds.favicon", false);
user_pref("browser.newtabpage.activity-stream.feeds.messagecenterfeed", false); // FF61+
user_pref("browser.newtabpage.activity-stream.feeds.migration", false);
user_pref("browser.newtabpage.activity-stream.feeds.newtabinit", false);
user_pref("browser.newtabpage.activity-stream.feeds.places", false);
user_pref("browser.newtabpage.activity-stream.feeds.prefs", false);
user_pref("browser.newtabpage.activity-stream.feeds.section.highlights", false);
user_pref("browser.newtabpage.activity-stream.feeds.section.topstories", false);
user_pref("browser.newtabpage.activity-stream.feeds.section.topstories.options", "");
user_pref("browser.newtabpage.activity-stream.feeds.sections", false);
user_pref("browser.newtabpage.activity-stream.feeds.snippets", false);
user_pref("browser.newtabpage.activity-stream.feeds.systemtick", false);
user_pref("browser.newtabpage.activity-stream.feeds.telemetry", false);
user_pref("browser.newtabpage.activity-stream.feeds.theme", false);
user_pref("browser.newtabpage.activity-stream.feeds.topsites", false);
user_pref("browser.newtabpage.activity-stream.filterAdult", false);
user_pref("browser.newtabpage.activity-stream.fxaccounts.endpoint", "");
user_pref("browser.newtabpage.activity-stream.improvesearch.noDefaultSearchTile", false);
[...]

PR #604, opened 2018-12-17
0105s: disable AS Improve Search Experiment

Thorin-Oakenpants:
see https://github.com/ghacksuserjs/ghacks-user.js/issues/528#issuecomment-447933180 and subsequent comment

### DIFF
--- a/user.js
+++ b/user.js
@@ -117,7 +117,10 @@ user_pref("browser.newtabpage.activity-stream.feeds.snippets", false);
 user_pref("browser.newtabpage.activity-stream.feeds.section.topstories", false);
 user_pref("browser.newtabpage.activity-stream.section.highlights.includePocket", false);
 user_pref("browser.newtabpage.activity-stream.showSponsored", false);
-/* 0105d: disable Activity Stream recent Highlights in the Library [FF57+] ***/
+/* 0105d: disable Activity Stream Improve Search Experiment ***/
+user_pref("browser.newtabpage.activity-stream.improvesearch.noDefaultSearchTile", false);
+user_pref("browser.newtabpage.activity-stream.improvesearch.topSiteSearchShortcuts", false);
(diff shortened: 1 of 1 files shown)


earthlng:
why are you testing in an early beta?
Yeah there are obviously still bugs with this "experiment" and that's why it's still disabled in Release.

Have you actually clicked on one of these things? all it does is put `@amazon` in the urlbar for easy searching. Granted that's a stupid thing to "teach" newbies because you can assign a keyword like `a` for amazon and search with `a firefox movie`. But they obviously want to push AS and make things easy for the casuals. Sadly we disable that very useful search keyword feature with 0850 but that's another issue.

And besides, we already disable AS anyway so what's the big deal? Just more unnecessary double and triple-fallbacks for something that's not even privacy or security related.

Thorin-Oakenpants:
> And besides, we already disable AS anyway so what's the big deal

The deal is that end users who want to use AS as their start/new/home page, can do so, and have a clean interface that doesn't "compromise" them in some way. It's not always about us. Sure, some it could be seen as a fallback/future-proof, but for others, it is the only "protection"

> why are you testing in an early beta

I just have the one portable test FF, that's a dev version (if I need it for current I can close and rename my profile folder). And I might as well "test" in the next version to save time :) IDK, maybe I'm an idiot. I can already see the default values in stable, so I'm not worried about them.

> Have you actually clicked on one of these things? all it does is put @amazon in the urlbar for easy searching

Well, nope. Didn't know that. Does it break with keywords not working in default js? I guess they're harmless then. I want to test something. I think it only shows those if you have an existing search engine with the same name, in which case, then they are truly harmless. BRB

Thorin-Oakenpants:
OK, thought as much. I didn't even have an amazon search engine. I also removed the google one. Here's what happens when I use it
- `https://www.google.com/search?client=firefox-b-ab&q=test`

It's sending metadata (not that we hide the browser anyway in the header), so I guess it's harmless

PR #624, opened 2019-01-30
FF65 removals [DRAFT]

earthlng:
do we need to keep the info for "autocomplete" under 9999?

not sure what to do with app.update.auto, maybe kind of a tag behind the pref, like `// FF65+: non-Windows only` or a line saying something like "In FF65+ on Windows this setting is now stored in a file and the pref was removed"

### DIFF
--- a/user.js
+++ b/user.js
@@ -361,8 +361,6 @@ user_pref("browser.safebrowsing.provider.google4.dataSharingURL", "");
  * [1] https://bugzilla.mozilla.org/buglist.cgi?bug_id=1170190,1141814 ***/
    // user_pref("privacy.trackingprotection.annotate_channels", false);
    // user_pref("privacy.trackingprotection.lower_network_priority", false);
-/* 0426: enforce Content Blocking (required to block cookies) [FF63+] ***/
-user_pref("browser.contentblocking.enabled", true); // [DEFAULT: true]
 
 /*** [SECTION 0500]: SYSTEM ADD-ONS / EXPERIMENTS
      System Add-ons are a method for shipping extensions, considered to be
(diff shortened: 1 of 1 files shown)


bogachenko:
https://github.com/ghacksuserjs/ghacks-user.js/blob/0d87fa3bc2903963645914207dbc03eeec1b0ad9/user.js#L183 make a note "for everyone except Windows" (or delete it altogether)

Thorin-Oakenpants:
> do we need to keep the info for "autocomplete" under 9999?

I'd hate to lose the info until its totally obsolete. It still applies to ESR (although esr users can just use v60, but then they would miss out on our improvements that aren't new prefs). I think we can leave it in until we hit ESR68, but I'll edit it a tad to point back to the non-deprecated part

Thorin-Oakenpants:
> not sure what to do with app.update.auto

Is this actually used by other OSes? I thought Linux (IANAExpert) used their own updaters etc. NFI about Macs. And Android for sure would use the playstore/google settings.

Anyway, assuming it is staying active, then a one-liner note AND a [NON-WINDOWS FF65+] tag (don;t start the tag with `[FF` or :bird: 's script will think that was when it was added

Thorin-Oakenpants:
I'm happy if you are. Although I wouldn't mind knowing how to stop this (you can't allow portables to update via the FF mechanism) once I'm 65. Depending on the info, maybe we could add that in? What's the bugzilla?

earthlng:
> you can't allow portables to update via the FF mechanism

why not? I do it all the time

As for how to stop it, the UI is still there. Bugzilla: probably https://bugzilla.mozilla.org/1458308

earthlng:
If you want to add more info, feel free to do that and then you can merge it.

Thorin-Oakenpants:
> why not? I do it all the time

I haven't for a while now (over a year), because AFAIKnew it will only update the 64bit (portable comes with both - i actually deleted the 32 bit part once and the launcher threw a hissy fit), and then the launcher fails. Maybe I'll try it again, and it was just a one off bad experience.

> As for how to stop it, the UI is still there

Oh, the UI still applies. I didn't realize that.

earthlng:
> i actually deleted the 32 bit part once and the launcher threw a hissy fit

you just need to keep the firefox.exe, everything else in the 32bit folder can be deleted AFAIK

Thorin-Oakenpants:
> you just need to keep the firefox.exe, everything else in the 32bit folder can be deleted AFAIK

cool, that worked. Not that I was too worried about it (takes 10 seconds to copy/pasta for local backups and seconds to sync changes only to external backups etc), but the size of FF portable has been creeping up (seriously, what keeps bloating this this, I'm not heavy at all on extensions). I do get that uBO and uM hosts/filters can add size, but i'm talking about the core .. and startup caches etc ..

FF64 has been floating just over the 400mb mark .. 405, 411, 410 .. every backup. 63 was around 10mb less ... 62 was less than that. I don't store any persistent data (history, forms, cookies except a half dozen, no IDB, etc).

Anyway .. that worked. Shaved off 156mb

PR #625, opened 2019-01-30
Update user.js

bogachenko:
https://github.com/ghacksuserjs/ghacks-user.js/issues/610#issuecomment-458931521

### DIFF
--- a/user.js
+++ b/user.js
@@ -473,6 +473,7 @@ user_pref("network.dns.disableIPv6", true);
 user_pref("network.http.spdy.enabled", false);
 user_pref("network.http.spdy.enabled.deps", false);
 user_pref("network.http.spdy.enabled.http2", false);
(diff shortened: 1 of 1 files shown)


bogachenko:
```
// Network Connectivity Services -- unneeded entries unless you use Sync account and maybe Push notifcations
// These FF 65 entries go against the security principle of a reduced attack surface, as there's a DNS Hijack opportunity.
pref("network.connectivity-service.enabled", false);
pref("network.connectivity-service.DNSv4.domain", "");
pref("network.connectivity-service.DNSv6.domain", "");
pref("network.connectivity-service.IPv4.url", "http://0.0.0.0");
pref("network.connectivity-service.IPv6.url", "http://0.0.0.0");

```
where to insert

bogachenko:
I could fix it. if you said what and where .... but ok.

Issue #605, opened 2018-12-18
Analysis: overrides+tags+troubleshooter


Thorin-Oakenpants:
ToDo:
 - [x] flip some things to inactive (relax things a little)
 - [x] remove dead wood if any
 - [x] make sure active items get a [setup] (web/chrome/perf) tag (if required)
 - [x] make sure inactive items use [warning] not [setup] (web/chrome/perf)
    * 4502 is the only exception, as RFP is active, so in effect window sizing is already happening
 - [ ] revisit inactive items (the whole js) to see what we could add [setup-harden] to
 - [ ] create relaxed/hardened sticky for easy overrides cut & paste
 - [ ] add sticky link in readme
 - [ ] add sticky link in wiki implementation
 - [ ] close this issue, relax, drink copious amounts of beer, in the nude

---

### :small_blue_diamond:  Hi guys

First of all, I know.. I know... you're all excited.... but just relax ... I appreciate feedback etc, but please let "us" analyze it first and trim the list down (there's some junk in there) and actually create an end product. We already have the info, so please no suggestions - that can come later when we refine.

### :small_blue_diamond: What are we trying to do?

 Well, based on what's getting changed in overrides, what we already have tags on (we put em there for a reason), what's already in the troubleshooter (TS for short), and what's important (just because someone changed it doesn't mean its worthy):
- cleanup the list
- refine the troubleshooter based on tags/overrides
- refine tags based on ts/overrides
- create draft relaxed & hardened lists - a relaxed list should then practically mirror the setup-tags, and cover the vast bulk of what end users change

### :small_blue_diamond: Notes

Green (relaxed) vs red (hardened) isn't necessarily that but generally fits that pattern (we have active prefs that can actually hardened further). The green is actually active prefs, red, inactive. In general, active prefs should be where almost all our `setup` tags go (to relax), and `warning` and `harden` tags would primarily be on inactive prefs (to harden or warn, doh!). The TS info is a hybrid, as earthlng is trying to cover any end-user's config.

PS: when we create a hardened list you can see immediately that items stand out (from the red lot)

```
1. disable history
2. disable password saving
3. make permissions memory only
4. make certs memory only
5. disable svg
6. block websites setting a context menu?
7. blank restricted domains
8. disable memory cache (not in the results, but I'd add this)
```

The same process is just as easy for relaxed/setup items, just more of them which we can breakup into perf, web & chrome - and we've already done most of it.

---

### :small_blue_diamond: The Data

[...]

PR #626, opened 2019-01-30
location bar changes

Thorin-Oakenpants:
@earthlng . now that location bar autocomplete is gone, users will have to resort to css to stop the dropdown. So I see no point anymore in disabling all the suggestion types - because otherwise all you get is a single line blue dropdown.

These three items, 850a, are also on the list in #605 as the ones most often changed - small sample, but seems to be a pattern there. Note: I'm not concerned about any tags right now, the 605 issue can still address that later down the track

This should also solves search engine keywords, right?

What other items should we make inactive here? Which one is the one that shows your search engines as icons on the bottom of the dropdown - that can be make inactive IMO

### DIFF
--- a/user.js
+++ b/user.js
@@ -585,7 +585,7 @@ user_pref("browser.urlbar.speculativeConnect.enabled", false);
 user_pref("browser.urlbar.autoFill", false);
 /* 0850e: disable location bar one-off searches [FF51+]
  * [1] https://www.ghacks.net/2016/08/09/firefox-one-off-searches-address-bar/ ***/
-user_pref("browser.urlbar.oneOffSearches", false);
+   // user_pref("browser.urlbar.oneOffSearches", false);
 /* 0850f: disable location bar suggesting local search history [FF57+]
  * [1] https://bugzilla.mozilla.org/1181644 ***/
 user_pref("browser.urlbar.maxHistoricalSearchSuggestions", 0);
--- a/user.js
+++ b/user.js
@@ -516,10 +516,10 @@ user_pref("network.file.disable_unc_paths", true); // [HIDDEN PREF]
 user_pref("network.gio.supported-protocols", ""); // [HIDDEN PREF]
 
 /*** [SECTION 0800]: LOCATION BAR / SEARCH BAR / SUGGESTIONS / HISTORY / FORMS [SETUP-CHROME]
-     If you are in a private environment (no unwanted eyeballs) and your device is private
-     (restricted access), and the device is secure when unattended (locked, encrypted, forensic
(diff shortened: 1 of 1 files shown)


Thorin-Oakenpants:
OK .. the search icons on the bottom row is `browser.urlbar.oneOffSearches` - I'll do that one as well

earthlng:
next we'll build that ~wall~ ~fence~ fence-wall :)

Thorin-Oakenpants:
what about the other prefs in 850's?

crssi:
850f
``` js
user_pref("browser.urlbar.maxHistoricalSearchSuggestions", 0);
```

Is anyway default `0`, at least on FF 65.

As of 850d, I like it as is now, but if you will comment out 850d and 850f, then all 850 can go under 5000?

Thorin-Oakenpants:
same on 64. I'll check back to stable 60

Thorin-Oakenpants:
> but if you will comment out 850d and 850f, then all 850 can go under 5000?

I've commented them out in a new commit. But no, they can't go under 5000's because they have privacy implications

Thorin-Oakenpants:
`browser.urlbar.maxHistoricalSearchSuggestions` is default false in FF stable 60-64 and 66. I'm not on 65 yet but I'll just take a wild stab in the light and say it too is false.

If ESR60 is also false, then I suggest we remove this from the user.js. Spose I better go get a portable ESR. BRB

crssi:
On 65 is zero `0` and not `false.`

Thorin-Oakenpants:
^^ sorry .. I meant 0

on ESR 60.1 (from july 2018) and in ESR 60.5 (released 20 hrs ago) .. the value is also 1

So I'm going to remove it

bogachenko:
> now that location bar autocomplete is gone, users will have to resort to css

just for reference. can someone come in handy
```css
#PopupAutoCompleteRichResult,
{
    display: none!important
}
```

Thorin-Oakenpants:
^^ https://github.com/ghacksuserjs/ghacks-user.js/commit/6147fed61c4f6f0a13f38947598ee20343c2531f

crssi:
Should also the following be inactive and user to decide to force with overrides or not?
```
browser.startup.homepage
browser.newtabpage.enabled
```

earthlng:
@crssi we blank the homepage so that hopefully no sites load before extensions are loaded and ready to go.
And Activity Stream is probably something that most of our users don't really like and that's why we disable it by default and those who want it can override instead of the other way around.

But everything is up for discussion and you can, if you want, open an issue and do a poll on whether more users want AS enabled by default or not, or something like that.

PR #627, opened 2019-01-31
clean up "Firefox Data Collection & Use"

Thorin-Oakenpants:
- telemetry prefs together in 330's
- Firefox Data Collection & Use prefs to 340's (but leave crash reports in 350s)
- move `app.shield.optoutstudies.enabled` to 330's - this is an internal pref which controls if you get the system addon
- make notes that `datareporting.healthreport.uploadEnabled` controls studies and ext recommendations


Thorin-Oakenpants:
We could probably combine the crash reports into one "disable Crash Reports"

Thorin-Oakenpants:
the only thing I'm not happy about is that the combined crash reports doesn't make it clear which pref (autosubmit2) applies to the UI. So I'd actually like to split the other three off into a new 351, or we could add a comment at the end of the pref?

earthlng:
> So I'd actually like to split the other three off into a new 351, or we could add a comment at the end of the pref?

either 1 is fine with me but I prefer the former. The rest LGTM

PR #631, opened 2019-02-04
0707: clean up DoH

Thorin-Oakenpants:
- remove network.trr.bootstrapAddress = default blank anyway
- remove network.trr.uri - this is mozilla's default value. Unfortunately, that value is hard-coded in FF65+ into the UI settings as an option to use the default URI, so changing this could be misleading
- In FF65 there is now a custom URI field in the UI settings, and that's what users should be using - but it actually controls two prefs if you check to use custom - see https://github.com/ghacksuserjs/ghacks-user.js/issues/610#issuecomment-460169782

I am not interested in maintaining how this UI works (and it could change). The single pref left above is all that is needed to turns it on or off, that's all I care about. I've added the [SETTING] info and that's where people can do their shit. That's good enough for me

### DIFF
--- a/user.js
+++ b/user.js
@@ -501,7 +501,7 @@ user_pref("network.proxy.socks_remote_dns", true);
  * CVE-2017-5384: Information disclosure via Proxy Auto-Config (PAC)
  * [1] https://bugzilla.mozilla.org/1255474 ***/
 user_pref("network.proxy.autoconfig_url.include_path", false); // [DEFAULT: false]
-/* 0707: disable (or setup) DNS-over-HTTPS (DoH) [FF60+]
+/* 0707: disable DNS-over-HTTPS (DoH) [FF60+]
  * TRR = Trusted Recursive Resolver
  * .mode: 0=off, 1=race, 2=TRR first, 3=TRR only, 4=race for stats but always use native result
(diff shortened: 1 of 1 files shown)


earthlng:
I don't like this change because the UI options are very limited and using the prefs directly allows better control. f.e. in 66 it changes the value to 2 (TRR first) instead of 3 (TRR only) and the UI has no option to set the bootstrap IP address.

earthlng:
> Given that all we wanted to do was disable it

IMO that's not all we wanted to do. I mean we titled it "disable **(or setup)** DoH", added a link explaining what DoH is and even a "warning" about Cloudflare. It's there to provide information and has all the details about available options for those who want to use it (or force-disable it).
The UI on the other hand will always be a dumbed down version with very limited options compared to whats actually available.

Thorin-Oakenpants:
Just adding the new pref does not do enough if you want to expand on how to set it up - as already outlined, it can lead to incorrect info displayed in the UI.

Do an edit to this commit then, add the new pref (and put the other two back in), and build in the info that just changing the custom uri is not enough, etc, and that you also need to change the other pref as well. Messy as fuck. All I wanted to do was disable it - but if you want to change it, do an edit

Issue #648, opened 2019-02-13
0307: lightweightThemes.update.enabled


Gitoffthelawn:
Section 0307 reads:

```
/* 0307: disable auto updating of personas (themes) ***/
user_pref("lightweightThemes.update.enabled", false);
```

Is the description accurate?  Does the pref really disable auto-updating of personas (themes)?  I think if the pref is `true`, but `Update addons automatically` is not enabled in `about:addons`, then the personas will not be auto-updated.

Thorin-Oakenpants:
Looks like a rubbish pref to me. I don't even know if it's a thing any more, since Quantum. And themes are not an issue AFAIK (unlike extensions which can go rogue). If you look at themes vs addons in options, you will see they actually use the same cog wheel and menu for controlling them.

So yeah, I think we can remove this pref altogether - after the Big E scopes it out

Thorin-Oakenpants:
when the big E says its OK to remove it, if he doesn't, I will ..so just waiting on @earthlng

Thorin-Oakenpants:
lightweightThemes are, I guess, what they are using for themes? e.g I use the inbuilt dark theme, and I use compact .. and I have this
- `lightweightThemes.selectedThemeID` - `firefox-compact-dark@mozilla.org`

But if we set extensions to not auto-update, then we should also set themes to not update to be consistent. I will install a theme on a test profile, an older version of a theme and see what happens when I check for updates

Thorin-Oakenpants:
> If you look at themes vs addons in options, you will see they actually use the same cog wheel and menu for controlling them

Exactly. I set all updates to manual. Installed an older version of a theme. When I went to about:addons > Extensions> cog wheel > check for updates ... it dutifully found an update for the theme.

So it really begs the question of do we want to keep this or ditch it because it has no privacy etc issues. And if we keep it, maybe make it inactive - but then what purpose does it serve? None, so I think we should ditch it - **and update* the extension pref to include the word themes

Thorin-Oakenpants:
the pref doesn't seem to affect any updating
![themes](https://user-images.githubusercontent.com/16656956/52707946-d35e4200-2f80-11e9-9041-57b7bb72becf.png)

Thorin-Oakenpants:
I wonder if search update is the same

earthlng:
[...]

PR #649, opened 2019-02-13
remove 0307: themes #648

Thorin-Oakenpants:
since 57(?) themes and web extensions are controlled by the same mechanism (301b+302b)

### DIFF
--- a/scratchpad-scripts/ghacks-clear-[removed].js
+++ b/scratchpad-scripts/ghacks-clear-[removed].js
@@ -1,7 +1,7 @@
 /***
  This will reset the preferences that have been removed completely from the ghacks user.js.
 
- Last updated: 13-Feb-2019
+ Last updated: 30-Jan-2019
 
  For instructions see:
  https://github.com/ghacksuserjs/ghacks-user.js/wiki/3.1-Resetting-Inactive-Prefs-[Scripts]
@@ -120,8 +120,6 @@
     /* 65-beta */
     'browser.contentblocking.enabled',
     'browser.urlbar.maxHistoricalSearchSuggestions',
-    /* 66-beta */
-    'lightweightThemes.update.enabled',
     /* reset parrot: check your open about:config after running the script */
     '_user.js.parrot'
   ]
--- a/scratchpad-scripts/ghacks-clear-[removed].js
+++ b/scratchpad-scripts/ghacks-clear-[removed].js
@@ -1,7 +1,7 @@
 /***
  This will reset the preferences that have been removed completely from the ghacks user.js.
 
- Last updated: 30-Jan-2019
+ Last updated: 13-Feb-2019
 
  For instructions see:
  https://github.com/ghacksuserjs/ghacks-user.js/wiki/3.1-Resetting-Inactive-Prefs-[Scripts]
@@ -120,6 +120,8 @@
     /* 65-beta */
     'browser.contentblocking.enabled',
     'browser.urlbar.maxHistoricalSearchSuggestions',
+    /* 66-beta */
+    'lightweightThemes.update.enabled',
     /* reset parrot: check your open about:config after running the script */
     '_user.js.parrot'
   ]
--- a/user.js
+++ b/user.js
@@ -179,13 +179,13 @@ user_pref("geo.provider.use_gpsd", false); // [LINUX]
      monetized extensions, time constraints, legacy issues, and fear of breakage/bugs.
      It is still important to do updates for security reasons, please do so manually. ***/
 user_pref("_user.js.parrot", "0300 syntax error: the parrot's not pinin' for the fjords!");
-/* 0301b: disable auto-update checks for extensions and themes ***/
+/* 0301b: disable auto-CHECKING for extension and theme updates ***/
    // user_pref("extensions.update.enabled", false);
-/* 0302a: disable auto update installing for Firefox [NON-WINDOWS FF65+]
+/* 0302a: disable auto-INSTALLING Firefox updates [NON-WINDOWS FF65+]
  * [NOTE] In FF65+ on Windows this SETTING (below) is now stored in a file and the pref was removed
  * [SETTING] General>Firefox Updates>Check for updates but let you choose... ***/
 user_pref("app.update.auto", false);
-/* 0302b: disable auto update installing for extensions and themes (after the check in 0301b)
+/* 0302b: disable auto-INSTALLING extension and theme updates (after the check in 0301b)
  * [SETTING] about:addons>Extensions>[cog-wheel-icon]>Update Add-ons Automatically (toggle) ***/
 user_pref("extensions.update.autoUpdateDefault", false);
 /* 0303: disable background update service [WINDOWS]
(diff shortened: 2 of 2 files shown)


earthlng:
It's a bit weird because when both 301b+302b are set to false then the UI setting `Update Add-ons Automatically` is unchecked (as expected). When you toggle the UI to enable, it sets both to true but when you uncheck it again, it only sets 0302b to false.
There doesn't seem to be a way to set 0301b to false via the UI other than about:config.

Some people might want to have extension updates enabled but themes updates disabled and I don't like to remove prefs that still do something and which to some extend keep FF a bit quieter.

Thorin-Oakenpants:
> It's a bit weird because

Yeah I just commented in the issue about that. The [setting] for 0301b should go

Thorin-Oakenpants:
> Some people might want to have extension updates enabled but themes updates disabled and I don't like to remove prefs that still do something and which to some extend keep FF a bit quieter.

The permutations are doing my head in a little. Need coffee. I tested the update with this pref at false, and it did not stop me **manually** checking (doh!) or updating.

The question is, if extensions update master switch is allowed, but themes isn't, what happens? I can try to test for it, but not if it takes a day

earthlng:
> if extensions update master switch is allowed, but themes isn't, what happens?

it'll look for extension updates but not theme updates

Thorin-Oakenpants:
^^ https://github.com/ghacksuserjs/ghacks-user.js/issues/648#issuecomment-463187686

earthlng:
`extensions.update.enabled` is the master switch for the extension and theme update checks (+install in some cases) during the daily background update check. You can't disable the daily background checks entirely but you can make it not really do anything (other than update the timestamps).

if `extensions.update.enabled` is true, it will first do the LWT checks which are controlled by the pref you want to remove here. There doesn't seem to be a pref to just disable the auto-installing of LWTs, meaning that pref controls both LWT checking as well as installing. Or in other words LWT updates don't care about `extensions.update.autoUpdateDefault` - that's apparently just for extensions. (unless I'm missing something in the code)
Next it will go on to check extensions (and install them automatically depending on .autoUpdateDefault and/or each extensions individual auto-install setting)

And lastly the daily bg-check will also look for system addon updates but that's controlled in its own way and is besides the point here. The SA part is NOT affected by `extensions.update.enabled`!

Thorin-Oakenpants:
So there must be a difference between a manual check and the one run internally by Firefox - I would have thought the code (spaghettified over the years?) would have been "reused".

I'm going to test this then: steps:
- `extensions.update.enabled` = true
- `extensions.update.autoUpdateDefault` = true (so I know the background update fired)
- `lightweightThemes.update.enabled` = false
- install 1 ext, 1 theme, both outdated versions
- open the test profile in 24+ hours

I'll get back to you. Regardless of the outcome (not doubting your code-reading skills), the pref will have to be modified (e.g only used in background checks), and most of the other changes still need to stand in some way or another

I actual decided to install a theme - it's almost identical to dark theme, but gets rid of the transparency in the title bar when the app loses focus. I had trouble with that in userChrome. And of course I have a coupel of other tweaks.

Thanks for looking into it

Thorin-Oakenpants:
^^ and extensions.update.interval = 3600 (1hr, instead of 24 hrs)

Thorin-Oakenpants:
OK, that sucked. There's an extra step .. set extensions.update.enabled = false before adding the theme, as it seems to update the theme as you install it - and then set it to true afterwards. It doesn't do this with the extension. I'm not entirely sure, because I could have sworn it was the older theme added, but the version numbers are so similar I may have made a mistake. Removing, restarting, and adding the old version theme when extensions.update.enabled = true simply gave me the latest up to date version.

When I looked at my test (timer is set for an hour), the theme was up to date but the extension wasn't (which lead me to the above paragraph), and then the extension updated while I was in there. Might set the timer for 10minutes and take a break.

Thorin-Oakenpants:
I can confirm that the theme **IS UPDATED** by the daily background checks

If you want to confirm it: STR

-  `extensions.update.enabled` = false :exclamation: important as when true the theme installed would always be the latest one
-  `extensions.update.autoUpdateDefault` = true - so I know the background update fired
-  `lightweightThemes.update.enabled` = false
-  `extensions.update.interval` = 300 (5 minutes)
-  install 1 ext, 1 theme, both outdated versions
-  check versions installed are the old ones
-  `extensions.update.enabled` = true :exclamation: important to set this back for the test
- recheck versions are the old ones
- close
- open the test profile in 5+ minutes and give it a little time to fire. Took about 30 seconds and I was just waiting and saw the chrome "flicker", then checked the theme version, it was on the latest

I suspect the extension updates just include everything (and probably include search updates? which we could also check out)

Thorin-Oakenpants:
@earthlng bump

earthlng:
I think the thing is that most themes aren't really Lightweight Themes anymore. fe https://bugzilla.mozilla.org/1508777 mentions ...

> When AMO begins converting lightweight themes to packaged themes, users with recommended themes will end up with two copies of every converted lightweight theme.

They're apparently also in the process of removing LWTs completely, see [1469930](https://bugzilla.mozilla.org/show_bug.cgi?id=1469930) and [1525511](https://bugzilla.mozilla.org/show_bug.cgi?id=1525511) (opened 10 days ago). (and [Convert built-in LWTs to static themes](https://bugzilla.mozilla.org/show_bug.cgi?id=1525762) which was opened 10 days ago as well.)
I'm not sure but the only "lightweight" themes are seemingly just the 5 which are shipped with Firefox and only 3 of those are shown in the customize UI.

So yeah, the pref is probably mostly dead at this point but I'd still prefer to wait with removing it until they remove LWTs from Firefox entirely, including all their prefs, which looks like it should happen pretty soon.

Thorin-Oakenpants:
- https://blog.mozilla.org/addons/2009/10/21/personas-now-on-amo/ - weeeeee .. personas back in 3.6, ahh the good old days
- https://blog.mozilla.org/addons/2018/07/12/upcoming-changes-for-themes/
- https://blog.mozilla.org/addons/2018/09/20/future-themes-here/

I'll put it back. I can understand someone may be using a persona they modified and don;t want it to update, but they do want extensions auto-updated. Pretty rare, but still. I guess we can wait for the pref to get removed .. in like 5 or 6 releases time

Thorin-Oakenpants:
I still think it needs MOAR info that themes in 0301+0302 are not the same as 0307

Thorin-Oakenpants:
bump

```js
/* 0307: disable auto updating of personas (lightweight themes / LWT)
 * LWT: not to be confused with those in 0301* + 0302*, which use the FF55+ Theme API
 * Mozilla plan to convert existing LWTs and remove persona support in the future, see [1]
 * [1] https://blog.mozilla.org/addons/2018/09/20/future-themes-here/  ***/
user_pref("lightweightThemes.update.enabled", false);
```

I think something like this removes all confusion. When the pref is deprecated we can trim that down or even just remove it from the js.

earthlng:
looks great but why mention "personas" at all?
how about this?
```js
/* 0307: disable auto updating of lightweight themes (LWT)
 * LWT: not to be confused with themes mentioned in 0301* + 0302*, which use the FF55+ Theme API
 * Mozilla plan to convert existing LWTs and remove LWT support in the future, see [1]
 * [1] https://blog.mozilla.org/addons/2018/09/20/future-themes-here/  ***/
user_pref("lightweightThemes.update.enabled", false);
```

Issue #692, opened 2019-04-15
discussion: make a bunch of items inactive etc


Thorin-Oakenpants:
from #605 (and from comments I have read elsewhere, which has made me step back a little, although to be honest I was always aware of this - it's just more of a kick in the pants to get something done about it, and to move the relaxed sticky on)

Here's the thing. We don't need to go all out super hard (and we don't: a lot of items we could but don't). The threat model just isn't there for most of us, if not all of us. And the prefs would still be there for those who want them. I just think we could come up with a better default

Here's a pic - as you can see, since I made that list back in Dec last year, we've already relaxed a number of items "naturally" (i.e not because of this list)
![update](https://user-images.githubusercontent.com/16656956/56168888-039b7e80-5fcc-11e9-963e-0ccc165d2e67.png)

You can read [this comment](https://github.com/ghacksuserjs/ghacks-user.js/issues/605#issuecomment-483433499) if you want and the one above it, but essentially some of these are the **threat model**, and some are a **law of diminishing returns** (if that's the right law - I am not a lawyer!!), and some are just **extra hoops** that make no sense AFAICT

For example: since we've already relaxed most of the shoulder surfing stuff, why should be still block disabling saving to desktop. Why should we stop users "opening with" - it's not a risk as it requires the user to make a choice, etc (AFAIK), i.e exe's won't execute on their own, and if they want to open file types with their own program, FFS, let them do it. This sort of hardening is just going a tad too far IMO. 

I know the Big E will scream, but hear me out. ASMJS and WASM: sure, one is new'ish, and both can be security issues - but in all seriousness, the chances of getting hit by anything from these are practically zero. The risk is so low it's ridiculous. It might be years before an exploit is found and even then, not in the wild. It's not like we're not careful internet users, but even a non savvy user would struggle to ever to get bitten. That's my take on it. If it's good enough for major browsers to enable, it's good enough for the DEFAULT template.

GMP + openh264: do we really need to block these. Aren't they just extra things to add to overrides in order to get OTHER things working: i.e in and of themselves, they are not a threat?

Cache: the chances that sites use cache to track are probably extremely slim. I'm talking about embedding UUIDs in the cached item or something. Timing attacks I'm not sure about. I think the threat model here doesn't exist for us. ETAGs are not the issue: you have to disable memory cache as well, and we don't do that.

[...]

PR #697, opened 2019-04-21
start harden->inactives #692

Thorin-Oakenpants:
Read the issue


Thorin-Oakenpants:
summary so far. I have left the one switch char on, so it's easier to read for now

**section 6000**
- http2 & altsvc
- document fonts
- asmjs & wasm
- ION & baseline JIT
   * for some reason I keep thinking these are tied to asmjs (we have them inactive)
   * if they aren't then I will move them back
- HWA
- mathml & disable open with
   * I think we could maybe just move these back but leave inactive, but if we do this too much then we might start needing setup-harden tags

**made inactive** (not in 6000, no harden tag)
- downloads.folderList
   * we can either remove it, or leave it inactive, or shove it in personal. Personally I use this, otherwise every session I have to relocate my non-standard dir on first download
- extension scopes

There is some major'ish rewriting of the tags etc in 6000 items. I honestly don't know how you will like these changes (so far: I want to add some more, but not overboard: video stuff, more font stuff), but there's a bigger picture here. Totally open to discussion and finding the right mix.

Issue #710, opened 2019-04-28
SB+TP etc: extreme pruning


Thorin-Oakenpants:
- [x] remove inactive dead wood 0400s
- [x] cleanup setup tag on real-time binary checks
- [x] adjust wiki
- [x] check (and remove reporting 0415 if ok to remove)
- [x] remove URL for data sharing (not needed)

---
original post

suggestions
- `0403`: kinto: never been used
- `0410` + `0411` + `0412`: safe browsing: There's a UI for these: totally not needed, also all inactive
- `0413`: safe browsing urls: these aren't even needed due to master prefs, also inactive
- `0420` + `0422` + `0423` + `0424` + `0425`: tracking protection, the entire lot, there's a UI for all that, also all inactive
- also any deprecated prefs that relate to these (edit: only those still in the user.js after I archived pre FF61 deprecated items)

I am not interested in providing info and maintaining it for other blogs etc to use while screaming about Mozilla spying on it's users and being in cahoots with google etc, and then pointing back to here as a source. Every second comment about us (reddit, hackernews, etc) assumes we're also anti-mozilla, and that we put users at risk.

- `0415`: reporting urls: active, but these require the end user reporting them (help> report..., just need to double check that)

This will trim at least 64 lines (more when I clean up the header), and remove any maintenance moving forward, and make me a happy camper. Section 400 then becomes nice and small and compact, and the wiki warning about binaries becomes much clearer.

**If you disagree then you better speak up, and fast. Because I'm just going to pull the trigger**

claustromaniac:
I don't know what comments you've been reading, but if they point to a bunch of inactive prefs as source to say we're anti-mozilla and such shit, I wouldn't care about their opinions. Correction: I wouldn't care either way.

Thorin-Oakenpants:
Did you like my massive pruning of old deprecated shit ... the user.js feels so light and fluffy now

Thorin-Oakenpants:
> I don't know what comments you've been reading

"I read and I know things" - Tyron Lannister

I read a lot, really . seriously .. a fucking lot. I'm over-exaggerating, but for example, on reddit they actively do not like this repo (that's the vibe and I have seen posts linking to us as a useful source with disclaimers to be careful, getting removed, not just deleted, but removed so there is no trace). Any mention of us gets almost zero votes, and yet shit things like privacytools.io little list gets tons of mentions and upvotes.

[...]

PR #711, opened 2019-04-28
Part 1: inactive dead wood, see #710

Thorin-Oakenpants:


Thorin-Oakenpants:
That's over 100 lines of crap removed and future maintenance forgone. I can also remove the section header and just list the required reading, if you want.

We can deal with reporting in another commit, since they are active, and I want to double check they actually require user interaction

Thorin-Oakenpants:
godamnit, I hate conflicts. Why is it showing stlll showing 474 lines removed .. I don't trust it. closing and will re-do

PR #712, opened 2019-04-28
inactive dead wood, see #710

Thorin-Oakenpants:
Lets try this again.

### DIFF
--- a/user.js
+++ b/user.js
@@ -44,7 +44,7 @@
      0100: STARTUP
      0200: GEOLOCATION
      0300: QUIET FOX
-     0400: BLOCKLISTS / SAFE BROWSING / TRACKING PROTECTION
+     0400: BLOCKLISTS / SAFE BROWSING
      0500: SYSTEM ADD-ONS / EXPERIMENTS
      0600: BLOCK IMPLICIT OUTBOUND
      0700: HTTP* / TCP/IP / DNS / PROXY / SOCKS etc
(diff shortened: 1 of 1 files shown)


user.js
```diff
@@ -275,111 +275,44 @@ user_pref("network.captive-portal-service.enabled", false); // [FF52+]
  * [1] https://bugzilla.mozilla.org/1460537 ***/
 user_pref("network.connectivity-service.enabled", false);
 
-/*** [SECTION 0400]: BLOCKLISTS / SAFE BROWSING / TRACKING PROTECTION
-     This section has security & tracking protection implications vs privacy concerns vs effectiveness
-     vs 3rd party 'censorship'. We DO NOT advocate no protection. If you disable Tracking Protection (TP)
-     and/or Safe Browsing (SB), then SECTION 0400 REQUIRES YOU HAVE uBLOCK ORIGIN INSTALLED.
```
claustromaniac: Are you perhaps removing this part just because some people apparently don't read it, or is there any other reason? I kinda like that statement.

Thorin-Oakenpants: The TP section is entirely removed, so 3rd party, control of lists, uBO etc are not relevant.

claustromaniac: > We DO NOT advocate no protection.

>  3rd party, control of lists, uBO etc are not relevant.

You're not removing SB though. That was my observation.

claustromaniac: Actually, scratch that. Sorry, I'm doing like 4 things at the same time.

Thorin-Oakenpants: Sure. But I think it's pretty obvious that we're saying there is nothing wrong with SB (except the bit we provide about binaries)

If you want to craft up something more eloquent, go ahead :)

Thorin-Oakenpants: We'll just keep patching this PR until everyone is happy (or a 2 to 1 majority)

Thorin-Oakenpants: only 4 things?

claustromaniac: `4 things != 4 tabs`, you know. There's a physical world out there too, or so they say.

I'm fine with it as it is, no need to reword anything IMO.

Thorin-Oakenpants: things != tabs. I didn't say tabs .. why did you assume I always live in a browser .. I also drink and know things. Seeing as you're so busy (as noted by your status), I'll refrain from asking you to help me get my DTD locale leak PoC working

PR #742, opened 2019-06-10
updater.bat -ESR option

### LINKED ISSUE #741: Any way to enable ESR user settings with the updater.bat script ?
I'm currently using the auto user.js updater.bat. With the Linux script, the -e option enable the ESR settings but there's no option for Windows, I tried a few different way by playing with the script code but I can't achieve anything, the only solution for now is to put every entry 1 by 1 in the user.js-overrides but if any entries is changed or added I'll miss it. So, how can I auto enable those entries with the script ?

Thanks

claustromaniac:
Closes #741

### Summary:
Adds an `-ESR` parameter that, after the `user.js` is downloaded (as `user.js.new`), replaces this line:
```js
/* ESR60.x still uses all the following prefs
```
with this:
```js
// ESR60.x still uses all the following prefs
```
That's all.

### DIFF
--- a/updater.bat
+++ b/updater.bat
@@ -3,10 +3,10 @@ TITLE ghacks user.js updater
 
 REM ## ghacks-user.js updater for Windows
 REM ## author: @claustromaniac
-REM ## version: 4.6
+REM ## version: 4.8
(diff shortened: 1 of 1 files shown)


earthlng:
Feel free to commit and if you don't mind please also update the wiki. Thanks Catman

claustromaniac:
All done. Thanks, Earthman.

PR #745, opened 2019-06-11
add -RFPalts option

claustromaniac:
Déjà vu. (#741)


earthlng:
I also updated the version and the wiki. Thanks again, Catman :) 💋

Issue #753, opened 2019-07-01
Linted updater.sh with Shellcheck


tyler-public:
Hi,

I've run `updater.sh` through shellcheck and fixed the problems it reported, I'm working on making the `updater.sh` run with `/bin/sh` and be POSIX compliant, but that's going to take more time.

https://gist.github.com/iamtpage/1936f4373c2f5a2bc01d95ac58e8d426

Thank you so much for the work on this project, hopefully this will be of use

claustromaniac:
At a first glance it looks good to me. Would you like to submit a pull request? If not, I can submit one myself. Still, when we can, we should at least test this first to make sure it works well for everyone. @earthlng's outdated VM is (or used to be?) especially fussy.

I'm summoning @overdodactyl in advance. (🐦, can you still flap those wings?)

iamtpage:
I'd like to have someone else test this before talking about merging it in. I've tested all functionality except for the updating part (since updating would overwrite the linted shell script with the original)

earthlng:
Hi!

thanks for contributing here, mate!

I've never heard of shellcheck, which tbh doesn't mean much because I haven't worked with a lot of linux/mac scripting for while, but is that some kind of defacto standard in shell script linting?
I mean, is this just about "style" or did the script not actually work for you without these changes?

> Would you like to submit a pull request? If not, I can submit one myself.

Yes, someone please submit a pull request so that we can work on it there. I've already spotted a mistake in our current script thanks to this but also something that I believe is not quite right. A PR makes it a lot easier to comment on and change each line where necessary. Thanks

iamtpage:
For shell scripts, shellcheck tends to be the "defacto standard" for linting and syntax checking. It is a bit of a mix of "best coding styles" and best practices for preventing things like undefined behavior, common mistakes, etc.

I'll open a PR for the BASH version, then once it's fixed I'll open another for the POSIX compliant one

earthlng:
Thanks for the explanation and for opening the PR. We'll get to it hopefully soon but the next FF release is out in a couple day and we'll have to get that done first. I also don't want to change too much (or really anything) before @overdodactyl gets involved here because the script is kinda his baby.

overdodactyl:
Thanks for the pings and sorry I've been totally MIA as of lately.  I'll take a look at the PR this weekend :)

Thorin-Oakenpants:
any chance of getting this sorted out?

[...]

PR #755, opened 2019-07-03
Improve updater.sh with shellcheck

tyler-public:
As requested in #753 

Also for reference: https://github.com/koalaman/shellcheck

### DIFF
--- a/updater.sh
+++ b/updater.sh
@@ -156,7 +156,7 @@ readIniFile () { # expects one argument: absolute path of profiles.ini
     read -p 'Select the profile number ( 0 for Profile0, 1 for Profile1, etc ) : ' -r
     echo -e "\n"
     if [[ $REPLY =~ ^(0|[1-9][0-9]*)$ ]]; then
-      if grep "^\[Profile'${REPLY}" -A 4 "$inifile" | grep -v "^\[Profile'${REPLY}" > "$tfile"; then
+      if ! grep "^\[Profile'${REPLY}" -A 4 "$inifile" | grep -v "^\[Profile'${REPLY}" > "$tfile"; then
         echo "Profile${REPLY} does not exist!" && exit 1
       fi
     else
--- a/updater.sh
+++ b/updater.sh
@@ -2,11 +2,13 @@
 
 ## ghacks-user.js updater for macOS and Linux
 
-## version: 2.5
+## version: 2.5.1
 ## Author: Pat Johnson (@overdodactyl)
-## Additional contributors: @earthlng, @ema-pe, @claustromaniac
+## Additional contributors: @earthlng, @ema-pe, @claustromaniac, @iamtpage
 
 ## DON'T GO HIGHER THAN VERSION x.9 !! ( because of ASCII comparison in update_updater() )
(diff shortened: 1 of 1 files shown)


updater.sh
```diff
@@ -2,12 +2,14 @@
...
 
-## version: 2.5
+## version: 2.5.1
 ## Author: Pat Johnson (@overdodactyl)
-## Additional contributors: @earthlng, @ema-pe, @claustromaniac
+## Additional contributors: @earthlng, @ema-pe, @claustromaniac, @iamtpage
 
 ## DON'T GO HIGHER THAN VERSION x.9 !! ( because of ASCII comparison in update_updater() )
 
+set -e
```
claustromaniac: doesn't this break the (changed) `if grep...` statement in line 159?

iamtpage: `if <command>; then x else y fi` means *if the command succeeds, do x, otherwise do y*

where succeeds = 0 exit code of `command`, and not succeeds is anything but 0 exit code of `command`

I don't think it breaks anything, its just 2 grep commands being piped to a temp file (`$tmpfile`), and a decision being made if it succeeds or not

claustromaniac: grep exit codes:
```
   0     One or more lines were selected.
   1     No lines were selected.
   >1    An error occurred - syntax error in pattern, inaccessible input files, or other system error.
```
the conditional is meant to succeed if grep returns 1 or more (i.e. `$? != 0` as was written before)

iamtpage: Oh I see what you're saying, let me fix that

iamtpage: Reversed the condition, now it echos if grep returns non zero, will test with `set -e` and see how it behaves

iamtpage:
```
tyler@TYLER-DESKTOP:~$ cat test.sh
#!/bin/sh

set -e

if ! /bin/false; then
  echo "false"
else
  echo "true"
fi

tyler@TYLER-DESKTOP:~$ sh ./test.sh
false
tyler@TYLER-DESKTOP:~$
```

Looks like `set -e` doesn't break conditional checks if they return non-zero exit codes

earthlng:
What's wrong with `declare -r a=b`?
According to https://ss64.com/bash/declare.html that's perfectly valid syntax:

> SYNTAX
>       declare [-afFrxi] [-p] **[name[=value]]**

In fact, doesn't the following create a global variable first and then an additional local, readonly one with the same name?? ...
```bash
readIniFile () {
  inifile="$1"
  declare -r inifile
}
```
I mean, why else would there be a `local` cmd if variables assigned without it would be local anyway?
Here it's `declare` instead of `local` but that's functionally identical as mentioned in the above URL as well:

> When used in a function, declare makes each name local, as with the local command.

iamtpage:
@earthlng currently doing `declare -r a=b` is fine, however in the POSIX version (using `/bin/sh`), `declare` is undefined. I did a little bit of pre-work to make the POSIX version easier to migrate to. I'll resolve the double declaration with `inifile` thanks for catching it!

Thorin-Oakenpants:
abandoned

Issue #759, opened 2019-07-12
0864 is patched in FF68


Aeriem:
Preference 0864 was patched in FF68 to respect RFP: https://bugzilla.mozilla.org/show_bug.cgi?id=1492587

Sorry for the new issue, #20 is locked so I couldn't remind you there.

I'll be posting a PR shortly and update this post accordingly.

EDIT: My PR: #760

earthlng:
Thanks for bringing this to our attention and thanks for the PR! 👍 
We'll discuss and deal with it in the PR so I'm closing this. Thanks mate!

Aeriem:
Oh, didn't see your comment here. Forgot to check my notifications.
I'm glad I could be of assistance, thank you for the continued work on this repo. 😄

Aeriem:
Off-topic: BRUH!
I just realised I'm now a "contributor" to this repo since my PR got merged.
@claustromaniac, the creepy stalker is finally coming out of the shadows!

claustromaniac:
Congrats.

I went from nothing to collaborator, then to owner, and *then* to contributor. Looks like I may end up taking your place as a stalker at this rate.

Aeriem:
Don't worry, we'll take good care of you, in the shadows. 😈

Thorin-Oakenpants:
>  then to owner, and then to contributor

Yeah, what's up with that? Where did the three amigos go? -> https://github.com/orgs/ghacksuserjs/people  <- did you demote yourself?

claustromaniac:
> Don't worry, we'll take good care of you, in the shadows. 😈

That's... creepy. \*hides in da box\* Wait... I was already there. NVM

> did you demote yourself?

I left the organization. I had been hiding my "rank" from non-members for some time anyway (I set it to *private*); most people won't notice the difference.

PR #760, opened 2019-07-12
Preference 0864

Aeriem:
As I said in #759, this is my PR for pref 0864.


Thorin-Oakenpants:
@earthlng read my other comment. I think it's cleaner if we just remove the pref now it's obsolete for our purposes

Aeriem:
To answer your comment, I had thought about putting it into RFP alternatives but chose not to as I wasn't confident that it would belong there.
Also, I chose the deprecated / removed / renamed section because I mistook it for the archived section, silly me.
Lastly, I do not use the RFP alternatives section so I'm also counting on Earthling's opinion to make a choice about this pref, though I would tend to agree with your reasoning for removing it.

earthlng:
> I think it's cleaner if we just remove the pref now it's obsolete for our purposes

I agree 👍

PR #768, opened 2019-08-03
remove [NOTE] from 1704

rusty-snake:
`privacy.userContext.longPressBehavior = 1`; show a option for open a non-container tab


Thorin-Oakenpants:
- the user.js sets the options as 2: when the option is `2`, there is no menu item for a non-container: the information provided is correct
- when it is `1` there is an option, but that means *every* single new tab action now takes two clicks

rusty-snake:
> when the option is `2`, there is no menu item for a non-container

Because it would be useless to show a option to open a non-container-tab because a normal/short click opens it.

Thorin-Oakenpants:
Oh, that's what you meant. Yeah, it is redundant. I'll do a patch

Issue #772, opened 2019-08-16
Possibly redundant blanked out URLs and zeroed out capacities


Aeriem:
Hello!

Some prefs are in two parts: the first disables the tech, and the second blanks the then no longer used URL or zeroes out the capacity limit.
This seems redundant and I believe I remember you started clearing those out.
Preferences 0503, 0505 and 1003 are three examples I came up with, but I can search for more if you want me to.
```
/* 0503: disable Normandy/Shield [FF60+]
 * Shield is an telemetry system (including Heartbeat) that can also push and test "recipes"
 * [1] https://wiki.mozilla.org/Firefox/Shield
 * [2] https://github.com/mozilla/normandy ***/
user_pref("app.normandy.enabled", false);
user_pref("app.normandy.api_url", "");

/* 0505: disable System Add-on updates ***/
user_pref("extensions.systemAddon.update.enabled", false); // [FF62+]
user_pref("extensions.systemAddon.update.url", ""); // [FF44+]

/* 1003: disable memory cache
 * [NOTE] Not recommended due to performance issues ***/
   // user_pref("browser.cache.memory.enable", false);
   // user_pref("browser.cache.memory.capacity", 0); // [HIDDEN PREF]
```

Please correct me if I'm wrong, I'm simply looking to help keep the user.js as tidy as possible. :)
As Pants would say: Class, discuss!

Aeriem:
P.S.: I would gladly make the PR for this issue, should you give your approval! :)

Thorin-Oakenpants:
> I can search for more if you want me to

By all means, let's have a look at them. Always a good idea to revise content

> ... seems redundant and I believe I remember you started clearing those out

Redundancy vs defense in depth. I'm not sure what I have removed in the past, but each one was (99% sure) done on it's own merits. If you're referring to the `snappening` I did several months ago, that was a different reason, albeit the same goal: less clutter, less items to flip: and yes, a metric shit ton of (Safe Browsing, Reporting) URLs were cleaned out, but so were the other master prefs: so not the same.

---

Just making a point here that removing a secondary pref doesn't make the user.js any less work: i.e it still contains the itemized number with content

Looking at the above: 
- `0503` in depth against normandy is never a bad thing considering how often it gets changes
- `0505` at the time, it was unclear how this was going to work, and it is even broken in some configs, and there were/are plans to change it - hence the defense in depth. 
[...]

PR #773, opened 2019-08-17
Possibly redundant prefs

Aeriem:
https://github.com/ghacksuserjs/ghacks-user.js/issues/772
Section 5000: WELCOME & WHAT's NEW NOTICES is also a candidate but I'd like @Thorin-Oakenpants opinion first as they know more about this.


Thorin-Oakenpants:
like a few other of the URLs in prefs used by mozilla, these are good defense in depth prefs to have. The capacity pref was addressed elsewhere

PR #782, opened 2019-09-03
Update user.js

earthlng:
- TRR mode removals: https://bugzilla.mozilla.org/1552438
- 0105b (twice): disable-pref before url-clearing-pref
- 1405: you had the `[-]` url after the pref line


Thorin-Oakenpants:
disable-pref before url-clearing-pref: ahh OK, I just always put these things in alphabetical order. I don't think it makes a difference - the prefs.js is also ordered alphabetically

PR #789, opened 2019-09-07
0707: Add mode 5

rusty-snake:
https://wiki.mozilla.org/Trusted_Recursive_Resolver


Thorin-Oakenpants:
Pretty sure we decided not to include this, as at the time it doesn't really do anything. That said, as a user-set value, 5 would enable end-users to enforce it off when Mozilla flip the pref. - @earthlng

rusty-snake:
IMHO there should be an info that the value 5 exists and what it means, so that users can set it in their user-overriede.js if they want

earthlng:
Sure, why not. Using an enterprise policy to disable DoH also [sets the value to 5](https://dxr.mozilla.org/mozilla-central/source/browser/components/enterprisepolicies/Policies.jsm#571)

I'll add the missing comma and merge it. Thanks Rusty

PR #807, opened 2019-10-01
revamp 0200s

Thorin-Oakenpants:
- split geo related vs language/locale related
- we could remove the last item in the patch `intl.locale.requested`, or leave it in for info's sake: but I'm not even sure if what I wrote is still correct, and I'd rather not follow it: there's a UI for it
- we could add `privacy.spoof_english` in here as well (inactive)
- because these are active by default, for non en-US bundles we could add something about resetting `2010-2013` - as has been brought up at least three times in issues
- this is a draft, so fixups to typos, better wording etc are welcome

### DIFF
--- a/user.js
+++ b/user.js
@@ -177,12 +177,12 @@ user_pref("javascript.use_us_english_locale", true); // [HIDDEN PREF]
 user_pref("intl.regional_prefs.use_os_locales", false);
 /* 0213: enforce fallback text encoding to match en-US
  * When the content or server doesn't declare its charset the browser will fallback
- * to the "Current locale" based on your application locale
+ * to the "Current locale" based on your application language (see 0214)
  * [SETTING] General>Language and Appearance>Fonts and Colors>Advanced>Text Encoding for Legacy Content
  * [TEST] https://hsivonen.com/test/moz/check-charset.htm
  * [1] https://trac.torproject.org/projects/tor/ticket/20025 ***/
 user_pref("intl.charset.fallback.override", "windows-1252");
-/* REMOVE: set Firefox language [FF59+] [RESTART]
+/* 0214: set Firefox language [FF59+] [RESTART]
  * Go to the end of about:support to view Internationalization & Localization settings
  * If set to empty, the OS locales are used. If not set at all, default locale is used
  * This is the language used in menus, about pages, messages, and notifications from Firefox ***/
--- a/user.js
+++ b/user.js
@@ -188,7 +188,6 @@ user_pref("intl.charset.fallback.override", "windows-1252");
  * This is the language used in menus, about pages, messages, and notifications from Firefox ***/
    // user_pref("intl.locale.requested", "en-US"); // [HIDDEN PREF]
 
-
 /*** [SECTION 0300]: QUIET FOX
      Starting in user.js v67, we only disable the auto-INSTALL of Firefox. You still get prompts
      to update, in one click. We have NEVER disabled auto-CHECKING, and highly discourage that.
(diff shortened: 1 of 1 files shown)


Thorin-Oakenpants:
bump

Thorin-Oakenpants:
bump bump

earthlng:
bump bump bump

earthlng:
> split geo related vs language/locale related

👍 

Where's the UI for `intl.locale.requested`?
I think this pref only works if you have language packs installed. Like, if you use a non-EN-US version without any lang packs, forcing this pref to en-US probably doesn't do anything or it might even break FF, IDK.

FYI I also found another pref (`intl.locale.privacy.web_exposed`) in one of the files where `intl.locale.requested` is used but IDK why or where `intl.locale.requested` would be exposed to the web in the 1st place.

So yeah, lots of prefs with presumably some interaction between them and I don't really know what we should do here. Best to do some testing with non-en-US versions and some lang. packs but I don't have time for that atm

Thorin-Oakenpants:
`intl.locale.requested` - the UI
- General > Language and Appearance > Choose the languages used to display menus...

rusty-snake:
setting 0214 may require increasing 2660. A cross note would be helpfull, because if your lang-packs are broken you first look in a section like languages and not in a section called extensions.

Thorin-Oakenpants:
OK. So I initially suggested getting rid of the pref, because it just makes things confusing. Then in a subsequent PR edit, I left it in so I could refer to it in `0213`. But now I'm keen to just rip it out. The "app language" *shouldn't* have anything anymore to do with web content or leaks (except for PoCs that need patching)

Thorin-Oakenpants:
PS: there's been two attempts by me, and other issues (atomGit started one) where the whole language thing was going to be investigated. Part of the problem is it's messy AF, and in a state of change - that is the UI recently changed, is still changing(?), and you can add language packs on demand (currently requires restart, but will become restartless), etc - and additionally, in all of this, I'm not sure exactly what we are testing. I mentioned this to acat months ago - see https://trac.torproject.org/projects/tor/ticket/31767

I can easily set up a few non en-US portables: I already have some, as well as 31 non en-US Tor Browsers based on ESR68.

---

However, all of that is immaterial to the "app language" (assuming it has zero to do with any FPing stuff). If we can agree on cleaning this one up, push this commit, and worry about all the rest later

I'm going to pump for ripping it out - just like acat suggests in the tor ticket: we shouldn't use it directly. Users should use the UI, because it initiates the language download etc. Otherwise we need to add that to the pref info

Thorin-Oakenpants:
- https://dxr.mozilla.org/mozilla-central/search?q=privacy.web_exposed&redirect=false
- https://dxr.mozilla.org/mozilla-central/source/intl/docs/locale.rst#435

> The pref has no function while :js:`privacy.spoof_english` is set to 2, where *"en-US"* will always
be returned.

Umm, so what happens to those of us on en-US builds who do not ever set `privacy.spoof_english`. IDK. I don't think this is something we need to worry about

Thorin-Oakenpants:
> `intl.locale.requested` - the UI
> 
>     * General > Language and Appearance > Choose the languages used to display menus...

I hate this language shit. That UI is now gone, sorry: i was looking in my FF68 (still haven't updated). This is what I mean about all the UI changes and other stuff.

Thorin-Oakenpants:
I'm going to push this in a few hours (I'll open yet another issue for the rest including if anyone can tell WTF I am supposed to be testing) .. it's holding me up.

PR #828, opened 2019-10-29
Rip out 1607

rusty-snake:
Use the tor-browser if you want to visit onion sites


Thorin-Oakenpants:
I specifically wanted to leave that one in for two reasons. First, the pref name itself is a bit ambiguous and I've seen lots of people thinking it needs to be false in order to visit hidden services. And secondly, so I can tell them to go use Tor Browser

PR #837, opened 2019-11-08
re-add relevant deprecated items for ESR users

earthlng:
makes the prefsCleaner scripts useful again for users updating from ESR60 to ESR68


Thorin-Oakenpants:
yeah, I actually thought about that the other day. But they're going in the next alpha

Issue #839, opened 2019-11-08
ciphers


Thorin-Oakenpants:
I'd like to strip out the ciphers BS. At  the end of the day, there is no need for someone to disable any of them: they should leave it up to Mozilla and their experts to decide when to deprecate something

At the very least this
```
   Note that your cipher and other settings can be used server side as a fingerprint attack
   vector, see [1] (It's quite technical but the first part is easy to understand
   and you can stop reading when you reach the second section titled "Enter Bro")

   Option 1: Use defaults for ciphers (1260's). There is nothing *weak* about these, but
             due to breakage, browsers can't deprecate them until the web stops using them
   Option 2: Disable the ciphers in 1261, 1262 and 1263. These shouldn't break anything.
             Optionally, disable the ciphers in 1264.

   [1] https://www.securityartwork.es/2017/02/02/tls-client-fingerprinting-with-bro/
```

can be changed to
```
   Note that your cipher and other settings can be used in server side fingerprinting
```

We could even not have that and just change this
```
/** CIPHERS [see the section 1200 intro] ***/
```

to 
```
/** CIPHERS [don't fuck with these, u morons] ***/
```

@earthlng : please advise

Thorin-Oakenpants:
^^ bump

TomasGutierrez0:
Mozilla is planning to [disable support for TLS 1.0 and 1.1](https://blog.mozilla.org/security/2018/10/15/removing-old-versions-of-tls/) in march 2020. It would be pointless to disable TLS 1.0 and 1.1 ciphers if later they are going to be disabled anyway.

I don't know if Firefox disables them by default or not but user.js should disable null ciphers as it only authenticates but doesn't encrypt the connection which is a security problem.

Thorin-Oakenpants:
^^ yup, that's already noted, and the prefs are inactive (commented out)

```js
/* 1202: control TLS versions with min and max
 * 1=TLS 1.0, 2=TLS 1.1, 3=TLS 1.2, 4=TLS 1.3
 * [WARNING] Leave these at default, otherwise you alter your TLS fingerprint ....
```

What I'm talking about here is trimming fat and keeping things as simple as possible

earthlng:
To me that's useful information because it highlights the prefs (among all the other `security.ssl3.*` prefs) that should ideally be disabled by default but still aren't.

And nobody is forced to read the link in [1] but those who want to know more have a good starting point.

[...]

PR #844, opened 2019-11-22
simplify ciphers #839

Thorin-Oakenpants:
- let's not encourage (remove options 1, 2) changing your cipher suite FP
- remove "it's quite technical ..." (everything is technical to someone), trim to one line
- add test link so users can just see that it's FP'able
- reinforce not to fuck with the cipher suite in the cipher's sub-section


Thorin-Oakenpants:
I'm not 100% happy about getting too verbose in this bit
```
/** CIPHERS
   Do not meddle with your cipher suite [see the section 1200 intro]. There is practically zero
   gain doing this, but a huge downside in passive fingerprinting. Leave it up to the experts ***/
```

and originally had
```
/** CIPHERS [WARNING: do not meddle with your cipher suite: see the section 1200 intro]
```

which I think is better but am afraid people will miss it. Then again, they should learn to fucking read.

Issue #636, opened 2019-02-07
DNT - is it still relevant?


seanob86:
Recent articles of Apple removing DNT in latest beta of Safari. 

https://www.macrumors.com/2019/02/06/apple-removes-safari-do-not-track-option/

Is DNT still relevant today even with RFP and utilising extensions such as ublock origin in “medium” mode?

I’ve also read that DNT can be used for fingerprinting. 

The current version of user.js has DNT set to true. Are there any plans to reconsider this option?

Thorin-Oakenpants:
Forget about what apple or google or microsoft etc do - because this is only about Firefox: i.e you are easily distinguishable as a Firefox user - so what anyone else does is irrelevant.

DNT is a mess, let's be honest. For starters it must be opt-in: that was the only way to get advertising networks on board. And then basically no advertising networks came on board - so it is essentially useless .. BUT .. since the [GDPR](https://en.wikipedia.org/wiki/General_Data_Protection_Regulation) came into affect, it may have some bearing, IDK. I read somewhere that a study found tracking in the EU dropped around 30% (not from DNT, just overall as companies made themselves compliant).

Anyway .. back to DNT. So this has uses in Firefox (remember: forget about other browsers) - it is actually used to NOT allow GA (google analytics) in some places: I think this is correct. e.g an AMO (mozilla addons site), in activity stream?, etc. Not that I am concerned with that per se, because there are other ways to make extensions block GA etc on AMO, and to hide the "Get Add-ons" panel in about:addons, and to not use Activity Stream etc. But it does have a role in Firefox - e.g it is enabled regardless of the DNT setting in Private Browsing windows (the opt-in is implied because you choose to use PB mode even though the landing page for PB mode doesn't mention it - not sure how that's meant to work)

In fingerprinting terms, it can have three states: undefined, true (on), false (off). That's not a lot of entropy, but it's still entropy. Don't believe a single stat on this from sites like Panopticlick etc - all of which seem to suggest the figure is 50/50. 95% of the world wouldn't even have a clue what it is and have never changed it.

We used to have it on for a while back in the dark ages. Then it was off for 3 years, because, yeah, it's useless and the feeling was it would just increase entropy. Then we recently decided to enable it because it can do some good, and FP'ing is not really the issue - everyone is already unique if you dig deep enough, and you control most FP'ing by controlling the attack surface (3rd party etc) at this point.

[...]

PR #852, opened 2019-11-27
Update user.js

rusty-snake:
1610: fix description

### DIFF
--- a/user.js
+++ b/user.js
@@ -829,7 +829,7 @@ user_pref("network.http.referer.XOriginTrimmingPolicy", 0); // [DEFAULT: 0]
  * [1] https://bugzilla.mozilla.org/1305144 ***/
 user_pref("network.http.referer.hideOnionSource", true);
 /* 1610: ALL: enable the DNT (Do Not Track) HTTP header
- * [NOTE] DNT is enforced with Tracking Protection regardless of this pref
+ * [NOTE] DNT is enforced with Tracking Protection in strict or custom mode regardless of this pref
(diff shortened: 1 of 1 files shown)


Thorin-Oakenpants:
ETP has three modes: standard-strict-custom : and custom can be any configuration

What makes you think the ETP mode decides DNT? It's decided by the "tracking content" part: it may even now be integrated with the "cross-site tracking cookies" part (and that's on in standard mode).

Other things to think of here
- ETP is being constantly ramped up: e.g fingerprinters in 72+ moved into standard. There is a ticket to also change "tracking content" to all windows by default
- DNT's specs (by design) are opt-in: otherwise it breaks the standard
- by using PB mode, users were "silently (?) " agreeing to use DNT
- but turning on tracking protection in all windows might break that spec
- the thing with DNT, is that it's not really maintained at Mozilla IMO - and I don't think anyone there really cares for it, but it's still kept probably because of a couple of sites: guys like Arthur, Steven, Ehsan (and gk from tor) would all be happy for it to be ripped out: and that's the general consensus I get from them (because I asked a question about it a meeting with them, that I was invited to)

I'd rather not get too detailed in the comment about DNT being sent in ETP, that way I don't have to maintain it.

rusty-snake:
+1 for you, I was a bit confused with the German translation on about:preferences and keeping the maintenance low sound good.

But maybe it can commented or removed (see https://github.com/ghacksuserjs/ghacks-user.js/issues/636#issuecomment-559613724).

PR #865, opened 2019-12-17
0000: config.xhtml in FF73+

earthlng:


Thorin-Oakenpants:
go ahead and squash and merge

Issue #862, opened 2019-12-16
browser.contentblocking.category blocks ETP prefs


crssi:
``` js
/* 2701 */
user_pref("network.cookie.cookieBehavior", 1);
```
is ignored, unless
``` js
user_pref("browser.contentblocking.category", "custom");
```

`browser.contentblocking.category` is missing in user.js  and thus set by default to `standard`.
ATM user must manually change in GUI to `custom`.

Cheers

Thorin-Oakenpants:
if you change your cookie behavior to 1, the custom setting is automatically applied. Try it on a new profile.

crssi:
No, it is not. 
I have tried with a new profile before posting an issue 😉 .
Maybe this is new to latest FF.

Thorin-Oakenpants:
Lemme do some tests: I have ESR60, ESR68, FF60-FF73 portables with practically vanilla profiles all here as my test suite

crssi:
Its not only the latest FF, it looks like FF 70 also.

Thorin-Oakenpants:
I cannot replicate

🐧 info for reference: network.cookie.cookieBehavior
 * 0=Accept cookies and site data
 * 1=(Block) All third-party cookies
 * 2=(Block) All cookies
 * 3=(Block) Cookies from unvisited websites
 * 4=(Block) Cross-site and social media trackers (FF63+) (default FF69+)

🐧 Tests / STR

**FF71-73 (default 4)**
- tab1: about:preferences#privacy = standard
- tab2: about:config change pref to `1`
- tab1: **automagically changes to custom** - change it back to standard
- tab2: about:config change pref to `2`
- tab1: **automagically** changes to custom
- not going to test 0 or 3
- Interestingly, if you reset the pref (in about:config), it stays in custom mode (even after a restart) even though nothing is different. This might be by design and is immaterial IMO: it's just cosmetic.

**ESR68 (default = 0)
- results with `1` and `2` **and** `4`
- they all **automagically** change to custom

earthlng:
Prefs in about:config are (often) tied to a prefs-observer which then does things like "automatically" change this thing to "custom" etc. But that's not (or not always) the case when a pref is applied from a user.js.
So I guess the STR would be to let the user.js apply the cookie pref and then check if the GUI correctly changed to custom

Thorin-Oakenpants:
[...]

PR #866, opened 2019-12-17
2701: make sure cookieBehavior is always honored

earthlng:
see https://github.com/ghacksuserjs/ghacks-user.js/issues/862

thanks @crssi !

### DIFF
--- a/user.js
+++ b/user.js
@@ -1245,6 +1245,7 @@ user_pref("_user.js.parrot", "2700 syntax error: the parrot's joined the bleedin
  * 0=Accept cookies and site data, 1=(Block) All third-party cookies, 2=(Block) All cookies,
  * 3=(Block) Cookies from unvisited websites, 4=(Block) Cross-site and social media trackers (FF63+) (default FF69+)
  * [NOTE] You can set exceptions under site permissions or use an extension
+ * [NOTE] Enforcing category to custom ensures ETP related prefs are always honored
  * [SETTING] Privacy & Security>Content Blocking>Custom>Choose what to block>Cookies ***/
 user_pref("network.cookie.cookieBehavior", 1);
 user_pref("browser.contentblocking.category", "custom");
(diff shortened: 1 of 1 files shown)


Thorin-Oakenpants:
I think we need to add some info

[NOTE] Enforcing category to custom ensures the cookie pref is honored due to a Firefox bug

earthlng:
sure, why not.

```js
 * [NOTE] Enforcing category to custom ensures the cookie pref is always honored. see [x]
 * [x] https://bugzilla.mozilla.org/...
```

Thorin-Oakenpants:
yup, we'll just add the bug number when it lands. go ahead and do that and commit

Thorin-Oakenpants:
wait a sec

Thorin-Oakenpants:
in my email: i said this

> Note: we have a workaround: which is to enforce the ETP category to custom (which doesn't hurt even if all ETP prefs are at default and technically it should be at standard)

I think this might be good to work in (just not that wordy) .. i.e the category pref at custom "can't hurt" regardless of what ETP config they have

earthlng:
wait, you sent an email to your top-notch security/privacy contacts about this rather insignificant edge-case scenario?!? so now *they* have to open a bugzilla ticket? Don't you think they have better things to do than *that*? I mean it's not for me to decide but I think you're stretching it a bit here.

> pref at custom "can't hurt" 

I don't think that's necessary but whatever

Thorin-Oakenpants:
Yes, I contacted via email for **them** to open
- I think it is more than an edge case: enterprise policies might be affected (if they have ETP), and enterprises using cfg (or whatever it is) will have this issue
- re enterprise: I've found Kaply to be a little dismissive at times (just my perception), so coming via tom might be better if he's involved (not that we don't have a workaround anyway)
- because you publicly asked me to open the ticket, I can't unless I create another account

> I don't think that's necessary but whatever

I can see people asking .. my prefs are at default but it still shows custom. Not a deal breaker. I'm going to bed.

If you want, add the following and commit. We'll add the bugzilla later if they create one
```
 * [NOTE] Enforcing category to custom ensures the cookie pref is always honored. see [1]
 * [1] https://bugzilla.mozilla.org/... (bug number to come)
```

PR #926, opened 2020-04-13
Add notice about breakage of ibus

graywolf:


Thorin-Oakenpants:
Thanks for the info:

- I don't want to "pollute" (for lack of a better word) the user.js with breakage info: especially for "obscure" (for lack of a better word) things like gaming console input (excuse my ignorance if this is not what you're talking about, re xim). I know nothing about ibus, sorry : https://github.com/ibus/ibus

Changing anything from Firefox's defaults can and probably will cause breakage somewhere

Maybe @earthlng has some thoughts ... but for me it's a no (and Mozilla might fix that 6 hr old bug)

graywolf:
Sorry, I probably was not clear. It's input method for writing japanese (well, any CJK language) on english keyboard (kinda like https://snapforum.s3.amazonaws.com/original/2X/5/50ad0724696f8c6777e92ca075df636308b2f2c0.gif , it's actually used in Japan as well).

Only reason I decided to make the pull request is that the breakage is completely non-intuitive, there should be no relation at all. After spending like 2 hours debugging what is wrong I've wanted to spare others the same process.

But I get your point.

Thorin-Oakenpants:
> After spending like 2 hours debugging ...

When I use the [binary/halving](https://github.com/ghacksuserjs/ghacks-user.js/wiki/1.4-Troubleshooting#small_orange_diamond-manual-troubleshooting) method, at most it takes me five to ten minutes - maybe four or five resets and then I'm into flipping prefs. Granted, I have vanilla profiles already set up etc in my test suite, and a good knowledge for educated guesses

> kinda like ... pic

Ahhh. OK. That makes more sense

The pref `javascript.use_us_english_locale` =`true` is used by Tor Browser when a non-English build (on first web site connection) asks the user if they wish to spoof as en-US to websites, and they select 'yes' . And Tor Browser ships japanese, korean, zh-CH, zh-TW .. so it's definitely a resist fingerprinting breakage type bug they should look at. I guess the question is how widespread the problem would be

I'll give the Tor Uplift team a nudge on the bugzilla

Thorin-Oakenpants:
@graywolf .. I pinged Tom and he left you a message on bugzilla

Thorin-Oakenpants:
@graywolf  I can't edit your PR, so I'll paste my revision, which I'm happy to merge if you update

```js
/* 0211: enforce US English locale regardless of the system locale
 * [SETUP-WEB] May break some input methods e.g xim/ibus for CJK languages, see [2]
 * [1] https://bugzilla.mozilla.org/867501
 * [2] https://bugzilla.mozilla.org/1629630 ***/
user_pref("javascript.use_us_english_locale", true); // [HIDDEN PREF]
```

graywolf:
> I pinged Tom and he left you a message on bugzilla

I've tried the mozregression tool and it seems like this never worked.

> if you update

Your formulation is way better, I've updated the PR. Thx :)

PR #944, opened 2020-05-18
Update user.js

Klein1985:


Thorin-Oakenpants:
the pref says
> /* 0411: **disable** SB checks for downloads (both local lookups + remote)
>  * This is the master switch for the safebrowsing.downloads* prefs (0412, 0413)
>  * [SETTING] Privacy & Security>Security>... "Block dangerous downloads" ***/
>    // user_pref("browser.safebrowsing.downloads.enabled", **false**);

disable means the setting needs to be false. Descriptions match the value regardless of whether or not the pref is inactive or active (i.e commented out or not)

PR #947, opened 2020-05-25
fix updater.sh when dealing with multiple overrides

mloberg:


Thorin-Oakenpants:
This is a job for @overdodactyl and/or @earthlng

earthlng:
good find, it was indeed broken without this fix. Thanks @mloberg for the pull request!

PR #953, opened 2020-06-09
cookieBehavior: value 5

Thorin-Oakenpants:
@earthlng This isn't actually in the UI except on Nightly (79). Is the value 5 actually usable in 77? Or should we leave it out of the user.js until they deem it's ready for stable?

### DIFF
--- a/user.js
+++ b/user.js
@@ -1,7 +1,7 @@
 /******
 * name: ghacks user.js
-* date: 24 May 2020
-* version 77-alpha
+* date: 09 Jun 2020
+* version 77-beta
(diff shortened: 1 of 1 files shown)


Thorin-Oakenpants:
OK, never mind, this is a waste of time for 77 since the fix didn't even land until 78, and the fix is **silent** (the pref value is not changed to reflect reality), and 77 (and 78) don't even have the option in the UI yet.

I guess we can add it in 79 or 80 or something

earthlng:
Yeah I think we should wait until they expose value 5 in the UI in stable. Just keep this PR open until then.

PR #958, opened 2020-06-18
update trac tor tickets

Thorin-Oakenpants:
- note: 1403 is a different path: `legacy`
- deprecated is excluded (the section will be cleaned out at EOL for ESR68 anyway)


earthlng:
The old links all still work. Will they not create new tickets on TRAC anymore?

I actually prefer the (old?) trac links because it displays everything without needing to allow JS.

Thorin-Oakenpants:
The old trac is no longer used, so any open tickets won't get updated comments, status etc. We have at least one open trac ticket (0212 about txt-encoding). Same with any new/open tickets we add going forward

The old trac won't even be up after Dec [source](https://trac.torproject.org/projects/tor) while it lasts
> This server will be completely archived and retired in December 2020

We might as just switch everything to the new one for ESR78+

FYI: https://help.torproject.org/tsa/howto/gitlab/ - you can scroll down to migration, migration issues, etc .. including **Trac Archival**

earthlng:
ok thanks for the info 👍

PR #962, opened 2020-06-27
78 deprecated, add 2032

Thorin-Oakenpants:


Thorin-Oakenpants:
2032 needs some work, also, scan the rest because I'm human, thanks

earthlng:
As I understand it, we want this to be value 2 to get the same behavior as user-gestures-needed=false.

I think we can keep the number at 2031 and maybe even keep the title too because it's easier to understand. We also probably don't need to include the timeout pref. And I'd keep the pref active, as before. So maybe something like this:
```js
/* 2031: disable autoplay of HTML5 media if you interacted with the site [FF78+]
 * 0=sticky (default), 1=transient, 2=user
 * [1] https://html.spec.whatwg.org/multipage/interaction.html#sticky-activation ***/
user_pref("media.autoplay.blocking_policy", 2);
```

re: cosmeticAnimations - shouldn't we replace that with `ui.prefersReducedMotion=1` ?
RFP spoofs that value for **website content** and therefore it's safe and not detectable if someone still wants to disable all the UI animations that are now controlled by this pref, right?
I think in FF79+ it will also cover the expanding urlbar and users don't need userchrome for that anymore if they set this pref.

rusty-snake:
Firefox takes the OS settings for animations into account. https://developer.mozilla.org/en-US/docs/Web/CSS/@media/prefers-reduced-motion#User_preferences
If you have a slow system you disable animations system-wide. IMHO we don't need to add it.

Thorin-Oakenpants:
yeah, that looks cleaner

---

`ui.prefersReducedMotion` is already in the user.js, inactive at `0` in the RFP alts section (I don't want to list it twice), because it's FP'able. Currently RFP doesn't give a shit about what principal is used, so setting it for RFP users won't do jack shit, and non-RFP users should be at `0` if they cared. I don't like how they tied a web content setting to chrome (but I can see why)

RFP users will report spoofed values to chrome and web-extensions until they sort it out
- https://bugzilla.mozilla.org/show_bug.cgi?id=1635603
- https://bugzilla.mozilla.org/show_bug.cgi?id=1450398

earthlng:
Yeah that's true but I have a pretty fast system and don't mind some of the animations in my OS but I still want to disable the animations in FF because I find them distracting, especially the ugly AF expanding urlbar.

earthlng:
> so setting it for RFP users won't do jack shit

that's not true. I have it set to 1 and so far it definitely disables the reload animations in 77. I assume it'll also stop the expanding urlbar in 79 despite RFP but I haven't tested it.

rusty-snake:
>  I assume it'll also stop the expanding urlbar in 79 despite RFP but I haven't tested it.

`ui.prefersReducedMotion=1` disables it. (tested with nightly-79)

Thorin-Oakenpants:
review again or squash and merge when you're ready

@rusty-snake : well, we had some stuff in the personal section, so if we can replace something that gets taken out, it's probably worth it. That said, prefers-motion is FPable

@earthlng I agree, that megabar sucks donkeys balls - but it's a personal thing, I mean if you get used to it, then it's no longer an issue: most people around the world couldn't care less. And I ditched the bouncing ball and throbber when quantum came out

Thorin-Oakenpants:
AFAIK PRM should be the same as prefers-colors-scheme (PCS): i.e RFP overrides it for everything. I've tested PCS a fair bit. Maybe they did a special exemption with PRM because of epilepsy?

Are you sure? RFP=`on` and PRM=1 returns `no-preference` on web pages **and** stops animations in chrome? In FF78?

Thorin-Oakenpants:
my machine
- default: PRM on web page = `no-preference`
- RFP on: PRM on web page = `no-preference`
- add `ui.prefersReducedMotion` = `1` - PRM on web pages = `reduce`

earthlng:
Oops, I just merged it before I saw your 2 latest comments

earthlng:
did you restart after you set prefersReducedMotion to 1? I'm pretty sure it needs a restart
But I only tested it in 77, not 78

earthlng:
RFP=on and PRM=1 returns `no-preference` on your TZP *and* stops (some?) animations in chrome in my FF77

Thorin-Oakenpants:
> did you restart after you set prefersReducedMotion to 1? I'm pretty sure it needs a restart

Yeah, I was just doing STR for Tom in an email, and that's what it is, as I went true double checking it. A restart is required. I guess it's designed that way. I think they should take RFP and FPI away from the webext privacy API

Other things leak when people toggle RFP on and off. I've seen workers leak your real UA, and date/times get leaked as well in the main document (more than in the past). Something happened in 77 and I'm still trying to nail it down.

---

So do we still want to do anything with the PRM pref?

earthlng:
> I'm pretty sure it needs a restart

.. ie to work correctly **in combination with RFP**. The animations in chrome stop immediately but you might get `reduce` on web pages until you restart.

Thorin-Oakenpants:
> So do we still want to do anything with the PRM pref?

Users with RFP on should be aware they can use this to kill the distractions, just as long as we make it inactive. It's a shame 5000s come after RFP alts. Any ideas?

earthlng:
The only solution I see (if we want to include it at all) is to add it under 4500 instead of 5000. That way it will get overwritten for non-RFP when they enable the RFP-alts section. We could then even make it active by default if we want.

I also don't like that we referred back to a pref in the 4600 section for the "replaced by" part in the newly-removed cosmeticAnimations pref. If we'd add PRM to 4500 we could change that too and referrer there instead. That would be less confusing IMO

Thorin-Oakenpants:
> I also don't really like that we referred back to a pref in the 4600 section

I know, the whole thing's messy AF. Show what you want to do in a PR (but I get the idea: these prefs are only safe with RFP hence in the RFP section), but I'm torn on making it active because it breaks the rules. Then again, fuck that megabar and stupid bouncing ball etc

earthlng:
we don't have to move it, we can just have it twice.

IDK if the other one behaves the same as PRM

Thorin-Oakenpants:
> IDK if the other one behaves the same as PRM

It does. It overrides the system setting, and RFP overrides the pref.

Issue #959, opened 2020-06-23
ToDo: diffs FF77-FF78


earthlng:
FF78 is [scheduled](https://wiki.mozilla.org/Release_Management/Calendar) for release June 30th

[FF78 release notes](https://www.mozilla.org/firefox/78.0/releasenotes/) [when ready]
[FF78 for developers](https://developer.mozilla.org/docs/Mozilla/Firefox/Releases/78)
[FF78 compatibility](https://www.fxsitecompat.dev/en-CA/versions/78/)
[FF78 security advisories](https://www.mozilla.org/en-US/security/advisories/mfsa2020-24/)

---

### other
* [x] add `ui.prefersReducedMotion=1` - https://github.com/ghacksuserjs/ghacks-user.js/commit/3d18af19e3936beecca8506dc57319a428b98dc2
* FF76+: sort out `2200`: `dom.disable_window_open_feature` - see [1507375](https://bugzilla.mozilla.org/show_bug.cgi?id=1507375), [compat](https://www.fxsitecompat.dev/en-CA/docs/2020/browser-ui-can-no-longer-be-controlled-with-window-open-features-parameter/)

100 diffs ( 63 new, 19 gone, 18 different )

### new in v78.0:
* [x] pref("browser.urlbar.suggest.topsites", true); - https://github.com/ghacksuserjs/ghacks-user.js/commit/618f7bed3f693f0931689740b2ce540e64b0e651
* [x] pref("browser.urlbar.dnsResolveSingleWordsAfterSearch", 1); - https://github.com/ghacksuserjs/ghacks-user.js/commit/1a389c021417833ccd688a52d05cd4b0b08734fa
* [x] pref("media.autoplay.blocking_policy", 0); - https://github.com/ghacksuserjs/ghacks-user.js/commit/77ecef8be3a0a6b1a0f32c9d9ef501478065b00f
* FYI
   - pref("network.dns.disabled", false);
   - pref("layout.css.font-visibility.level", 3); - not important; to be dealt with in #970 
   - pref("privacy.dynamic_firstparty.use_site", true); // changes OA key of dFPI to use `sites` instead of `domains`
   - pref("privacy.firstparty.isolate.use_site", false); // changes OA key of FPI to use `sites` instead of `domains`
   - pref("privacy.partition.network_state", false); // we want this to be true eventually as it covers a few things FPI missed

### removed, renamed or hidden in  v78.0:

[...]

PR #967, opened 2020-07-04
various, #959

Thorin-Oakenpants:
- `0105e`
   - we disable AS, and clearing the default does not stop someone adding their own (it's a different pref) should they chose to use AS
   - otherwise, to get rid of the topsites when clicking in an empty urlbar, users have to either know about this pref or enable AS and go manually remove them all
   - this is cleaner than disabling topsites with `browser.newtabpage.activity-stream.default.sites` as that actually breaks the functionality (should users want to use AS and topsite suggestions)
   - we don't disable topsite suggestions: 0850a: 78+, `browser.urlbar.suggest.topsites` is inactive
   - this is the list you see when you click in an empty urlbar, and obviously if you have no topsites then the 0850a pref is rendered useless for suggestions as you type, but doesn't break the functionality if a user does add their own topsites
   - capisce? easy peasy? clear as mud?
- 0850a
   - we could elaborate that this is for suggestions when you start typing/pasting (i.e not the same as topsites) - IDK
- settings (changed)
   - we can remove these two when ESR68 hits EOL, meanwhile I just tagged them as (FF72-)
- settings (not changed)
   - `0517` does not seem to exist: [SETTING] Privacy & Security>Forms and Autofill>Autofill addresses (FF74+)
   - I tested vanilla profiles in 74+, and system addon `formautofill@mozilla.org.xpi` is present
   - maybe I'm doing something wrong
- `0420`: ui.PRM
   - maybe you can word it better

### DIFF
--- a/user.js
+++ b/user.js
@@ -125,6 +125,9 @@ user_pref("browser.newtabpage.activity-stream.showSponsored", false);
 user_pref("browser.newtabpage.activity-stream.feeds.discoverystreamfeed", false); // [FF66+]
 /* 0105d: disable Activity Stream recent Highlights in the Library [FF57+] ***/
    // user_pref("browser.library.activity-stream.enabled", false);
+/* 0105e: clear default topsites
+ * [NOTE] This does not block you from adding your own */
+user_pref("browser.newtabpage.activity-stream.default.sites", false);
 /* 0110: start Firefox in PB (Private Browsing) mode
  * [NOTE] In this mode *all* windows are "private windows" and the PB mode icon is not displayed
  * [WARNING] The P in PB mode is misleading: it means no "persistent" disk storage such as history,
@@ -175,7 +178,7 @@ user_pref("javascript.use_us_english_locale", true); // [HIDDEN PREF]
 /* 0212: enforce fallback text encoding to match en-US
  * When the content or server doesn't declare a charset the browser will
  * fallback to the "Current locale" based on your application language
- * [SETTING] General>Language and Appearance>Fonts and Colors>Advanced>Text Encoding for Legacy Content
+ * [SETTING] General>Language and Appearance>Fonts and Colors>Advanced>Text Encoding for Legacy Content (FF72-)
  * [TEST] https://hsivonen.com/test/moz/check-charset.htm
  * [1] https://gitlab.torproject.org/tpo/applications/tor-browser/-/issues/20025 ***/
 user_pref("intl.charset.fallback.override", "windows-1252");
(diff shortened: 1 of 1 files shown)


Thorin-Oakenpants:
@guser-sudo feel free to join in

user.js
```diff
@@ -125,6 +125,9 @@ user_pref("browser.newtabpage.activity-stream.showSponsored", false);
 user_pref("browser.newtabpage.activity-stream.feeds.discoverystreamfeed", false); // [FF66+]
 /* 0105d: disable Activity Stream recent Highlights in the Library [FF57+] ***/
    // user_pref("browser.library.activity-stream.enabled", false);
+/* 0105e: clear default topsites
+ * [NOTE] This does not block you from adding your own */
```
rusty-snake, line 129:
```suggestion
 * [NOTE] This does not block you from adding your own ***/
```

user.js
```diff
@@ -175,7 +178,7 @@ user_pref("javascript.use_us_english_locale", true); // [HIDDEN PREF]
 /* 0212: enforce fallback text encoding to match en-US
  * When the content or server doesn't declare a charset the browser will
  * fallback to the "Current locale" based on your application language
- * [SETTING] General>Language and Appearance>Fonts and Colors>Advanced>Text Encoding for Legacy Content
+ * [SETTING] General>Language and Appearance>Fonts and Colors>Advanced>Text Encoding for Legacy Content (FF72-)
```
rusty-snake, line 181:
```suggestion
 * [SETTING] General>Language and Appearance>Fonts and Colors>Advanced>Text Encoding for Legacy Content [FF72-]
```

user.js
```diff
@@ -205,7 +208,7 @@ user_pref("app.update.auto", false);
    // user_pref("extensions.getAddons.cache.enabled", false);
 /* 0308: disable search engine updates (e.g. OpenSearch)
  * [NOTE] This does not affect Mozilla's built-in or Web Extension search engines
- * [SETTING] General>Firefox Updates>Automatically update search engines ***/
+ * [SETTING] General>Firefox Updates>Automatically update search engines (FF72-) ***/
```
rusty-snake, line 211:
```suggestion
 * [SETTING] General>Firefox Updates>Automatically update search engines [FF72-] ***/
```

Thorin-Oakenpants: square brackets are only for numbered items of individual prefs, search for `(FF` and you'll see what I mean

earthlng:
`browser.newtabpage.activity-stream.default.sites` is a string pref. Otherwise 👍

Thorin-Oakenpants:
what about `0517` 's [setting] tag ?

earthlng:
re: 0517 - I tested it in a vanilla FF78. Because by default `extensions.formautofill.available` is set to `detect` and `extensions.formautofill.supportedCountries` is `US` only, it depends on the region lookup on whether it shows up in the UI or not.
No outgoing connections were allowed for FF in my testing but I was able to make the UI setting show up by setting `browser.search.region` to `US` and restart FF

PR #968, opened 2020-07-07
dnsResolveSingleWordsAfterSearch

Thorin-Oakenpants:
I almost feel as if everyone (myself included) forgot the pref name .. **AfterSearch** and we don't even need to ~~add a note at all~~ mention 0801 at all: it's not about `keyword.enabled` .. it's about searching


earthlng:
> it's not about `keyword.enabled`

dnsResolveSingleWordsAfterSearch does exactly nothing when keyword.enabled is false but okay, whatever.

Thorin-Oakenpants:
that's the point. The pref explicitly says it only applies after searching: so there's no need to mention anything else: e.g:
- cookie lifetimePolicy only applies if you allow a cookie
- startup caching of AS only affects those who use AS

If you've blocked searching from the urlbar, then the pref clearly does not apply. How you blocked searching is irrelevant

PR #997, opened 2020-08-20
improve readability, remove legacy arguments

h88e22dgpeps56sg:
Hey,
not long ago I forked your updater.sh to create a version for a user.js of another website. That's why I had a thorough look on the whole script and found several ways to improve it. Now I want to give something back and implement my ideas which hopefully you'll find appealing. So brace yourselves, there's a lot more to come.

First one:
* remove lots of unnecessary echo commands which don't help at all
* replace `\t`-alignments with spaces so that it is agnostic of the user's settings and always looks the same. this also improves the readability of the code
* replace some small if-statements (especially one-liners) by `[ if ] && then || else`
* those legacy arguments where introduced two years ago. that's more than enough to safely remove them now

### DIFF
--- a/updater.sh
+++ b/updater.sh
@@ -11,7 +11,7 @@
 readonly CURRDIR=$(pwd)
 
 sfp=$(readlink -f "${BASH_SOURCE[0]}" 2>/dev/null || greadlink -f "${BASH_SOURCE[0]}" 2>/dev/null)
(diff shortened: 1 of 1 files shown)


atomGit:
i might also suggest printing the current Firefox version in the into text and warning if the new user.js version doesn't match the major Firefox version the user is currently running

h88e22dgpeps56sg:
Thanks so far. Just brought another one on the way. #1000

PR #998, opened 2020-08-23
1409: RFP changes in FF81+

Thorin-Oakenpants:
I almost feel as if this should move to the 4600's: introduced in FF52 it would mean a complete nudging of all the numbers in that section

### DIFF
--- a/user.js
+++ b/user.js
@@ -809,7 +809,8 @@ user_pref("gfx.font_rendering.opentype_svg.enabled", false);
 user_pref("gfx.font_rendering.graphite.enabled", false);
 /* 1409: limit system font exposure to a whitelist [FF52+] [RESTART]
  * If the whitelist is empty, then whitelisting is considered disabled and all fonts are allowed
- * [NOTE] RFP in FF80+ restricts the whitelist to bundled and "Base Fonts" (see 4618)
+ * [NOTE] in FF80 RFP restricts the whitelist to bundled and "Base Fonts"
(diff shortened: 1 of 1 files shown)


earthlng:
@Thorin-Oakenpants can you ask your fonts contact at mozilla if he knows why the font-visibility doesn't work with sandboxie, please?

Setting the whitelist works but the visibility pref doesn't. It should be mostly a font reduction based on hardcoded lists so I don't understand why that shouldn't work with sandboxie. Maybe/hopefully it's an easy fix that they'd be willing to implement.
Implementing a fix could also potentially prevent future problems w/o sandboxie when they crank up FF's own sandbox level.

Thorin-Oakenpants:
@earthlng - he'll just say to log an issue. I lack precise info you have (software/os details) and STR - might be easier if you log it

- Product: Core 
- Component: Layout: Text and Fonts
- add a needinfo for `jfkthame`

Thorin-Oakenpants:
@earthlng ... do you still want me to poke Jonathan Kew?

earthlng:
yes, I'd appreciate it, thanks. I thought since he already knows you he's more likely to look into it if you poke him instead of me.

I imagine he probably knows from the top of his head which parts of the code go beyond the hardcoded list lookup and could be responsible for the breakage in sandboxie. But if he needs more infos or STR I can provide that too.

Thorin-Oakenpants:
OK, I sent you an email

earthlng:
Thanks, I'll wait until he responds

PR #1000, opened 2020-08-23
Updater.sh rework 2

h88e22dgpeps56sg:
##### DOWNLOAD_METHOD
* merge two declarations of download command into one
* add max-redirects to prevent redirect loops
* add silent options (prettier than `> /dev/null`)
##### download_file()
* shorten `download_file ()`
* exit instead of continue with a "non-file", which can lead to unwanted behavior
##### open_file()
* open_file(): pipe instead of nested sub-shell
##### readIniFile()
* readIniFile(): `grep -c` equals `grep | wc -l`
* make output prettier
* work with variable instead of temporary file
* update `PROFILE_PATH` more directly

### DIFF
--- a/updater.sh
+++ b/updater.sh
@@ -238,7 +238,7 @@ add_override () {
     do
       add_override "$f"
     done
-    IFS=$SAVEIFS # restore $IFS
+    IFS=$FSAVEIFS # restore $IFS
   else
     echo -e "${ORANGE}Warning: Could not find override file:${NC} ${input}"
   fi
--- a/updater.sh
+++ b/updater.sh
@@ -122,28 +122,36 @@ open_file () { #expects one argument: file_path
 
 readIniFile () { # expects one argument: absolute path of profiles.ini
   declare -r inifile="$1"
+  declare -r tfile=$(mktemp)
 
-  # tempIni will contain: [ProfileX], Name=, IsRelative= and Path= of the only (if) or the selected (else) profile
-  if [ "$(grep -c '^\[Profile' "${inifile}")" == "1" ]; then ### only 1 profile found
-    tempIni="$(grep '^\[Profile' -A 4 "${inifile}")"
+  if [ $(grep '^\[Profile' "$inifile" | wc -l) == "1" ]; then ### only 1 profile found
+    grep '^\[Profile' -A 4 "$inifile" | grep -v '^\[Profile' > $tfile
   else
-    echo 'Profiles found:'
-    echo '––––––––––––––––––––––––––––––'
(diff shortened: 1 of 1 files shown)


h88e22dgpeps56sg:
Sorry, I got a little carried away. Next time, I will split it into more commits.

earthlng:
Hi again,

I like most of it but am not too keen on changing readIniFile() so radically. Also looks like `profpath` is undefined.

re: `open_file()` - are you sure `cut` is always available on all Nixes **and** all Mac's ??

h88e22dgpeps56sg:
I could split the `download_file` stuff into 2 or more PRs, so it is not too much at once.

Fixed `profpath`.

Yes, `cut` is a POSIX command and therefore included in every *nix OS and MacOS is unix based, too:
[Wikipedia](https://en.wikipedia.org/wiki/List_of_Unix_commands) says: cut – Mandatory
[An A-Z Index of the Apple macOS command line (macOS bash)](https://ss64.com/osx/) lists: cut

earthlng:
Ok, thanks for the info about `cut`. `download_file` is not the issue, it's readIniFile(). I spent quite some time writing and testing that function and I don't have the time or willingness to redo that at the moment. If you have more changes you want to make, let's keep the readIniFile() mods for last so that I can look at it when I have some time and you don't have to wait that long for more PR's.
The only problem I see with the new `download_file` is that if we end up exiting at `download_file` in `update_userjs` we don't restore the original working directory.

ps there's also a leftover `legacy_lc` [line](https://github.com/ghacksuserjs/ghacks-user.js/blob/master/updater.sh#L336) that we can remove

h88e22dgpeps56sg:
Sorry, my initial comment had `download_file()` but I edited it to `readIniFile()`.

* reverted `readIniFile()` back to normal.
* removed leftover line
* changed the exits to returns, so that the updater will cd back at the end
* added error message in case of a download failure (that was a little tricky in combination with `declare -r`)

earthlng:
Thanks mate. 
- How about we return empty string (`echo ''`) and check `newfile` directly so we don't need `ret`? That's cleaner IMO.
- ideally add a similar check to the other 2 places where we call download_file()
- There's an old typo/bug that I forgot to fix the last time: [FSAVEIFS](https://github.com/ghacksuserjs/ghacks-user.js/blob/master/updater.sh#L241) vs [SAVEIFS](https://github.com/ghacksuserjs/ghacks-user.js/blob/master/updater.sh#L248)
- if you want, add yourself to the [list of contributors](https://github.com/ghacksuserjs/ghacks-user.js/blob/master/updater.sh#L7)
- with all that in place, we can rollout a new version so please increase the version number

h88e22dgpeps56sg:
All done. :)

earthlng:
Perfect! Thank you

PR #1001, opened 2020-08-25
https-only mode

Thorin-Oakenpants:


earthlng:
`When "https_only_mode" (all) is true, "https_only_mode_pbm" (Private Browsing Mode) is ignored`

in my lingo "Private Browsing Mode" means `browser.privatebrowsing.autostart=true`. "(all)" isn't necessarily very clear either.

How about something like

`When "https_only_mode" (normal + private windows) is true, "https_only_mode_pbm" (private windows only) is ignored`

Thorin-Oakenpants:
true. I remember when DNT had the same setup, and we used "windows" not "mode" - I just auto went PBM .. M = mode

Thorin-Oakenpants:
commit if you're happy

Issue #1004, opened 2020-08-28
changelog: v80-beta


Thorin-Oakenpants:
* date: 28-Aug-2020
* foreword: These are all the changes since the last changelog ([v79-beta](https://github.com/arkenfox/user.js/issues/991)).

[FF80 release notes](https://www.mozilla.org/firefox/80.0/releasenotes/)
[FF80 for developers](https://developer.mozilla.org/docs/Mozilla/Firefox/Releases/80)
[FF80 compatibility](https://www.fxsitecompat.dev/en-CA/versions/80/)
[FF80 security advisories](https://www.mozilla.org/security/advisories/mfsa2020-36/)

---

#### changelog: [[all changes](https://github.com/arkenfox/user.js/compare/v79.0-beta...v80.0-beta)]
* all user.js **updates for Firefox v79** are detailed in the [ToDo: diffs FF79-FF80](https://github.com/arkenfox/user.js/issues/994) issue
  * includes links to the commits made for each pref, links to bugzilla tickets, our discussions etc.
* for all the rest see the full list of pref changes below

---

#### all pref changes:

* changed in 80beta, see #992
```js
user_pref("browser.newtabpage.activity-stream.asrouter.providers.snippets", "{}"); // was ""
```

* new in 80beta but commented out by default
```js
   // user_pref("dom.security.https_only_mode_pbm", true);
```

* moved to `9999: DEPRECATED / REMOVED`
   - see #979
   - the code was [removed in FF76](https://bugzilla.mozilla.org/show_bug.cgi?id=1507375) (but they forgot to remove the `dead prefs` which they have now done [in FF82](https://bugzilla.mozilla.org/show_bug.cgi?id=1660524))
   - so note: they are under the deprecated FF76 section
```js
user_pref("dom.disable_window_open_feature.close", true);
user_pref("dom.disable_window_open_feature.location", true);
user_pref("dom.disable_window_open_feature.menubar", true);
user_pref("dom.disable_window_open_feature.minimizable", true);
user_pref("dom.disable_window_open_feature.personalbar", true);
user_pref("dom.disable_window_open_feature.resizable", true);
user_pref("dom.disable_window_open_feature.status", true);
user_pref("dom.disable_window_open_feature.titlebar", true);
user_pref("dom.disable_window_open_feature.toolbar", true);
```

---

Any and all help, suggestions, recommendations, links, tips and tricks, questions, thank you's or what have you are welcome - signup/login and start typing - thanks

[...]

PR #1006, opened 2020-09-03
restart info

Thorin-Oakenpants:
- font whitelist doesn't require a restart to affect web content, which is all that matters. Do note though, that the chrome may be affected after restarted (who cares) - edit: although that **may** affect widgets
- prefersReducedMotion - see #1004 and @rusty-snake : I lack the ability to understand or confirm WTF is going on, so over to you


earthlng:
prefersReducedMotion - see https://github.com/ghacksuserjs/ghacks-user.js/pull/962#issuecomment-650553415 and followup comments

Thorin-Oakenpants:
so does that mean it requires a restart? just a yes/no is all I need (I am too zonked and swamped to read or test, sorry)

PR #1007, opened 2020-09-03
miscellaneous

Thorin-Oakenpants:
- less active prefs
   - now that ESR68 is EOL, at least a whopping two (0602, 1273)
   - also I don't know when the default changed - another whopping whole one (1240)
   - and where we do enforce/reset a pref to default, lets say that
   - this is not a definitive list, sing out if there is anything else
- IPv6 info
   - especially for Iron Heart who likes to claim that this pref breaks 5% of sites
- cleanup of settings tags now we only care abut ESR78+

### DIFF
--- a/user.js
+++ b/user.js
@@ -1689,7 +1689,6 @@ user_pref("_user.js.parrot", "9999 syntax error: the parrot's deprecated!");
 // 0212: enforce fallback text encoding to match en-US
    // When the content or server doesn't declare a charset the browser will
    // fallback to the "Current locale" based on your application language
-   // [SETTING] General>Language and Appearance>Fonts and Colors>Advanced>Text Encoding for Legacy Content (FF72-)
    // [TEST] https://hsivonen.com/test/moz/check-charset.htm
    // [1] https://gitlab.torproject.org/tpo/applications/tor-browser/-/issues/20025
    // [-] https://bugzilla.mozilla.org/1603712
--- a/user.js
+++ b/user.js
@@ -206,8 +206,7 @@ user_pref("app.update.auto", false);
  * when false, extension detail tabs will have no description ***/
    // user_pref("extensions.getAddons.cache.enabled", false);
 /* 0308: disable search engine updates (e.g. OpenSearch)
- * [NOTE] This does not affect Mozilla's built-in or Web Extension search engines
- * [SETTING] General>Firefox Updates>Automatically update search engines (FF72-) ***/
+ * [NOTE] This does not affect Mozilla's built-in or Web Extension search engines ***/
 user_pref("browser.search.update", false);
 /* 0309: disable sending Flash crash reports ***/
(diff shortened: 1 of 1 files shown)


theltalpha:
>     * less active prefs
>       
>       * now that ESR68 is EOL, at least a whopping two (0602, 1273)
>       * also I don't know when the default changed - another whopping whole one (1240)
>       * and where we do enforce/reset a pref to default, lets say that
>       * this is not a definitive list, sing out if there is anything else
> 
>     * IPv6 info
>       
>       * especially for Iron Heart who likes to claim that this pref breaks 5% of sites
> 
>     * cleanup of settings tags now we only care abut ESR78+

Hallo!

ESR68 is still supported by Mozilla, currently as version 68.12. This is the version which for example is used in Debian GNU/Linux. Support ceases at 22 September 2020, so the corresponding ESR settings for the ESR68.x branch in user.js should be kept until that date.

Thorin-Oakenpants:
This is an alpha release based on 81

theltalpha:
> @theltalpha This is an alpha version based on 81

You are so completely right, I am sorry!

earthlng:
I don't like making the prefs inactive. We want these to be the value we set even if mozilla change them back for whatever reason. 1273 is just a visual thing so I don't care too much about that but the other 2 should stay active. For 1240 we should also change the title to "enforce" or something like that IMO.

The 0517 SETTING does not exist in the UI due the prefs we have in 0517. Not sure if we should remove that line completely (?).

The rest LGTM

Thorin-Oakenpants:
> The 0517 SETTING does not exist in the UI due the prefs we have in 0517

Oh yeah, that's right. I'm only dropping the `[74+]` notation anyway

> I don't like making the prefs inactive. We want these to be the value we set even if mozilla change them back for whatever reason

They're not likely to ever change (and we have diffs per release) `0602` prefetch `1240` mixed active `1273` its just a visual thing. Prefetch might be the only one of those three, if say someone decided that they wanted to push loading/speed times and now HTTPS is the main driver. Note it was only flipped in 70. I just don't think it will ever happen, especially given just how deep privacy and tracking (and FPing) comes up in consideration in tickets etc (even from the font people!!)

But OK. Yup enforce for 1240, and 1240+0602 can stay active. I'm just always aware that if we need to keep cleaning out dead wood and trimming back active prefs when possible, and ESR-EOL is a good time.

Thorin-Oakenpants:
merge if you're happy

Thorin-Oakenpants:
Wait: how does this sound for the forceMediaCache

```js
/* 1007: disable media cache from writing to disk in Private Browsing
 * [NOTE] ESR78 playback might break on subsequent loading (1650281)  // <-- this is the new line
 * [NOTE] MSE (Media Source Extensions) are already stored in-memory in PB ***/
user_pref("browser.privatebrowsing.forceMediaMemoryCache", true); // [FF75+]
user_pref("media.memory_cache_max_size", 65536);
```

Issue #1025, opened 2020-10-01
RFP and document fonts


syncng:
This might be an incredibly dumb question...

Currently, user.js is preventing the browser from using any document provided fonts.
Now that RFP covers font visibility, can document fonts still be considered a "high entropy fingerprinting vector" when RFP is active or are document fonts and font visibility completely unrelated?

KOLANICH:
Related. When one supplies document fonts files, he eliminates variability in the files, this way he can probe the variation among environments (libs, drivers, OSes, rendering engines, browser settings) separately.

smokemyshoes:
I'm also interested in this topic. If one could share some knowledge on what is the best settings/approach to fonts now with the new font-features in RFP, it would be most appreciated. TIA

KOLANICH:
IDK what is better for now: use the most mitigated browser or not to use the Web at all. Remember - fingerprinting will never be properly mitigated, because it requires consensus of all stakeholders, but the largest stakeholder has conflict of interests.

Thorin-Oakenpants:
It's a little messy, RFP doesn't kick in with font visibility limitations until FF80, so how do we handle ESR users (I need to think about it more).... that said ESR users technically should be using the 78 release and not come back for another year: but that doesn't stop them updating with current and flipping the ESR section: because we provide that feature

... **BUT** .. I wanted to move document fonts into the RFP alts section (so it's off by default since that section is commented out) and said as much to earthlng (I think it was in a PR), but at the time we didn't and of course RFP protecting fonts only just landed in 80

so it's just a matter of working out how to do it so it works for both FF80+ and ESR78

---

As I've always said, if you mess with RFP then you lose "immunity" - [read this](https://github.com/arkenfox/user.js/wiki/4.1-Extensions#small_orange_diamond-%EF%B8%8F-anti-fingerprinting-extensions-fk-no): of course document fonts should be inactive if you're using RFP - live by the sword, die by the sword

.. as much as it seems that you're giving away some free data on fonts - you're not: you can't hide your OS and the lists are quite limiting. There will be some entropy within each OS, but that's far better than being one of a very few RFP users with one or three fonts detected (depending on the script)

[...]

PR #1028, opened 2020-10-04
fixup font prefs vs RFP, see #1025

Thorin-Oakenpants:
- make 1401 inactive: it affects RFP's FPing
- remove old warning/setup-web: we do not care about documenting breakage or FPing risks when we have a warning and they are inactive. If someone uses them, that's on them
- new warnings


Thorin-Oakenpants:
bump @earthlng

earthlng (approved):

PR #1040, opened 2020-10-15
misc

Thorin-Oakenpants:
- cleanup of old release notation in comments: e.g. if it's not applicable to ESR78+
- same with default version info
- simplify and save bytes on section 4700
- update 4500 header
  - and unify the message about using extensions as counterproductive
- letterboxing
   - provide info on stepped ranged (and drop crap about FF67)
   - don't judge users who dislike seeing margins (I don't like them either, but I force my window to exact dimensions and stay there)
- screenshots uploading was disabled in FF67+ : [67 release notes](https://www.mozilla.org/en-US/firefox/67.0/releasenotes/)
   - the pref is still there (default false) but so far I'm 99% sure this pref now does anything
   - I will add it to the scatchpad script if this change sticks

### DIFF
--- a/scratchpad-scripts/arkenfox-clear-removed.js
+++ b/scratchpad-scripts/arkenfox-clear-removed.js
@@ -1,7 +1,7 @@
 /***
  This will reset the preferences that have been removed completely from the arkenfox user.js.
 
- Last updated: 2-Oct-2020
+ Last updated: 18-Oct-2020
 
  For instructions see:
  https://github.com/arkenfox/user.js/wiki/3.1-Resetting-Inactive-Prefs-[Scripts]
@@ -225,6 +225,10 @@
     'browser.urlbar.usepreloadedtopurls.enabled',
     /* 80 */
     'dom.IntersectionObserver.enabled',
+    /* 82-beta */
+    'extensions.screenshots.upload-disabled',
+    'security.ssl3.dhe_rsa_aes_128_sha',
+    'security.ssl3.dhe_rsa_aes_256_sha',
     /* reset parrot: check your open about:config after running the script */
     '_user.js.parrot'
   ]
--- a/user.js
+++ b/user.js
@@ -68,7 +68,7 @@
   4000: FPI (FIRST PARTY ISOLATION)
   4500: RFP (RESIST FINGERPRINTING)
   4600: RFP ALTERNATIVES
-  4700: RFP ALTERNATIVES (NAVIGATOR / USER AGENT (UA) SPOOFING)
+  4700: RFP ALTERNATIVES (USER AGENT SPOOFING)
   5000: PERSONAL
   9999: DEPRECATED / REMOVED / LEGACY / RENAMED
 
@@ -340,10 +340,8 @@ user_pref("extensions.systemAddon.update.url", ""); // [FF44+]
 /* 0506: disable PingCentre telemetry (used in several System Add-ons) [FF57+]
  * Currently blocked by 'datareporting.healthreport.uploadEnabled' (see 0340) ***/
 user_pref("browser.ping-centre.telemetry", false);
-/* 0515: disable Screenshots
- * alternatively in FF60+, disable uploading to the Screenshots server ***/
+/* 0515: disable Screenshots ***/
    // user_pref("extensions.screenshots.disabled", true); // [FF55+]
-   // user_pref("extensions.screenshots.upload-disabled", true); // [FF60+]
 /* 0517: disable Form Autofill
  * [NOTE] Stored data is NOT secure (uses a JSON file)
  * [NOTE] Heuristics controls Form Autofill on forms without @autocomplete attributes
@@ -365,7 +363,7 @@ user_pref("network.prefetch-next", false);
 /* 0602: disable DNS prefetching
  * [1] https://developer.mozilla.org/docs/Web/HTTP/Headers/X-DNS-Prefetch-Control ***/
 user_pref("network.dns.disablePrefetch", true);
-user_pref("network.dns.disablePrefetchFromHTTPS", true); // [DEFAULT: true FF70+]
+user_pref("network.dns.disablePrefetchFromHTTPS", true); // [DEFAULT: true]
 /* 0603: disable predictor / prefetching ***/
 user_pref("network.predictor.enabled", false);
 user_pref("network.predictor.enable-prefetch", false); // [FF48+]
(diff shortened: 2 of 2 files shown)


earthlng:
- didn't we want to remove 1263 with the latest ESR changes? ie now
- `1273` - I'd also remove the FF59+ and FF60+ tags behind the 2 prefs
- `2701` - I'd remove the version infos from the brackets and just put `(default)`
- `4701` - the item title is a bit weird IMO. Makes it look like prefs: `navigator.*.overrides` - I'd prefer something like `navigator.* overrides` or `navigator.* DOM object overrides` or something like that

👍 for the rest

Thorin-Oakenpants:
- don't know what I was thinking when I typed `navigator.*` - that's not even the nomenclature of the prefs
- the 59/60 info on 1273 is consistent with every other pref: it's when it was added to the code. We have them going back as far as FF22: so we should either rip them all out except those since 79+ (or maybe 78+): in total there are 118 instances of `[FF`

Thorin-Oakenpants:
we could also do something about the list in RFP: e.g.

we don't need all this: besides that though I don't see too much more we could simplify: any ideas?
```
 ** 1333651 & 1383495 & 1396468 - spoof User Agent & Navigator API (see section 4700) (FF56+)
      FF56: Version: rounded down to the nearest multiple of 10
      FF57: Version: match current ESR (1393283, 1418672, 1418162, 1511763)
      FF59: OS: Windows, OSX, Android, or Linux (to reduce breakage) (1404608)
      FF66: OS: HTTP Headers reduced to Windows or Android (1509829)
      FF68: OS: updated to Windows 10, OS 10.14, and Android 8.1 (1511434)
      FF78: OS: updated to OS 10.15 and Android 9.0 (1635011)
```

Thorin-Oakenpants:
> 2701 - I'd remove the version infos from the brackets and just put (default)

Just remember that FF69+ info for when they change the default to dFPI :)

Thorin-Oakenpants:
removed uMatrix info: we can deal with the wiki page later: I'm gong to still use it for 6 months to a year and keep an eye on it. I like the worker scope: so it's a wait and see game for some feature parity

I also have quite the new and slimmed down design/RFP section: but maybe we could do that in another PR?

Push this if you're happy

earthlng:
- I'd keep the uMatrix info in the workers section
- 1273: if you want to keep the version tags then I wouldn't remove the FF70+ in the DEFAULT tag

Thorin-Oakenpants:
We shouldn't be encouraging an unmaintained extension, so I really don't want to add it back: even if it currently works and you and I know the risks etc.

---

Adding the default version info back in `1273`. OK, I thought this was because a numbered item has the `[FFver+]` on the pref lines versus on the numbered item line

My logic has always been: `[FFver+]` is when the pref was added to Firefox, `[DEFAULT: x]` is it's current state (but we are nice enough to add notation when it flipped during ESR cycles: which also gives us info on when we could possibly make things inactive, or remove them, etc). I never wanted to maintain a history of when prefs flipped

That said, there are 22 defaults listed: I'm jumping between them now. Almost all of those prefs don't have a `[FFver+]` notation, so this is probably a bit of a new case. I can certainly see someone being confused

here they are
```js
/* 1273: display "insecure" icon and "Not Secure" text on HTTP sites ***/
   // user_pref("security.insecure_connection_icon.enabled", true); // [FF59+] [DEFAULT: true]
user_pref("security.insecure_connection_text.enabled", true); // [FF60+]

/* 1604: CROSS ORIGIN: control the amount of information to send [FF52+] ***/
user_pref("network.http.referer.XOriginTrimmingPolicy", 0); // [DEFAULT: 0]

/* 1606: ALL: set the default Referrer Policy [FF59+] ***/
   // user_pref("network.http.referer.defaultPolicy", 3); // [DEFAULT: 3]
   // user_pref("network.http.referer.defaultPolicy.pbmode", 2); // [DEFAULT: 2]

/* 2429: enable (limited but sufficient) window.opener protection [FF65+] ***/
user_pref("dom.targetBlankNoOpener.enabled", true); // [DEFAULT: true FF79+] // <-- because it differs from ESR78

/* 4002: enforce FPI restriction for window.opener [FF54+] ***/
   // user_pref("privacy.firstparty.isolate.restrict_opener_access", true); // [DEFAULT: true]
```

So 1604, 1606 and 4002 are probably correct that these are the defaults since they were added: especially given that they have been added since we came to github, and between us we're pretty careful: but that's not the point: they weren't added for that purpose. I might come across a ticket where TLS prefs were added and append that version to `1202`: in other words this (FFver vs default info) would now become yet another thing to think of and maintain

How do people interpret prefs with no version information? e.g TLS versions: are they to assume it's always been 3 + 4? There's no consistency of logic here.

What happens in this made up case?

```js
/* 4002: enforce whizzy-do-dacky [FF82+] ***/
   // user_pref("dom.whizzy_do_dacky.enabled", true); // [DEFAULT: false in FF82; true in FF83; false in FF84+]
```

We have three options imo
- one: do nothing
- two: always make sure when we add FFver info to prefs, that we keep in mind, and track it's entire default status if we ever add a default status. We should probably make all default status info have version numbers, reduced the last change at a minimum e.g. "since X"
- three: at the top under tags: add that `[FFver+]` : when the pref was added, and `[DEFAULT` refers to the current release

@earthlng (so for the bit of a rambly answer: I need more coffee, and beer .. beercoffee)

Thorin-Oakenpants:
@earthlng https://www.fxsitecompat.dev/ is dead: we have two links
```
/* 0708: disable FTP [FF60+]
 * [1] https://www.fxsitecompat.dev/en-CA/docs/2020/ftp-support-will-be-removed/

/* 0912: limit (or disable) HTTP authentication credentials dialogs triggered by sub-resources [FF41+]
 * [1] https://www.fxsitecompat.com/en-CA/docs/2015/http-auth-dialog-can-no-longer-be-triggered-by-cross-origin-resources/ 

```

We could replace 0708 (it's not listed as a meta but it basically is), or we could just not have a link
```
 * [1] https://bugzilla.mozilla.org/1574475 [META]
 * [1] https://bugzilla.mozilla.org/1574475 [META: intent to remove FTP support]
```

not sure about the sub-resources one

earthlng:
damn! I can understand that they don't have the resources anymore to keep that site updated but they could've at least kept the old content intact.
0708 can probably do without a link so there's not much lost there but I doubt we'll find anything comparable to the short and concise info that was behind that link for 0912. Just remove it, I guess.

I'm really starting to get worried now. The house is slowly crumbling while the greedy bitch Baker is milking it for a couple more millions each year until it all collapses. It's really fucking sad

Thorin-Oakenpants:
I also don't get why they didn't leave the resources up and just apply some sticky notice. But TBH, there's not that much to keep track of, and it's all in the developer notes anyway ... fuuuuck .. touch wood ... here's hoping dev-docs isn't killed off

Thorin-Oakenpants:
is this any good:
https://web.archive.org/web/20181123134351/https://www.fxsitecompat.com/en-CA/docs/2015/http-auth-dialog-can-no-longer-be-triggered-by-cross-origin-resources/

Thorin-Oakenpants:
I want to merge this so we start clean. And then we can do a separate PR on adding an ESR section so I can see how it looks/fits

merge if you're happy, otherwise I'll probably pull the trigger after I've had some sleep

PR #1052, opened 2020-10-23
Update user.js

nikky696:


rusty-snake:
Any explanation why enabling offline cache?

Thorin-Oakenpants:
This is the AppCache. It is basically dead: I can't point to fx compat since that site is gone, but it has been slated for deprecation for about a decade, and usage is practically nil

There's really no need for it in this day and age. They put off disabling the appCache `storage` (not the `API`) in release due to covid-19 and possible issues with antique old-timey govt websites. Meanwhile, it's been ignored when it comes to next gen local storage and quota management and sanitizing fixes (it's not worth coding for) - I'm not even sure if FPI covers it

Old-timey proof that everyone wants this gone
- In [1204581](https://bugzilla.mozilla.org/show_bug.cgi?id=1204581) ... way back in FF44, they started warnings in the console for web devs
- In [1237782](https://bugzilla.mozilla.org/show_bug.cgi?id=1237782) they finally got around to disabling it in Nightly way back in FF71, just to show you how old this tech is

Here's the state of things
- [1584984](https://bugzilla.mozilla.org/show_bug.cgi?id=1584984) meta (a newer relevant one)
- [1619673](https://bugzilla.mozilla.org/show_bug.cgi?id=1619673) disable appCache in release .. FF84+ .. landed

At which point (FF84), I'll need to check if they backport this to ESR (which AFAIK will not happen), and also decide what to do with the pref.  See comment below for what they did back in FF71 nightly, which still holds true today. The API is still enabled (and tests should fail)

[1237782 comment71](https://bugzilla.mozilla.org/show_bug.cgi?id=1237782#c71)
> Technically we have also just removed storage for now, not support for the API

So come FF84 (assuming it rides the train to stable), then we have another ESR vs release issue: i.e we should make the pref inactive (to be like all the other Firefoxes) and we're still protected from the appCache storing data, but ESR users would (if they reset prefs) be left in the lurch

That was your ancient history lesson for the day. Class, dismissed!

PR #1054, opened 2020-10-25
2422 WASM, add reason for disabling, fixes #1037

### LINKED ISSUE #1037: Unable to access hanime.tv [solved: wasm]
Firefox 81.0.4
Using latest user.js with recomended setup.

Accessing the degenerate site hanime.tv results in "Your browser is too old", which is odd. Using safe-mode without extensions results in the same issue. Using 2nd clean profile with no extension and no custom `user.js` works just fine.

I tried going through most obvious prefs in `user.js` disabling them one-by-one but after 20/30 or so I gave up, there were no results and testing all 100+ is just an absolute chore.

if anyone has any suggestions that'd be appreciated

Thorin-Oakenpants:


Thorin-Oakenpants:
About the only thing missing is a `[setup` tag. But TBH, I think it's OK, unless we can trim it down a bit. Merge if you're happy

PR #1067, opened 2020-11-21
add FPI scheme, closes #1066

### LINKED ISSUE #1066: ToDo: add FPI "use_site" (inactive)
`privacy.firstparty.isolate.use_site`

- added in FF78 - [1637516](https://bugzilla.mozilla.org/show_bug.cgi?id=1637516)
- Would be inactive as it's experimental and indeed breaks HOM exceptions as one example
- We can then refer to it in our HOM stuff in #1047 

more to follow

Thorin-Oakenpants:


earthlng:
the title for 1637516 is "Consider changing the key of StoragePrincipal from registrable domain (eTLD+1) to site"

ie **from** eTLD+1 **to** site

IDK what exactly eTLD+1 means but you have it the other way around in the description.

the use_site pref adds the scheme (http, https, (etc?)) and port (if available) to the OA key.

I wouldn't be surprised if this (silently) breaks various "older" FPI features because most of the tests they wrote for FPI over the years have most likely not been adjusted to check if those features still work when this new pref is set.


ps this will be my last comment until you bring back the link - you know what I mean.

crssi:
The whole site name is known as the eTLD+1. For example, given a URL of `https://my-project.github.io` , the eTLD is `.github.io` and the eTLD+1 is `my-project.github.io`, which is considered a "site". In other words, the eTLD+1 is the effective TLD and the part of the domain just before it.

Source: https://web.dev/same-site-same-origin/

Issue #739, opened 2019-06-06
toolkit.telemetry.cachedClientID keeps repopulating since Firefox 67


roviavas:
On Firefox 66 and older versions, `toolkit.telemetry.cachedClientID` would not show up with telemetry disabled (by policies.json.).
However, after upgrading to 67, Firefox keeps setting `toolkit.telemetry.cachedClientID` to a UUID-like string. Resetting or setting it to an empty string does not help. It just repopulate the next day.
I have also tried modifying other telemetry-related entries, but cannot find the way to stop it.
Any idea?

claustromaniac:
Apparently, Firefox now resets the pref if the value is invalid.

https://dxr.mozilla.org/mozilla-central/source/toolkit/components/telemetry/app/ClientID.jsm#232-240

Thorin-Oakenpants:
If you're not sending any telemetry data back then it doesn't matter

claustromaniac:
> then it doesn't matter

Is there any point in keeping that pref around then? I'm asking more out of curiosity than anything else.

Thorin-Oakenpants:
Oh, I didn't realize it was in the user.js - got confused with some other `id` pref someone else mentioned several weeks ago

Thorin-Oakenpants:
FWIW, my `toolkit.telemetry.cachedClientID` is blank. Been on 67, now 67.0.1, for, IDK, 6 days or so. Loads of restarts.

Where is it getting set from? A system add-on?

claustromaniac:
> Where is it getting set from? A system add-on?

A function in the ClientID module itself is resetting the pref. Are there any errors in your console when you launch Firefox that may be related to that? Something in your setup must be preventing that code from running somehow... I can try to dig a bit more but it will take a while

Thorin-Oakenpants:
fuk, just did a big edit in my last post

Thorin-Oakenpants:
This is what Nightly does - it's a nilla profile, with just the user.js overwriting the pref
![console](https://user-images.githubusercontent.com/16656956/59054488-56364000-8883-11e9-9e7a-661233706427.png)

claustromaniac:
I think what's going on is simply that you already had that empty clientID cached, and the function was simply retrieving it from disk, see here

https://dxr.mozilla.org/mozilla-central/source/toolkit/components/telemetry/app/ClientID.jsm#220-229

[...]

PR #1112, opened 2021-02-06
New additions

the-weird-aquarian:
- added 4520, 4521: for better tracking protection against fingerprinting and cryptomining
- added 4522: use Mozilla's new built-in tracking protection to block tracking (i.e. Google Analytics) on privileged pages where add-ons that usually do that are disabled
- added 0392: delete cached client ID

edited by pants to remove `#` giant sized comments

### DIFF
--- a/user.js
+++ b/user.js
@@ -67,7 +67,7 @@
   2700: PERSISTENT STORAGE
   2800: SHUTDOWN
   4000: FPI (FIRST PARTY ISOLATION)
-  4500: RFP (RESIST FINGERPRINTING)
+  4500: RFP (RESIST FINGERPRINTING) & RCM (RESIST CRYPTOMINING)
   4600: RFP ALTERNATIVES
   4700: RFP ALTERNATIVES (USER AGENT SPOOFING)
   5000: PERSONAL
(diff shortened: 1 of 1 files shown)


rusty-snake:
You should break long descriptions into several lines, so they are better readable and consisted to the rest of the user.js.

> added 0392: delete cached client ID

What's the issues with a ID on your local machine?

And to which section should belongs this? It's floating somewhere in between in the user.js.

> added 4520, 4521: for better tracking protection against fingerprinting and cryptomining

Are true by default and have nothing to do with RFP. There's no reason to enforce the default for them (mozilla will very likely never flip them back).

rusty-snake:
> added 4520

```js
/* 4520: disable chrome animations [FF77+] [RESTART]
 * [NOTE] pref added in FF63, but applied to chrome in FF77. RFP spoofs this for web content ***/
user_pref("ui.prefersReducedMotion", 1); // [HIDDEN PREF]
```

Thorin-Oakenpants:
> 4500: RFP (RESIST FINGERPRINTING) & RCM (RESIST CRYPTOMINING)

RFP is a term .. RCM is not. RFP is specifically for countermeasures against a FP script when it runs. It has nothing to do with Enhanced Tracking Protection which is bloacklists for fingerprinters and cryptomining scripts

> user_pref("toolkit.telemetry.cachedClientID", "")

this is set at runtime, see #739 : also, what is the issue with it? explain why it is a privacy threat

> duplicated descriptions

why?

> user_pref("privacy.trackingprotection.fingerprinting.enabled", true);
> user_pref("privacy.trackingprotection.cryptomining.enabled", true);

nothing to do with RFP. both of them have been default true since at least FF78

> /* 4522: This is Mozilla's new built-in tracking protection. One of it's benefits is to block tracking (i.e. Google Analytics) on privileged pages where add-ons that usually do that are disabled. ***/
> user_pref("privacy.trackingprotection.enabled", true);

We already allow addons to run on privileged pages, and would expect users to be using uBO. This has been brought up before by someone at PTIO, and IDK if it's worth the bother
- this is the option to change TP from "Only in private windows" to "In all windows"
- if Mozilla aren't ready to make that the default, then I'm quite happy to not flip it: users should be using uBO at a bare minimum

Issue #1014, opened 2020-09-12
migration notes


Thorin-Oakenpants:
Some items to address
- [x] fixup any links / references in user.js
- [x] license needs `Copyright (c) 2017 ghacksuserjs` edited
- [x] updater scripts will need to check for arkenfox, not ghacks
- [x] readme
- [x] rename scratchpad scripts
- [x] code cleanups - [three left](https://github.com/arkenfox/user.js/search?q=ghacks&type=code) including updater scripts
- [x] check the wiki (script names, references, links)
- [ ] travis? what happened to the green ticks on commits - they stopped 3 days before the migration. Note we still have `verified` notation

earthlng:
I updated the updater.bat to let it recognise both ghacks and arkenfox user.js files.
If/when you update the "ghacks" strings in updater.bat just make sure to keep [line 104](https://github.com/ghacksuserjs/ghacks-user.js/blob/master/updater.bat#L104) intact!
That's the only place where "ghacks" is used as more than just static text (or a link).

Any other changes of ghacks references in any of the files should be done **after** the migration.

re: Releases - the current ones should all keep the ghacks name because they all contain "ghacks" files.
But I think we can remove some of the older "earthlng" pre-releases, maybe up to and including `earthlng user.js v68.0-beta` ?

Thorin-Oakenpants:
Yup, wasn't going to action anything until after the migration. The only question is when do we do it. I'm super keen to pull the trigger right now, as many users will be visiting / checking things as they start using the ESR78 - i.e, it's a prime opportunity to break something and force them to look. The other argument to that is don't upset/break anything until after, so no ESR users are affected

👍 pull the trigger now
👎 wait til after the next release

Thorin-Oakenpants:
> or assign me :)

sounds like a plan

Ne3tCode:
https://github.com/arkenfox/user.js/wiki/4.1-Extensions/_compare/bfa7353%5E...bfa7353
Broken signs ✔ ❗️ ⚠️

Thorin-Oakenpants:
thanks - fixed. I hope the other pages are OK. I just edited the entire wiki by cutting, search+replace in notepad and pasting back. I guess some of the symbols didn't like it

Thorin-Oakenpants:
IDK about the last item

lhindir:
[...]

Issue #1009, opened 2020-09-04
this repo is moving: new org name suggestions appreciated


Thorin-Oakenpants:
Nah, you can't ask why, sorry - well, you can, but I will not answer that question.

It will however be staying at Github, and nothing (important) should be lost. Followers still follow, commit histories etc are kept. And AFAIK almost every link will redirect to the new one. There will be some teething issues such as the updater scripts - but we'll fix those. Note: I plan to move TZP first as a test

What I do want to know is what to use for a new organization name - maybe something that means something about what we do, but I'm open to all suggestions (something unique perhaps). Something short and memorable

TIA

rusty-snake:
Brainstorming:
- pants-user.js
- veil-fox
- better-fox (add-on)
- secfox (add-on)
- Chameleon (not really unique)
- ChameleonBrowser (There is a browser with that name fron 1999)
- ChameleonFox 
- NoMoreFP (Used has hashtag in social-media)
- JohnsFP
- Ants
- MakeMeUnique
- DoNotTrackMe

crssi:
thorins-user.js

rprimus:
Fri Sep  4 11:56:57 BST 2020

hard-user-js
hardened-fox

geeknik:
Hush Fox
Hushed Fox

These seem pretty unique and convey what you're trying to accomplish, making Firefox less chatty with the Internet. (;

ArchangeGabriel:
I like hardened-user.js.

Thorin-Oakenpants:
some notes:
- no to anything `hush` or similar- that's not what we're aiming to do: while I appreciate many like to make "firefox quiet" = there's nothing dodgy or bad about automatic connections (updating safe browsing, blocklists, revocations (CRLite), checking for portals, etc) - even telemetry is OK if you want to help out (there's no PII involved)
- not keen on anything `FP*` because that's a zero sum end game
   - or similar like `chameleon`

If I had to sum up what we do the main thing would be **unlinkability**: i.e not letting sites link your traffic (by helping minimizing or eliminating tracking techniques such as local persistent web storage, 3rd parties, FPI, FPing which most rely on canvas to achieve, referers, and so on). Hardening sounds like a security thing (and FF is pretty much secure out of the box). The other two things are anonymity (which you need Tor or a VPN) and privacy (which no one can give except you). Privacy is such an over-used term to describe everything these days, I don't want to use that

...the beatings will continue until the suggested new name improves...
[...]

PR #1128, opened 2021-03-01
v3.0 - improve readIniFile()

earthlng:
- grep -c equals grep | wc -l
- make output prettier
- work with variable instead of temporary file

### DIFF
--- a/updater.sh
+++ b/updater.sh
@@ -2,7 +2,7 @@
 
 ## arkenfox user.js updater for macOS and Linux
 
-## version: 2.9
+## version: 3.0
 ## Author: Pat Johnson (@overdodactyl)
 ## Additional contributors: @earthlng, @ema-pe, @claustromaniac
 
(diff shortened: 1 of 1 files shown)


earthlng:
This was part of an older [pull request](https://github.com/arkenfox/user.js/pull/1000) by @h88e22dgpeps56sg that we didn't finish back then.

Now was a good time to revisit that because we may want to reuse it (or do something similar) in the prefsCleaner.sh.

I already tested it under Linux but would like to get some feedback, improvements or more testing done (especially for Mac) by someone else before I commit it.
@overdodactyl, @ema-pe, @claustromaniac, @h88e22dgpeps56sg or anyone else - thanks!

ema-pe:
Hi, in the next days I can give some feedback. What happened to `ghacks-user.js`?

rusty-snake:
> What happened to ghacks-user.js?

Was renamed, #1014 and #1009.

h88e22dgpeps56sg:
Looks good to me. :) Just 2 things:
1.) line 127: `if [ $(grep -c '^\[Profile' "${inifile}") == "1" ]`
I'd rather use `if [ $(grep -c '^\[Profile' "${inifile}") -eq "1" ]`. grep outputs an int and -eq compares ints. `==` is a pattern matching which works but is equally as fast at best but probably slower.
2.) line 133: `echo;`
Replacing this with another line of `echo -e "––––––––––––––––––––––––––––––"` might look nice.

earthlng:
Hey, thanks for your feedback!

> Looks good to me. :)

I'd hope so :)

> grep outputs an int

good to know, thanks. Done

re: line 133 - yeah looks nice

I'll commit it now but more feedback, improvements, etc is still welcome.

PR #1136, opened 2021-03-09
misc

Thorin-Oakenpants:
- 0000: remove old XUL info, dropped in FF73+
- 0201: save 3 chars
- 0350: add default status for unsubmittedCheck
- 0351: change to enforce: has been default false going back to at least FF60, including current Beta/Dev/Nightly
   - along with 0602 `network.dns.disablePrefetchFromHTTPS` and 0603 `network.predictor.enable-prefetch`, I considered making them inactive, but decided it was good to leave them active for non-stable users just in case they get flipped
- 0515: add default status
- 0850c: remove info: out of date: doesn't work lilke that anymore and can't be assed figuring it out what with megabar and urlbar2 changes
- 0871: make inactive: default false since at least FF60
   - no need to enforce for non-stable in case it is flipped. It's a pretty minor shoulder-surfer privacy issue and the previews are small. If you're not sure what this pref does. On false you get one tab shown, on true you get as many as can fit across your screen. I squeezed in 15, and after that it became a list
- fixup `***/`

### DIFF
--- a/user.js
+++ b/user.js
@@ -83,9 +83,8 @@
 user_pref("_user.js.parrot", "START: Oh yes, the Norwegian Blue... what's wrong with it?");
 
 /* 0000: disable about:config warning
- * FF72 or lower: chrome://global/content/config.xul
  * FF73-86: chrome://global/content/config.xhtml ***/
-user_pref("general.warnOnAboutConfig", false); // XUL/XHTML version
+user_pref("general.warnOnAboutConfig", false); // XHTML version
 user_pref("browser.aboutConfig.showWarning", false); // HTML version [FF71+]
 
 /*** [SECTION 0100]: STARTUP ***/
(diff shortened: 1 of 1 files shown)


earthlng:
I don't like `[DEFAULT: false except Nightly]` and `[DEFAULT: false in stable]` because we only ever cared about stable.
Nightly is constantly changing and quite different pref-wise from stable. If we have even just one of these nightly tags in the user.js, it'll suggest that all the other DEFAULT tags and default value notes in some of the descriptions equally apply to all releases and that's almost certainly not true.

Thorin-Oakenpants:
We do mention nightly in at least one other pref, but in the description. True, if we start to differentiate these few diffs... 

So what do you suggest?
- leave the defaults in but remove the channel info
- add a line somewhere e.g. readme at the top, that default info applies to the release version: other channels may differ?
- remove the defaults

Also, if we don't really care (much? at all?) about non-stable, then we could make those two prefs  in 0602 and 0603 inactive - I doubt they'll ever get flipped

earthlng:
> Also, if we don't really care (much? at all?) about non-stable

with "because we only ever cared about stable" I meant specifically the DEFAULT tags.
fe 0518 is default true in beta, dev and nightly and I would've kept that pref active because of that. But it could still keep the Default false tag.

When I first started adding these Default tags I always checked that the pref matches the default in Windows, Linux and Mac, and also in the latest ESR but that was a long time ago. Not sure if that's still the case today for all those tags we added since then.

earthlng:
> Web Compat doesn't do anything unless the user decides to send something

sure, but how useful is it to send compat reports when we've changed 200+ prefs?

Thorin-Oakenpants:
totally useless, but that's not the reason for doing it: the reason is to remove "noise" from the user.js: I know it's not likely to ever confuse anyone, or break anything, but it is one more active pref to look at when troubleshooting. I'm just trying to be consistent

that said ... maybe we should enforce it disabled so that NO-ONE with 200 odd pref flips ends up wasting Mozilla's time. They need that time to create new buttons and shit - so yeah, good point. I'll change it

PR #1140, opened 2021-03-12
tweak defaults

Thorin-Oakenpants:
- don't differentiate between channels
- both can be made inactive
   - webcompat requires user action: and I don't see this as a bad thing to have in non-stable
   - unsubmitted crashReports on Nightly is probably already covered by killing the URL, so no big deal


Thorin-Oakenpants:
@earthlng .. just merge if you're happy

PR #1160, opened 2021-04-06
Fix typo in comment

adrian5:


Thorin-Oakenpants:
Note: "Temporary Containers" is singular. It is the name of an extension.

This is basically a question about definite-indefinite, countable-uncountable, and idiomatic usages. AFAIK, in most instances, software is not countable (as they are proper nouns: e.g. that's their name).

Here are some rules I like to follow
- You use an article with a singular noun
   - I see you have a dog
   - I see you have dog <-- weird, right?
- You omit an article with a proper noun
   - I see you use a uBlock Origin <-- weird, right?
   - I see you use uBlock Origin
- Articles are optional with plurals - not in the sense that you can include them or not on a whim, but that they change the meaning

However, we are not "defining" the extension, but the configuration of it. Articles are actually adjectives because they describe the nouns that they **_precede_** (yes, I know hardened is an adjective, not a noun: blame the English, not me: I do not have enough thumbs for the rules of thumb)
   - or you use a hardened uBlock Origin
   - or you use hardened uBlock Origin <-- weird, right?

In this case we are describing "hardened" (which is not even a term or preset in TC that I know of), not "Temporary Containers". So the rule about proper nouns doesn't apply. Note: I am not a grammar expert.

What was the reason for you thinking it was a typo, or grammatically incorrect? Was it because containers is plural?

adrian5:
Ah ok, my bad. One would say _“using TC”_ but due to the adjective, one has to say _“using a hardened TC“_. That makes sense.

Svallinn:
TIL Pants can write just a few paragraphs on pretty much anything, no matter how simple or complex

Thorin-Oakenpants:
Another thing to keep in mind, is that using (in)definite articles can signify specificity (which is what we want)

PR #1185, opened 2021-06-05
Allow silent execution

4n0nym0u53:
I've added a commandline argument `-s` to allow for silent execution of the script. This is useful if the user wants to automate the process of updating the `user.js` and cleaning prefs.

### DIFF
--- a/prefsCleaner.sh
+++ b/prefsCleaner.sh
@@ -48,7 +48,7 @@ fPrint() {
 }
 
 fUsage() {
-	echo -e "\nUsage: $0 [-s]" 1>&2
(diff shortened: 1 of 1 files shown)


earthlng:
Hi, thanks for the PR.

I like the idea but I don't think we need getopts + a new variable for 1 parameter.
I also want to keep showing the banner + text and calling the parameter 'silent' isn't really accurate either.

I'll close this and create a new PR and you can let me know what you think

4n0nym0u53:
Well yea, I'm new to shell scripting, so I didn't know you could do a one-liner. I just borrowed some inspiration from `updater.sh` :sweat_smile:

PR #1189, opened 2021-06-06
v1.4

earthlng:
- add -s parameter to start immediately / skip prompt / run non-interactive

This is useful if the user wants to automate the process of updating the user.js and cleaning prefs.

- fQuit: error messages to stderr
- fFF_check: info msg to stderr

Better support for suppressing/redirecting stdout while still showing any error messages in the console.
Useful for example with `prefsCleaner.sh -s >/dev/null`

### DIFF
--- a/prefsCleaner.sh
+++ b/prefsCleaner.sh
@@ -2,7 +2,7 @@
 
 ## prefs.js cleaner for Linux/Mac
 ## author: @claustromaniac
-## version: 1.3
+## version: 1.4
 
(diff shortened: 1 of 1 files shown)


earthlng:
@4n0nym0u53, let me know what you think

4n0nym0u53:
@earthlng I like it, nice job!

4n0nym0u53:
Replying to #1185 

Well yea, I'm new to shell scripting, so I didn't know you could do a one-liner. I just borrowed some inspiration from `updater.sh` tbh :sweat_smile:

earthlng:
ok, thanks, glad you like it

Issue #1221, opened 2021-07-30
rebrand RFP ALTS [was look at dom.enable_performance_navigation_timing]


sr093906:
dom.enable_performance_navigation_timing = false

saw it on https://www.privacy-handbuch.de/handbuch_21n.htm#29_07_21

rusty-snake:
The arkenfox user.js uses (unlike the user.js from privacy-handbuch.de) which reduces timer precision and enables jittering for javascript timers.

Thorin-Oakenpants:
- RFP mitigates a lot of timing attacks
  - one exception may be `dom.enable_resource_timing` see [this coment](https://gitlab.torproject.org/legacy/trac/-/issues/27268#note_2294531)
  - [PerformanceNavigationTiming](https://developer.mozilla.org/en-US/docs/Web/API/PerformanceNavigationTiming) is disabled by RFP
  - you can test [here](https://arkenfox.github.io/TZP/tzp.html#misc) - you need to reload the page for Performance Navigation Timing API changes to take effect
  - FF78+ [1511941](https://bugzilla.mozilla.org/1511941) - tweaks the already mitigated API to return disabled
  - hence Tor Browser has `dom.enable_performance_navigation_timing` at default `true` and arkenfox doesn't have it
- note: there are still ways get high resolution timers despite TB's efforts and RFP, and is not something that is easily solved, e.g. see FuzzyFox


`dom.enable_performance_navigation_timing` would fit the 4600 section. It was introduced in FF58 I think, not sure what FF version RFP protected it: assuming I read FF78's 1511941 correctly. I'll have to check, maybe RFP mitigation was built into the patch that added the API in FF58. And I need that info in order to add it to 4600s

But I think anything outside of RFP is a waste of time: the RFP ALTS section doesn't achieve anything IMO. In terms of fingerprinting you're better off using an extension like Canvas Blocker and randomizing canvas and maybe audio for some extra depth. I do get that timing attacks covers more than fingerprinting risks - but disabling all these APIs generally just causes breakage. At this threat level the user should be using Tor Browser

So I'm actually in a mind to relabel the RFP ALTs section, strip it down a bit, and make it all inactive like the 4700's with a DO NOT USE warning and add "use Tor Browser if your threat model fits it... for FPing use CB"

@earthlng What are thoughts on
- adding the pref to rfp alts
- doing something about RFP ALTs

---

[...]

Issue #1218, opened 2021-07-26
reminder: wiki page


Thorin-Oakenpants:
- 1.3.1 FPI/dFPI
- 1.3.2 RFP

And link to them in the implementation page. This is to help the user decide if they want these two things - i.e threat model, what it protects, what to expect in terms of usability/breakage. People seem to be surprised, and yet the list is really small, and can be worked around - everyone's mileage will differ

### 🔻 FPI

only breaks some cross site logins, e.g. SSOs - either use a different browser for those sites or switch to dFPI now that it exists - e.g. override recipe - pretty simple page
- edit: note: existing FPI site exceptions are keyed by OA

### 🔻 RFP

🔹 breakage (compat)
- is limited to about three things: canvas (use site exceptions: more on that), timing jank, alt shortcut keys on some sites
- note: should add a pic of a randomized canvas pattern

🔹 usability (not compat issues)
- usability issues as distinct from actual breakage
- issues include timezone, new window sizes (there are prefs), and prefers-color (light/dark). And somewhere in between breakage and usability would be blurry images (due to devicePixelRatio spoof).

Of ~100+ things RFP protects, that's it (will be the other odd item). I'm sick and tired of people claiming (reddit a billion times) that RFP generally breaks everything. Timezone is not breakage. Using en-US is not breakage (and en-US is optional). Prefers light is not breakage. They are expectation issues, not breakage.

There are solutions, but first of all the user needs to decide if RFP is for them = threat model. If they can't handle minor things like looks and feel (prefers-light, non-native widgets, letterboxing etc : yes widgets is not RFP and LBing is separate: just pointing out other FPing examples) then maybe it's not for them. The FPing threat and what RFP can do also needs to be put into context. For most users, just randomizing canvas will achieve the same result in most cases: e.g. just use Canvas Blocker

And something should be said about not undermining RFP protected metrics - like installing Dark Reader. There are solutions - the user just needs to be selective
- just like you can allow a canvas site exception: it won't be used to link all your traffic because
   - one, it's not all sites
   - two: it's only one metric
   - three: canvas alone, depending on the canvas being tested, is usually equivalency
   - four: not all excepted sites use the same canvas test
   - five: not all excepted sites share FP data
   - six: if it's a site you log into all the time, you're already ID'ed (as a repeat visitor)
   - seven: is it even being used to FP you?
- you could do the same for timezone, e.g. use an extension to spoof the timezone as your real one for e.g. just gmail

[...]

PR #1225, opened 2021-08-04
change 4600s into do not use, #1221

Thorin-Oakenpants:
This is a draft
- merge 4700's into 4600s
- remove old numbers in the square brackets
- remove notation of when RFP kicked in (that info is in 4500s)
- since we now do not recommend this section
   - cleanup info on each release in README section
   - do away with one char flip
   - move 4616 to deprecated where it belongs
   - remove "optional if..." lines
- start cleaning up references, descriptions to shorten the section
   - will list what I removed: e.g. bugzillas to when the pref was added are a bit useless

todo / consider
- 4600 title
- 4600 section description can be a lot better
- 4600 link to wiki page on RFP ( issue #1218 - that is, if RFP is not for you, then just use Canvas Blocker, which can leak but should fool naive scripts if any get thru etc )
- do we want to add dom.enable_performance_navigation_timing

while these all fit together as "covered by RFP", some of these seem out of place
- maybe we could split this into two
   - 4600: "optional without RFP" - these won't hurt RFP but they also won't help your fingerprinting - e.g. font vis, prefers-color, prefers-reduced-motion
   - 4700: "do not use EVER especially with RFP" - these will affect RFP, can break shit, etc, and won't help your fingerprinting - e.g. all the timing stuff, disabling APIs, etc
   - also. the webgl one seems a bit out of place since we disable webgl
   - we could always move some items back to their relevant sections as inactive with some sort of RFP tag/warning

I'm not sure what's the cleanest way to convey this. Anyway, pushing a PR to get some discussion going

### DIFF
--- a/user.js
+++ b/user.js
@@ -1512,14 +1512,11 @@ user_pref("_user.js.parrot", "4600 syntax error: the parrot's crossed the Jordan
  * [1] https://wiki.mozilla.org/Security/Reviews/Firefox/NavigationTimingAPI ***/
    // user_pref("dom.enable_performance", false);
 /* 4604: disable device sensor API
- * [1] https://blog.lukaszolejnik.com/stealing-sensitive-browser-data-with-the-w3c-ambient-light-sensor-api/
- * [2] https://developer.mozilla.org/en-US/docs/Web/API/Sensor_APIs
- * [3] https://bugzilla.mozilla.org/buglist.cgi?bug_id=1462308,1292751 ***/
+ * [1] https://developer.mozilla.org/en-US/docs/Web/API/Sensor_APIs
+ * [2] https://bugzilla.mozilla.org/buglist.cgi?bug_id=1462308,1292751 ***/
    // user_pref("device.sensors.enabled", false);
-/* 4605: disable site specific zoom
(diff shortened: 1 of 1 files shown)


Thorin-Oakenpants:
references removed so far

4601: spoof number of CPU cores [FF48+]
- bugzilla is only for when it was added, and I don't think we need tor tickets or even to explain what it is
```
   // [1] https://bugzilla.mozilla.org/1008453
   // [2] https://gitlab.torproject.org/tpo/applications/tor-browser/-/issues/21675
   // [3] https://gitlab.torproject.org/tpo/applications/tor-browser/-/issues/22127
   // [4] https://html.spec.whatwg.org/multipage/workers.html#navigator.hardwareconcurrency
```

4607: disable giving away network info [FF31+]
- bugzilla is just when the API was added
```
   // [3] https://bugzilla.mozilla.org/960426
```

4619: [2510] disable Web Audio API [FF51+]
- bugzilla is just for when the pref was added
```
   // [1] https://bugzilla.mozilla.org/1288359
```

Thorin-Oakenpants:
I consider this ready to push
- 3500+ bytes and 63 lines saved
- also we don't need to link the the wiki #1218 : i.e "is RFP for you" - that's for the 4500 section header

PR #1247, opened 2021-08-29
proxy direct failover

Thorin-Oakenpants:
https://hg.mozilla.org/integration/autoland/rev/b633563955ad

### DIFF
--- a/user.js
+++ b/user.js
@@ -992,17 +992,14 @@ user_pref("privacy.firstparty.isolate", true);
 
    [WARNING] DO NOT USE extensions to alter RFP protected metrics
 
- FF41+
-    418986 - limit window.screen & CSS media queries leaking identifiable info
+    418986 - limit window.screen & CSS media queries (FF41)
       [TEST] https://arkenfox.github.io/TZP/tzp.html#screen
- FF50+
-   1281949 - spoof screen orientation
-   1281963 - hide contents of navigator.plugins and navigator.mimeTypes
- FF55+
-   1330890 - spoof timezone as UTC0
-   1360039 - spoof navigator.hardwareConcurrency as 2
-   1217238 - reduce precision of time exposed by javascript
- FF56+
+   1281949 - spoof screen orientation (FF50)
+   1281963 - hide contents of navigator.plugins and navigator.mimeTypes (FF50-88)
+   1330890 - spoof timezone as UTC0 (FF55)
+   1360039 - spoof navigator.hardwareConcurrency as 2 (FF55)
+   1217238 - reduce precision of time exposed by javascript (FF55)
+ FF56
    1369303 - spoof/disable performance API
    1333651 - spoof User Agent & Navigator API
       JS: FF91+ the version is spoofed as ESR, and the OS as Windows 10, OS 10.15, Android 10, or Linux
@@ -1012,7 +1009,7 @@ user_pref("privacy.firstparty.isolate", true);
    1337161 - hide gamepads from content
    1372072 - spoof network information API as "unknown" when dom.netinfo.enabled = true
    1333641 - reduce fingerprinting in WebSpeech API
- FF57+
+ FF57
    1369309 - spoof media statistics
    1382499 - reduce screen co-ordinate fingerprinting in Touch API
    1217290 & 1409677 - enable some fingerprinting resistance for WebGL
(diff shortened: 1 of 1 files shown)


Thorin-Oakenpants:
needs more cowbell.. feel free to edit and merge

Thorin-Oakenpants:
bump

Thorin-Oakenpants:
@Jee-Hex @rusty-snake .. looks like earthlng has gone again, trying to save the world from trumpism .. is 0706 correct? please advise .. I've got my 91 changelog post ready to go .. its a doozey!

rusty-snake:
I do not understood it to 100% yet (and the bugzilla and phabricator links require log-in :disappointed: ) but from what I understood so far (FF91+):
- `network.proxy.failover_direct=true` (default) if there is a broken proxy config firefox fails back to direct connections if 1. there is no other working proxy-config and 2, the proxy wasn't configured manually by the user.
- `network.proxy.failover_direct=false` firefox never skips a proxy. A broken proxy config will always break.

==> Setting `network.proxy.failover_direct=false` sounds right.

Points that I did not understood so far:
- What are system principal requests?
- Blocked bugzillas are usually about security, can this affect security?

Thorin-Oakenpants:
[system principal](https://blog.mozilla.org/attack-and-defense/2020/06/10/understanding-web-security-checks-in-firefox-part-1/)

Yeah, false is what we want, so nothing bypasses your proxy settings. I have no idea if the bugzilla is directly security related, if it was then it the default value would be false IMO. I just thought the description needs to be better. I copied the line from the patch to get something up for E to edit and push .. but he vanished again. Proxies and networking is not my area. Guess we'll wait :)

rusty-snake:
> system principal

Thanks.

>  just thought the description needs to be better.

So based on what I understood so far: `disable direct connections for privileged origins on misconfiguration proxys` (or maybe use chrome instead of privileged)

Thorin-Oakenpants:
> disable direct connections for privileged origins on misconfiguration proxys

Not very user friendly, and neither is mine

Jee-Hex:
>  I have no idea if the bugzilla is directly security related, if it was then it the default value would be false IMO. 

Apparently this pref was introduced to "combat malicious add-ons".

> the description 


Maybe something like `Prevent system requests from failing over outside the proxy` or
`Prevent system principal requests from falling back to DIRECT connections`, give or take?

earthlng:
> Apparently this pref was introduced to "combat malicious add-ons".

where did you find that?

Sure would be nice if the bugzilla wasn't hidden from the public and we could see the discussion and reasoning for this new pref,
and fe. the likely scenario that they think this needs to protect against.
I don't really see how falling back to Direct combats malicious addons if fe. an addon sets a proxy to spy on their users.
It does more of the opposite, doesn't it? Like when the intercepting proxy isn't reachable for whatever reason, users are less likely to notice that a malicious addon is intercepting them when Firefox just switches over to a direct connection, no?
The only other "malicious" use I can imagine, is an addon configuring a nonexistant proxy address to block all internet access but that doesn't really make sense either because it's kinda stupid, easily noticeable and can also be done with the "requests" API.
So yeah, IDK.

IMO, just "disable proxy failover to DIRECT", or something like that, is good enough. No need to mention "system requests". IDK if it needs an additional description either.

Thorin-Oakenpants:
> IMO, just "disable proxy failover to DIRECT", or something like that, is good enough. No need to mention "system requests". IDK if it needs an additional description either.

The system principal seems like a very important variable here. Since your proposal isn't really any different to mine (except the "system" bit), then I'll leave it as is. We can always update it later

> Apparently this pref was introduced to "combat malicious add-ons"

I'm confused. I thought web ext wasn't system principal, but expanded principal

Jee-Hex:
> where did you find that?

[here](https://gitlab.torproject.org/tpo/applications/tor-browser/-/issues/40548)

> I don't really see how falling back to Direct combats malicious addons if fe. an addon sets a proxy to spy on their users.

> I'm confused. I thought web ext wasn't system principal, but expanded principal

presumably they wanted to enforce the add-on blocklist?

Thorin-Oakenpants:
> here

Ahh OK. I pinged sysrqb on it, but he never replied. Not sure how I missed that issue. And here I was thinking it might be some weird edge cases for web content like service worker blobs or something

Thorin-Oakenpants:
I know more now, got an email from @tomrittervg (moz security dev) , not sure what I can share about it. 

From a security standpoint for Firefox, we want the pref to be true (default), and yes, it protects against malicious addons interfering with legit FF system principal proxy requests

For Tor Browser, I get the proxy leak aspect

So we should make this inactive and stick it under harden with a more info: at the end of the day, people need to vet their extensions, so I'm not terrible torn up about it - class discuss

PR #1288, opened 2021-12-04
remove rfpalts

Thorin-Oakenpants:


Thorin-Oakenpants:
optionally, if you want we could add `opsec` (5000s) and/or `harden` (5500s) and change their syntax to onechar switches

earthlng:
for rfpalts removal: also need to remove the 2 lines in `showhelp`.

If it's helpful to offer enabling of opsec and harden sections, we can add those.

PR #1309, opened 2021-12-21
Remove "-e" (ESR) argument from updater.sh

Minimalist73:
I was checking the updater.sh script and it seems the "-e" argument is no longer needed. 
When used, it only enable `user_pref("browser.urlbar.suggest.quicksuggest", false);` which is deprecated `9999`, it seems to have been replaced by 2 prefs in `0807`. Since it's a single entry, the user can put that in his user-overrides if needed.

### DIFF
--- a/updater.sh
+++ b/updater.sh
@@ -37,7 +37,6 @@ COMPARE=false
 SKIPOVERRIDE=false
 VIEW=false
 PROFILE_PATH=false
-ESR=false
 
 # Download method priority: curl -> wget
 DOWNLOAD_METHOD=''
@@ -93,8 +92,7 @@ Optional Arguments:
                          Ex: -o \"override folder\"
     -n           Do not append any overrides, even if user-overrides.js exists.
     -v           Open the resulting user.js file.
-    -r           Only download user.js to a temporary file and open it.
-    -e           Activate ESR related preferences."
+    -r           Only download user.js to a temporary file and open it."
   echo
   exit 1
 }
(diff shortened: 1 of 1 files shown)


Thorin-Oakenpants:
The purpose of it is to enable the ESR section in deprecated for ESR users, where those prefs are still useful. Deprecated items grow during the ESR lifecycle. We maintain the state (active/inactive) as it was before it got moved there

That the current lot only contains one active item is not relevant. If you look at #123 and scroll to the end of the first post, you can view archived ESR deprecated sections: for `FF79-91` for example, I count `12` items that are active

Minimalist73:
> That the current lot only contains one active item is not relevant. If you look at #123 and scroll to the end of the first post, you can view archived ESR deprecated sections: for `FF79-91` for example, I count `12` items that are active

I still don't get the point of that "-e" argument if you need to manually get those deprecated preferences and add them to the user.js or user-overrides.js file. The updater get the latest version, that's the point of it, so all removed preferences don't come with it automatically. Since it's mostly use for automatic update (at least, that's it for me), that "-e" argument is not that useful unless we add a way to recover those removed preferences and add it to the user.js, but once more it's better to do it manually since not all preferences are useful for everyone. Like I said in my PR message, if the user really want those deprecated pref activated, he can do so manually and add removed one's if he want to since they are all in the same section.

Thorin-Oakenpants:
> if you need to manually get those deprecated preferences and add them to the user.js or user-overrides.js file

you don't add them to anything (they are already in the user.js), you just pass the argument

Minimalist73:
> > if you need to manually get those deprecated preferences and add them to the user.js or user-overrides.js file
> 
> you don't add them to anything (they are already in the user.js), you just pass the argument

So the "-e" argument is there for a future deprecation of those active preferences and if the user want them anyway if I understand correctly? If that's it then it has a use case, it's just not "obvious" right now since there's only one entry available in the deprecated section, that's why I tough it was better to just remove that argument.

ByJumperX4:
-e enables options that have been deprecated for Firefox rolling, but that are still useful in Firefox ESR (currently 91.x).

PR #1372, opened 2022-02-15
start v98

Thorin-Oakenpants:
note: FF99+ new crlite mode 3
- bug: https://bugzilla.mozilla.org/show_bug.cgi?id=1753071
- patch: https://phabricator.services.mozilla.com/D137553


fxbrit:
what are your thoughts on this new mode 3? currently OCSP is enabled but CRL is not allowed to fallback to it, so I guess that since AF wants OCSP `stapled || (nostapled + hard fail)` it would make sense to use the combination that mode 3 provides, when it's available of course.

Thorin-Oakenpants:
I don't see the need for any other changes, except the date - @fxbrit

fxbrit:
^ I cosign this invented quote, making it a real quote from now on.

fxbrit (approved):

PR #1386, opened 2022-02-27
Replace troubleshooting-help.md with troubleshooting-help.yml

rusty-snake:
Together with #1385 this will look like
https://github.com/rusty-snake/user.js/issues/new/choose
https://github.com/rusty-snake/user.js/issues/new?assignees=&labels=&template=troubleshooting-help.yml


Thorin-Oakenpants:
interesting. I have a saved reply and was thinking of some sort of automation / action to reply and close as invalid if the checkbox wasn't checked (in the current md)

I like this, sorta: needs a few tweaks (combine some fields), reduce field title size maybe add 🔹 or something to break it up when posted - a new unfilled issue is very "long" and goes off screen.

The one thing I don't like about it, is that unless its filled in, people won't post - so I don't get to see what is upsetting users - and we could miss something. It doesn't have to be anything super important, it could just be someone saying Disney+ doesn't work with RFP and that's adds to the knowledge base that it's probably the same issue as Netflix

---

I like the wiki and troubleshooting links at https://github.com/rusty-snake/user.js/issues/new/choose
- instead of new buttons, in the existing current https://github.com/arkenfox/user.js/issues/new/choose
   - maybe we add sub bullet points to troubleshooting

*Troubleshooting help**
Ask for help to solve problems with user.js
- Did you read the [wiki](https://github.com/arkenfox/user.js/wiki)?
- Did you do these [checks](https://github.com/arkenfox/user.js/wiki/5.2-Troubleshooting#-check)
- Did you [confirm](https://github.com/arkenfox/user.js/wiki/5.2-Troubleshooting#-confirm) the issue is caused by arkenfox
- Did you try [manually](https://github.com/arkenfox/user.js/wiki/5.2-Troubleshooting#-manual) troubleshooting further

But then this feels like reiterating everything, and I don;t want multiple copies of lists to maintain

@fxbrit What are your thoughts. I saw you were doing something similar at the unicorn rainbow gitlab world of LW: closing shit tickets and going to implement some sort of checklist - e.g. does this happen in FF to eliminate the upstream stuff)

TBH, I can see some improvements we can use without all this. I'd be more interested in autoreply-and-close as invalid - because then there's a record and it can always be un-labeled and re-opened should the user provide more info and we can deal with it, and it could unearth something useful

fxbrit:
hello y'all :-) at first I really wanted to have a strict checklist of "must-do" stuff but I'm not super sold anymore. I would rather use the checklist as something that helps a maintainer decide whether he wants to invest time in the issue or not.

for example:
- something not caused by AF, but worth reporting upstream.
- something without a precise STR because user doesn't know better, but still worth investigating on our end (eg. possible hard to debug regression).

at the same time there's the counter argument that 90% of closed issues are invalid and rather useless, so discouraging them is good (not that pants isn't scary enoug, but still..). my two cents:
- behavior and STR should not be mandatory.
- OS, description and checkboxes being mandatory allows to determine whether the issue is worth it. at that point one can ask behavior and STR, quarantine the issue, and close it if they aren't provided + the issue isn't worth research time.
- even if the user has something useful to say, reading the troubleshooting guidelines should be a must. netiquette is good.
- does github allow to condense the layout? I fear length will scare issues away even when they exist..

Thorin-Oakenpants:
> I fear length will scare issues away even when they exist

I also noticed that - hence I wanted to combine a couple of fields (and reduce the size of the field titles - maybe remove the description). In my test I deliberately left a checkbox unchecked and by the time I got to the submit button (which was disabled) the cause was out of sight

> not that pants isn't scary enoug(h)

I can be MOAR scary when needed

> something not caused by AF, but worth reporting upstream

they can still report things using the other issue templates, or just tick the boxes and comment : "this is not an AF issue but worth noting"

> required checkboxes

people will just check them in order to submit, which defeats the purpose - I want them to actually read what is being asked - failure to check it tells me a lot

---

so I'm inclined to not use forms (takes too much space) and not use required fields (forms, checkboxes). I think the current setup suffices
- 🟥 red square and warnings\
- default title changed
- i can handle the volume of invalid shit with my saved replies (for now)

Thorin-Oakenpants:
@rusty-snake thanks for this, we may use it later - lets just see how things pan out over the next several months

PR #1402, opened 2022-03-22
Making the silent flag suppress output

sjoblomj:
The `-s` flag now suppresses output if no errors occur. This is useful for updating in scripts.

### DIFF
--- a/prefsCleaner.sh
+++ b/prefsCleaner.sh
@@ -20,7 +20,7 @@ cd "$(dirname "${sfp}")"
 fQuit() {
 	## change directory back to the original working directory
 	cd "${currdir}"
-	[ "$1" -eq 0 ] && echo -e "\n$2" || echo -e "\n$2" >&2
+	[ "$1" -ne 0 ] && echo -e "\n$2" >&2
 	exit $1
 }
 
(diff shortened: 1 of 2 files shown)


Thorin-Oakenpants:
Thanks. I've pinged earthlng to review, since I don't look after the updater scripts

earthlng:
Hi and thanks for the PR! 

`-s` isn't really the "silent" flag and I was careful not to describe it like that, see fe [here](https://github.com/arkenfox/user.js/commit/9018577a3e4880b881981e801891a3531c433aba).

The changes you're proposing aren't really necessary because you can just redirect STDOUT and/or STDERR in your updating scripts if you don't care about that output. fe
`prefsCleaner.sh -s >/dev/null`
or `prefsCleaner.sh -s >/dev/null 2>&1`
or `prefsCleaner.sh -s >/dev/null 2>/my/logs/prefscleaner.err`
or `prefsCleaner.sh -s >/dev/null 2>>/my/logs/prefscleaner.err` if you want to append to the logfile.

I don't want to make the script more complicated by adding additional parameters to be passed around, when the few people using an external script to call our scripts already have an easy solution to achieve the same.

Issue #1374, opened 2022-02-17
prefsCleaner.sh cannot be run on fresh install of Pop!_OS 21.10


Panja0:
- prefsCleaner.sh
  - Go to the Firefox profile directory and run prefsCleaner.sh
  - Error message: permission denied, 
  - Error message: When using sudo sh prefsCleaner.sh I get: 12: bad substitution, 15: bad substitution.

Fresh install of Pop!_OS 21.10, fully updated.
Any clue what's going on?

rusty-snake:
> Error message: permission denied,

How did you run it?

Did you made it executable?

>  sudo sh prefsCleaner.sh

DO NOT PRETEND `sudo` BEFORE A COMMAND IF IT IS NOT WORKING, `sudo` ISN'T A MAGICAL BUGFIXER!

prefsCleaner.sh is a bash script. You need to start it in bash, not dash, zsh, ksh or whatever, bash is the only right interpreter.

Panja0:
> > Error message: permission denied,
> 
> How did you run it?
> 
> Did you made it executable?
> 
> > sudo sh prefsCleaner.sh
> 
> DO NOT PRETEND `sudo` BEFORE A COMMAND IF IT IS NOT WORKING, `sudo` ISN'T A MAGICAL BUGFIXER!
> 
> prefsCleaner.sh is a bash script. You need to start it in bash, not dash, zsh, ksh or whatever, bash is the only right interpreter.

Cheers for the fast reply!
After applying `chmod +x prefsCleaner.sh` I was able to run the script without problems.
Thanks for pointing me in the right direction!

Strange thing is that updater.sh was already executable without applying chmod +x
It was executable right away so I thought prefsCleaner.sh had to be executable as well, but it wasn't.

PR #1416, opened 2022-04-06
Made prefsCleaner.sh executable

i-c-u-p:
Changed permissions of prefsCleaner.sh from 644 to 755 to be able to run it via "./prefsCleaner.sh" with out first executing "chmod +x prefsCleaner.sh". updater.sh is already executable so it's permissions don't need to be changed.


rusty-snake:
Fixes #1374

earthlng:
Thanks!

Issue #1418, opened 2022-04-08
ToDo: diffs FF98-FF99


earthlng:
FF99 is [scheduled](https://wiki.mozilla.org/Release_Management/Calendar) for release Apr. 5th

[FF99 release notes](https://www.mozilla.org/firefox/99.0/releasenotes/)
[FF99 for developers](https://developer.mozilla.org/docs/Mozilla/Firefox/Releases/99)
[FF99 security advisories](https://www.mozilla.org/en-US/security/advisories/mfsa2022-13/)

---

69 diffs ( 38 new, 11 gone, 20 different )

### new in v99.0:
nothing to see

### removed, renamed or hidden in  v99.0:
- [x] pref("security.csp.enable", true); - [1754301](https://bugzilla.mozilla.org/1754301) - https://github.com/arkenfox/user.js/pull/1419/commits/9aae0a62b03150e1c09d434ea51fe196751e9a75

### changed in  v99.0:
- for visibility/info
   - pref("dom.serviceWorkers.navigationPreload.enabled", true); // prev: false - [1750515](https://bugzilla.mozilla.org/show_bug.cgi?id=1750515) - [MDN NavigationPreloadManager](https://developer.mozilla.org/en-US/docs/Web/API/NavigationPreloadManager)

---

### ignore

<details><summary>click me for details</summary><p>

==NEW
```js
pref("browser.urlbar.bestMatch.blockingEnabled", false);
pref("browser.urlbar.suggest.bestmatch", true);
pref("devtools.inspector.selectedSidebar", "layoutview");
pref("devtools.netmonitor.customRequest", "{}");
pref("docshell.shistory.bfcache.ship_allow_beforeunload_listeners", false);
pref("dom.script_loader.delazification.max_size", 10485760);
pref("dom.script_loader.delazification.min_mem", 2);
pref("dom.script_loader.delazification.strategy", 255);
pref("dom.streams.pipeTo.enabled", false);
pref("dom.streams.transform_streams.enabled", false);
pref("extensions.formautofill.addresses.supported", "detect");
pref("extensions.formautofill.addresses.supportedCountries", "US,CA");
pref("extensions.formautofill.creditCards.supported", "detect");
pref("extensions.formautofill.creditCards.supportedCountries", "US,CA,UK,FR,DE");
pref("extensions.pocket.refresh.hideRecentSaves.enabled", false);
pref("gfx.font_rendering.directwrite.bold_simulation", 1);
[...]

PR #1419, opened 2022-04-08
v99

fxbrit:
#1418

### DIFF
--- a/user.js
+++ b/user.js
@@ -1081,9 +1081,6 @@ user_pref("extensions.blocklist.enabled", true); // [DEFAULT: true]
 /* 6002: enforce no referer spoofing
  * [WHY] Spoofing can affect CSRF (Cross-Site Request Forgery) protections ***/
 user_pref("network.http.referer.spoofSource", false); // [DEFAULT: false]
-/* 6003: enforce CSP (Content Security Policy)
- * [1] https://developer.mozilla.org/docs/Web/HTTP/CSP ***/
-user_pref("security.csp.enable", true); // [DEFAULT: true]
 /* 6004: enforce a security delay on some confirmation dialogs such as install, open/save
  * [1] https://www.squarefree.com/2004/07/01/race-conditions-in-security-dialogs/ ***/
 user_pref("security.dialog_enable_delay", 1000); // [DEFAULT: 1000]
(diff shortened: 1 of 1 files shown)


Thorin-Oakenpants:
am done with diffs, I didn't see anything there of any concern

one last thing I would like to do is look at `2620` `pdfjs.disabled` since RFP now reports it pdfjs as enabled (which is the default and which we enforce). I should check the browser behavior with an actual pdf (the pref directly affects the navigator property) - and depending on the results, add something to the effect that RFP enforces pdfjs as enabled

Not keen on moving it to `[SECTION 8000]: DON'T BOTHER: FINGERPRINTING` because those who don't use RFP may get the wrong message (it is perfectly fine to override as true it if you are not using RFP, in favor of some other pdf viewer - we set a good default and it's on the end user, not us)

Thorin-Oakenpants:
um, the pdfjs thing is not a blocker, it doesn't take affect until FF100

fxbrit:
yup, looks good considering that the only thing that stood out to me was the offscreen canvas stuff. merging.

szepeviktor:
@Thorin-Oakenpants Why have you put back `security.csp.enable` after removing it?
Thanks.

Thorin-Oakenpants:
I didn't remove it, fxbrit did. For deprecated items we **_move_** them to section 9999 (which is commented out), so I had to fixup the 🐟 's boo boo

fxbrit:
https://github.com/arkenfox/user.js/pull/1419#issuecomment-1093796489 maybe turn that into an issue as a reminder?

Thorin-Oakenpants:
^ IDK how to test this properly because handlers is a bit of a labyrinth to me
- e.g. https://bugzilla.mozilla.org/show_bug.cgi?id=1747343 just landed in 101
- and there have been some other pdf bugs as well that I've seen resolved/fixed (I check new bugzillas daily)
- and each OS may behave differently
- what test do we use, just a simple pdf to view? an inline one? IDK

So IDK if RFP affects actual handling, but it at least reports back to websites that pdfs are enabled and doesn't shouldn't you install adobe pdf viewer or whatever

fxbrit:
it seems kinda broken honestly, I will open a new issue to report a little test of mine.

PR #1423, opened 2022-04-16
v100

Thorin-Oakenpants:
FYI: 2654: https://bugzilla.mozilla.org/show_bug.cgi?id=1747343

### DIFF
--- a/user.js
+++ b/user.js
@@ -1,7 +1,7 @@
 /******
 *    name: arkenfox user.js
-*    date: 9 April 2022
-* version: 99
+*    date: 7 May 2022
+* version: 100
 *     url: https://github.com/arkenfox/user.js
 * license: MIT: https://github.com/arkenfox/user.js/blob/master/LICENSE.txt
 
@@ -741,6 +741,9 @@ user_pref("browser.download.useDownloadDir", false);
 user_pref("browser.download.alwaysOpenPanel", false);
 /* 2653: disable adding downloads to the system's "recent documents" list ***/
 user_pref("browser.download.manager.addToRecentDocs", false);
+/* 2654: enable user interaction for security by always asking how to handle new mimetypes [FF101+]
+ * [SETTING] General>Files and Applications>What should Firefox do with other files ***/
+user_pref("browser.download.always_ask_before_handling_new_types", true);
 
 /** EXTENSIONS ***/
 /* 2660: lock down allowed extension directories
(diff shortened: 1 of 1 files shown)


user.js
```diff
@@ -741,6 +741,9 @@ user_pref("browser.download.useDownloadDir", false);
 user_pref("browser.download.alwaysOpenPanel", false);
 /* 2653: disable adding downloads to the system's "recent documents" list ***/
 user_pref("browser.download.manager.addToRecentDocs", false);
+/* 2654: enable user interaction for security by always asking how to handle new mimetypes [FF101+]
+ * [SETTING] General>Files and Applications>What should Firefox do with other files ***/
+user_pref("browser.download.always_ask_before_handling_new_types", true);
```
fxbrit, line 746: is it really giving extra security in this case? I think that's mostly provided by `browser.download.useDownloadDir`, which is always going to kick in and prompt users in regardless of the mimetype.

this new pref mostly seems like an usability thing to give users the option to handle downloads with granularity, but I don't think picking a certain app to handle a file is necessarily security related. it's also somewhat annoying for users who want to just save a file and will be prompted twice like in past releases.

Thorin-Oakenpants: _probably_ (I didn't test and am only assuming 2651 kicks in correctly, I don't really care TBH) not in our default arkenfox

But as a standalone pref, the description is correct - it adds security, because 2651's default in FF is true - e.g. if users override 2651

fxbrit: so you plan to leave both flipped for redundancy? (speaking exclusively of the user.js)

Thorin-Oakenpants: is it though? asking where to save is not the same as asking how to handle a mime type
- https://www.thinkbroadband.com/download
- download the 5mb zip file (continue on HTTPS error)
- result: native save dialog (zero to do with FF options)

fxbrit: > asking how to handle a mime type

but is this really security related? it seems like a convenience thing more than anything (except if `2651` is set to false, but at that point the user knows he's taking the choice).

Thorin-Oakenpants: fuck ... when you resolve something with a comment and then un-resolve it (because I didn't like how it collapsed the thread), the comment is removed

Yes, it is. Because ultimately it ensures user interaction and decision making when first encountering a new mime type which has implications with disk leaks and app isolation. Maybe security isn't the right word, and the user probably meant to download/open the file (which means save to disk or open in a different app) - a default save to disk doesn't mean execution. But not changing it, it's consistent with the wording in 2651

fxbrit: k, I think we can safely collapse it now :-)

Thorin-Oakenpants: i'd rather leave patch discussions expanded

fxbrit: looks like v101 then

Thorin-Oakenpants: what looks like v101?

fxbrit: my mac screen is so small I had never noticed the `[FF101+]` in the damn comment because I had to scroll horizontally, don't mind me :-}

PR #1442, opened 2022-05-03
prefsCleaner.sh: Remove unnecessary chdir calls

madduck:
The `prefsCleaner.sh` script chdir()'s to the script's directory on
execution, which isn't necessary at all, and also means that has to
reside within the Firefox profile directory, rather than being callable
directly from the Git clone.

`updater.sh` has the same "problem", but honours the `-p` switch to let
the working directory be overridden.

This patch removes the unnecessary code, such that the `prefsCleaner.sh`
script works on `./user.js`.

Signed-off-by: martin f. krafft <madduck@madduck.net>

### DIFF
--- a/prefsCleaner.sh
+++ b/prefsCleaner.sh
@@ -6,20 +6,8 @@
 
 ## special thanks to @overdodactyl and @earthlng for a few snippets that I stol..*cough* borrowed from the updater.sh
 
-currdir=$(pwd)
-
-## get the full path of this script (readlink for Linux, greadlink for Mac with coreutils installed)
-sfp=$(readlink -f "${BASH_SOURCE[0]}" 2>/dev/null || greadlink -f "${BASH_SOURCE[0]}" 2>/dev/null)
-
-## fallback for Macs without coreutils
-if [ -z "$sfp" ]; then sfp=${BASH_SOURCE[0]}; fi
-
-## change directory to the Firefox profile directory
-cd "$(dirname "${sfp}")"
-
 fQuit() {
 	## change directory back to the original working directory
-	cd "${currdir}"
 	[ "$1" -eq 0 ] && echo -e "\n$2" || echo -e "\n$2" >&2
 	exit $1
 }


CelestialNebula:
Apparently this is the reason for it: https://github.com/arkenfox/user.js/issues/1403#issuecomment-1079008482

> The readlink stuff is there to make the script work with symlinks but
the other way around from what you want to do, ie create a symlink fe on
the desktop (fe named "update profile xyz") for an updater/prefscleaner
in a particular profile.  It's like that because at the time we ported
the script, this mimicked the code we already had in the Windows
version.

madduck:
I suppose you can't suit everyone, but there are separate scripts for Linux and Windows, and while the Windows ones can and should implement Windows-like behaviour, there is no real reason to stick to that with Linux, especially if it's a bit alien to the platform to do things that way, don't you think?

I don't want to copy the scripts out of the Git clone into the profile directory, but track the scripts with Git, so that I can easily switch branches and obtain the latest version, and then just always run the right one…

benoit-pierre:
In any case the code to save/restore the current directory is completely unnecessary and can be removed:
```shell
currdir=$(pwd)
[…]
cd "${currdir}"
```

> updater.sh has the same "problem", but honours the -p switch to let
the working directory be overridden.

How about a PR to add similar support to this script?

madduck:
@benoit-pierre honestly, I don't think it's worth the effort. I don't understand why we're maintaining Windows shortcut compatibility in a shell script, and it's not like `updater.sh`'s argument parsing is anything that should be replicated. I've just written a wrapper script that does what it needs to to get the job done.

madduck:
@Thorin-Oakenpants gotta love projects where pull requests are closed without comment…

Thorin-Oakenpants:
I commented on one of the other plethora of PRs to do with updater.bat, which I am sure you have seen or should have given the interest in updater.bat, that nothing will happen with these without earthlng and he will probably never have time

PR #1446, opened 2022-05-14
prefsCleaner: fix profile-in-use check on Linux

benoit-pierre:
I went as far back as Firefox 60, on Linux, `lock` is always a broken symlink to a target looking like this: `127.0.0.1:+8942`, where `8942` is the PID of the Firefox process. Consequently, the `[[ -e lock ]]` check always fail (can't dereference a broken symlink). Additionally, starting around version 80, it's not removed on exit, so just changing the check to `[[ -L lock ]]` on Linux does not work.

I don't have access to a macOS machine, I assumed the `[[ -e lock ]]` check is working fine on that OS, it would be great if someone could confirm that's the case.

### DIFF
--- a/prefsCleaner.sh
+++ b/prefsCleaner.sh
@@ -31,10 +31,33 @@ Optional Arguments:
     -s           Start immediately"
 }
 
-fFF_check() {
+fFF_check_linux() {
+	local target="$(readlink lock)" || return 1
+	local pid="${target#*+}"
+	# Account for PID reuse on Linux.
+	local procdir="/proc/$pid/fd"
+	[[ -d "$procdir" ]] || return 1
+	parentlock="$(realpath .parentlock)"
+	readlink "$procdir"/* | grep -Fqx "$parentlock"
+}
+
+fFF_check_macos() {
 	# there are many ways to see if firefox is running or not, some more reliable than others
 	# this isn't elegant and might not be future-proof but should at least be compatible with any environment
-	while [ -e lock ]; do
+	[[ -e lock ]]
+}
+
+fFF_check() {
+	local variant=''
+	case "$(uname -s)" in
+		Linux) variant='linux' ;;
+		Darwin) variant='macos' ;;
+		*)
+		echo -e "\nPlease make sure the Firefox profile is not in use before continuing.\n" >&2
+		read -r -p "Press any key to continue."
+		return
+	esac
+	while "fFF_check_$variant"; do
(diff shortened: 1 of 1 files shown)


benoit-pierre:
Oops, I initially pushed a test version instead of the right one…

CelestialNebula:
I tried to make the script closer to POSIX (https://github.com/arkenfox/user.js/pull/1430) and added some other files.  

Could you check if they are added when Firefox is running and
removed when it's closed?
```
  cookies.sqlite-wal
  favicons.sqlite-wal
  places.sqlite-wal
```
I also left `lock` since its been in awhile and no one complained so I assumed it works on mac.

benoit-pierre:
Looking at the presence of WAL files feel way more brittle... An they don't get removed in case of crash. Anyway, another option would to just drop the check, and just keep a warning to the user to ensure the Firefox profile is in use before continuing, like I did when the platform is not Linux / macOS.

CelestialNebula:
Could you check if this works for you (I'm assuming you are using gnu
coreutils).  This is mostly just trying to bring it closer to POSIX (+
formatting changes).  BusyBox readlink didn't seem to like that so I
switched to using find + readlink.

```
fFF_check_linux() {
	f_target="$(readlink lock)" || return 1
	f_pid="${f_target#*+}"

	# Account for PID reuse on Linux.
	f_procdir="/proc/$f_pid/fd"

	[ -d "$f_procdir" ] || return 1

	f_parentlock="$(realpath .parentlock)"

	find "$f_procdir" -type l -exec readlink {} \; \
	     | grep -Fqx "$f_parentlock"
}

fFF_check_macos() {
	      [ -e lock ]
}

# We need to see if THIS Firefox profile is running.
fFF_check() {
	f_variant=''

	case "$(uname -s)" in
	Linux)
		f_variant='linux'
		;;
	Darwin)
		f_variant='macos'
		;;
	*)
		printf "Please make sure the Firefox profile is not in use before continuing.\n" >&2
		printf "Press Enter to continue."
		read -r
		return
		;;
	esac

	while "fFF_check_${f_variant}"; do
		printf "\nThis Firefox profile seems to be in use.  Close Firefox and try again.\n" >&2
		printf "Press Enter to continue."
		read -r
	done
}
```

Thorin-Oakenpants:
just be aware that nobody will do anything about any of these PRs except earthlng, and he's only appears once a release to provide diffs. IF he has time he might look at these

PR #1430, opened 2022-04-25
Move prefsCleaner.sh closer to POSIX

CelestialNebula:

PR #1458, opened 2022-06-02
updater.sh: rewrite in POSIX sh

fluorescent-haze:
I rewrote updater.sh in POSIX sh, so that bash is no longer a dependency. I've omitted a lot of flags either because I haven't gotten around to implementing them, or I'm skeptical of their necessity.

All the most important features (except for updating the script itself) have been implemented.

Missing features (that I know of)
---------------------------------
- [ ] choose firefox profile from a list
- [ ] update updater.sh (planned)
- [ ] one user.js backup only
- [ ] create a diff file
- [ ] custom overrides file or directory
- [ ] open user.js
- [ ] download user.js to a temporary file

Changes
-------
- ISO date format for backup files
- Prompt for confirmation, if no prefs.js file is found (to prevent accidentally runnning in non-profile directory)
- More concise & consistent output.

If there is interest in merging this, I'm willing to add more features from the original bash script before doing so.

### DIFF
--- a/updater.sh
+++ b/updater.sh
@@ -35,11 +35,11 @@ read_yn() {
 
 usage() {
     cat << EOF
-Usage: updater.sh [options] [PROFILE]
+usage: updater.sh [options] [<profile>]
 
 Update the user.js file and append custom configuration.
 
-Options:
+options:
     -e  Activate ESR preferences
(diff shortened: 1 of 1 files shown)


rusty-snake:
See also #1430

PR #1477, opened 2022-06-14
v102

Thorin-Oakenpants:


Thorin-Oakenpants:
@fxbrit I think we are done here, please double-check #1490

fxbrit (approved):
LGTM

Thorin-Oakenpants:
The only thing I'm concerned about is that ESR102 will change, but it's cosmetic I think - i.e the exceptions are still honored in onShutdown

fxbrit:
you mean you are worried because ESR will have the unchecked "Delete cookies and site data when Firefox is closed" because they are not going to backport any of that (from https://github.com/arkenfox/user.js/issues/1489#issuecomment-1172020374)? I think the ESR user base is rather small and we could link to an explanation when new issues come, nah?

also, I hope everything will be backported at some point of the ESR102 cycle.

Thorin-Oakenpants:
I want ESR users to use the correct matching arkenfox version, so this is it, so I want it to be as correct as possible. I do try and support ESR in the live master, but it's a bit of a PITA what with a year between releases.

Right now, ESR91 users on 102-103 won't have cookie site exceptions honored because we flipped back to sanitize onShutdown - so I should really add something [here](https://github.com/arkenfox/user.js/blob/master/user.js#L33-L35)

As for ESR102, they will back out the patch - https://bugzilla.mozilla.org/show_bug.cgi?id=1777419#c4, so my understanding is the UI will get fixed and everything works as per 101, and onShutdown still respects exceptions in onShudown. So I'm nnot worried about the UI

So I _think_ the descriptions are OK and valid. Just wanted to double check. I thought LW moved to lifetime = default (0) and onShutdown (cookies, offlineApp) = true a couple of releases ago - so you should know, more than me. I was waiting until this release which was meant to be the one before the migration (but they messed up and we got it partially in 102)

fxbrit:
LW has `OnShutdown.cookies` = true, but the lifetime is still 2 at the moment (I plan to change that, I just haven't had the time to do it and check whether the UI handles that well etc).

but [these descriptions](https://github.com/arkenfox/user.js/pull/1477/files#diff-417e8f625f16252f8ace3b0791d24c9b073d7394e9216c7b5d14a516d2572277R768-R774) are indeed right when it comes to stable. and sure, ESR91 people using 102-103 will be screwed when it comes to exceptions, but we can add something later this week and provide support in the meantime. sounds like a deal?

Thorin-Oakenpants:
> LW has OnShutdown.cookies = true

but it respects site exceptions, so I'm good with the changes, which I would have added anyway in order to beat the migration code - just sucks that they added part of it by mistake (but they will roll back)

https://bugzilla.mozilla.org/show_bug.cgi?id=1777419#c9 re FF102
> As a workaround I suggest to use the History settings to clear data on shutdown. If you select Cookies, Cache, offline website data that has the same effect as using the "Delete cookies and site data when Firefox is closed" checkbox. It will take any exceptions you have set into account.

which confirms it (just being super sure here), and running prefsCleaner reverts lifetime pref to default

OK, I'm done here

PR #1511, opened 2022-07-19
make updater.sh check explicitly for Y/y instead of N/n

infinitewarp:
Hello user.js folks! 👋 Long time listener, first time caller here.

I recently used the `updater.sh` script to update one of my profiles, but I fat-fingered my reply to the confirmation prompt with a `T` instead of a `Y`, and it proceeded to update anyway. This seems bad! Confirmation prompts like this should err on the side of caution and only perform a destructive operation if the user explicitly enters an affirmative `Y`/`y` response. (Yes, I'm aware that the script makes a backup file, but still…)

I dug into the code and realized that you're doing the opposite: `updater.sh` only aborts upon receiving a `N`/`n` response. *Any* other input appears to be treated as a destructive `Y`/`y`. Yikes!

So, I swapped the logic around. It now treats only `Y`/`y` as an affirmative response.

Tested using `q` and then `y` in this demo of your version versus mine:

[asciicast](https://asciinema.org/a/509667)

### DIFF
--- a/updater.sh
+++ b/updater.sh
@@ -2,9 +2,9 @@
 
 ## arkenfox user.js updater for macOS and Linux
 
-## version: 3.4
+## version: 3.5
 ## Author: Pat Johnson (@overdodactyl)
-## Additional contributors: @earthlng, @ema-pe, @claustromaniac
+## Additional contributors: @earthlng, @ema-pe, @claustromaniac, @infinitewarp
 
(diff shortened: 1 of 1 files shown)


earthlng:
Hello 1st-time caller, I hope you enjoyed the show so far :)

Thanks for your PR, I agree with you and would gladly merge your PR but please change line 198 to
`[[ $REPLY =~ ^[Yy]$ ]] || return 0`

Thorin-Oakenpants:
Thanks @earthlng ❤️

PR #1514, opened 2022-07-25
Back up prefs.js to `prefsjs_backups` directory

xfzv:
`updater.sh` creates a `userjs_backups` directory and `user.js` backups are saved to it:

https://github.com/arkenfox/user.js/blob/996881aef184246a011ac29382e91d634cda7b65/updater.sh#L270-L273

Thought it would be nice to have the same behavior with `prefsCleaner.sh`, it unclutters the Firefox profile directory.

### DIFF
--- a/prefsCleaner.sh
+++ b/prefsCleaner.sh
@@ -69,7 +69,8 @@ fStart() {
 	fi
 
 	fFF_check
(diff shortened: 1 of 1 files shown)


earthlng:
good idea, thanks

PR #1515, opened 2022-07-27
Prevent leaking theme accent

SkewedZeppelin:
Tested working on Android/Mull 103 and Fedora/GNOME/Firefox 102

Reported here: https://forum.f-droid.org/t/websites-can-know-android-accent-color/19133


Thorin-Oakenpants:
For the record, AccentColor and AccentColorText are new in 103, but the values are always reported as fallback
- https://searchfox.org/mozilla-central/rev/2bbb0c0a90df20702df8c8011a8996536a83cb75/widget/nsXPLookAndFeel.cpp#1186

PR #1537, opened 2022-09-10
5506: Only link to CVE with firefox

rusty-snake:


rusty-snake:
Target branch is https://github.com/arkenfox/user.js/pull/1527.

Thorin-Oakenpants:
I prefer to keep it as is, rather than just show 3 Firefox CVEs. The existing link and 36 CVEs with six from this year, back up the item

>  * Vulnerabilities [1] have increasingly been found

rusty-snake:
Yes, but why link to CVEs oft other wasm runtimes which share no code with Firefox AFAIK.

Thorin-Oakenpants:
because the description is talking about wasm in general, not wasm in Firefox

PR #1541, opened 2022-09-11
v105

Thorin-Oakenpants:


user.js
```diff
@@ -87,25 +87,18 @@ user_pref("browser.shell.checkDefaultBrowser", false);
...
- * [SETTING] Home>Firefox Home Content>...  to show/hide what you want ***/
-user_pref("browser.newtabpage.activity-stream.feeds.telemetry", false);
-user_pref("browser.newtabpage.activity-stream.telemetry", false);
-user_pref("browser.newtabpage.activity-stream.feeds.snippets", false); // [DEFAULT: false]
-user_pref("browser.newtabpage.activity-stream.feeds.section.topstories", false);
-user_pref("browser.newtabpage.activity-stream.section.highlights.includePocket", false);
-user_pref("browser.newtabpage.activity-stream.showSponsored", false);
+/* 0105: disable Discovery Stream Feed and sponsored content on Firefox Home (Activity Stream)
+ * [SETTING] Home>Firefox Home Content ***/
 user_pref("browser.newtabpage.activity-stream.feeds.discoverystreamfeed", false); // [FF66+]
```
fxbrit, line 99: I'm still not sure what this pref does but the master switch for Discovery Stream is [browser.newtabpage.activity-stream.discoverystream.enabled](https://firefox-source-docs.mozilla.org/browser/components/newtab/docs/v2-system-addon/preferences.html#browser-newtabpage-activity-stream-discoverystream-enabled).

I also found https://bugzilla.mozilla.org/show_bug.cgi?id=1657306 and no other mention of that pref in the source code, I guess we could leave it alone?

user.js
```diff
@@ -1264,6 +1259,15 @@ user_pref("browser.startup.homepage_override.mstone", "ignore"); // master switc
...
    // user_pref("layout.spellcheckDefault", 2); // 0=none, 1-multi-line, 2=multi-line & single-line
+/* FIREFOX HOME CONTENT ***/
+   // user_pref("browser.newtabpage.activity-stream.showSearch", false); // Web Search
+   // user_pref("browser.newtabpage.activity-stream.feeds.topsites", false); // Shortcuts
+   // user_pref("browser.newtabpage.activity-stream.feeds.section.topstories", false); // Recommended by Pocket
+   // user_pref("browser.newtabpage.activity-stream.feeds.section.highlights", false); // Recent Activity
+   // user_pref("browser.newtabpage.activity-stream.section.highlights.includePocket", false);
+   // user_pref("browser.newtabpage.activity-stream.section.highlights.includeBookmarks", false);
+   // user_pref("browser.newtabpage.activity-stream.section.highlights.includeDownloads", false);
+   // user_pref("browser.newtabpage.activity-stream.section.highlights.includeVisited", false);
```
fxbrit, line 1270: I guess depending on you desired level of detail these could also go

user.js
```diff
@@ -1264,6 +1259,15 @@ user_pref("browser.startup.homepage_override.mstone", "ignore"); // master switc
    // user_pref("accessibility.typeaheadfind", true); // enable "Find As You Type"
    // user_pref("clipboard.autocopy", false); // disable autocopy default [LINUX]
    // user_pref("layout.spellcheckDefault", 2); // 0=none, 1-multi-line, 2=multi-line & single-line
+/* FIREFOX HOME CONTENT ***/
+   // user_pref("browser.newtabpage.activity-stream.showSearch", false); // Web Search
+   // user_pref("browser.newtabpage.activity-stream.feeds.topsites", false); // Shortcuts
+   // user_pref("browser.newtabpage.activity-stream.feeds.section.topstories", false); // Recommended by Pocket
+   // user_pref("browser.newtabpage.activity-stream.feeds.section.highlights", false); // Recent Activity
```
fxbrit, line 1266: FYI: default false in my Nightly install

user.js
Thorin-Oakenpants: I added those last three to make it more compete as a FYI (pocket doesn't have a UI but the others do)

I'm happy to remove the last three as it means they get reset every prefscleaner - and those who use Firefox Home will have to add overrides. I would use FF Home if only it could hold something like collections, or tabs of tiles. I'm even considering making HOME and NEWTABPAGE inactive or moving to personal

currently using this - the only downside is the folders when clicked produce a popup, rather than be sticky and be a subsection
supaspeeddial

user.js
Thorin-Oakenpants: I removed it anyway

user.js
Thorin-Oakenpants: > When this is set to true the Discovery Stream experience will show up if enabled is also true on browser.newtabpage.activity-stream.discoverystream.config. Otherwise the old Activity Stream experience will be shown.

> the master switch

not a master switch if it only reverts to **_the old Activity Stream experience_** .. whatever that is

Thorin-Oakenpants: maybe we should remove the Discovery Stream pref we have as it would interfer with user's choices

- browser.newtabpage.activity-stream.discoverystream.endpoints
- the only endpoints are `https://getpocket.cdn.mozilla.net/,https://spocs.getpocket.com/`
   - I'd rather not block anything pocket: if users don't have an account but want pocket stories, that's fine (it's local), otherwise if they have a pocket account, who are we to interfere

> A list of endpoints that are allowed to be used by Discovery Stream for remote content (eg: story metadata) and configuration (eg: remote layout definitions for experimentation).

If users don't want pocket they can turn it off, so fuck it, let's remove it

also, reading that bugzilla, which was 2 years ago, they talk about the old implementation not being very well supported - e.g https://bugzilla.mozilla.org/show_bug.cgi?id=1657306#c16

fxbrit (approved):
more trimming is always a good news, LGTM :-)

opusforlife2:
What does sessionstore interval do for us, exactly? I have seen the override recipe for session restore and that seems to be the only use case.

For those who don't use session restore, can this value be set to something insanely high (or turned off somehow) so that Firefox doesn't waste time doing this?

Thorin-Oakenpants:
it does nothing for us, hence moved. it was added way back in the early days, so seven years ago? I'm not going to comment on the merits of the number of disk writes and older machines, and TBH I think those days have sailed. Privacy wise it does absolutely zero. There will always be some state written to disk, changing how often it is updated does nothing

Thorin-Oakenpants:
I'll push this tomorrow because I changed the date in the user.js to 5th Oct :) 

and I'm also lazy as fuck and I want to make releases late so people release that these changes are minor and not a reason to not update (there are stupid comparisons being made about librewolf vs arkenfox updates - one is an application, the other is a pref change or two with minimal effect these days: they're not comparable AT ALL)
- https://discuss.privacyguides.org/t/librewolf-a-fork-of-firefox-focused-on-privacy-security-and-freedom/148/28
- https://old.reddit.com/r/PrivacyGuides/comments/x9p2rw/thought_about_putting_librewolf_on_the_website/int8bxb/

from the reddit link: what a fucking moron
> I just checked, as I'm writing this Librewolf is on 104.0.2 and was updated yesterday. Arkenfox, according to their github, is on 103 and was last updated 23 days ago. So we must have different ideas of what slow is.

fxbrit:
I thought _stay away from reddit_ was rule number one, shame on you.

PR #1544, opened 2022-09-15
Add devtools.accessibility.enabled to user.js

Tiagoquix:
Removes "Inspect Accessibility Properties" from the context menu.


Thorin-Oakenpants:
this does nothing for privacy, and just adds more roadblocks for usability/devs. The original pref was added with concerns for a11y/ARIA etc (which AFAICT does not leak anywhere), and had the added bonus of perf (well documented). That perf benefit may become obsolete with [CTW](https://wiki.mozilla.org/Accessibility/CacheTheWorld)

PR #1562, opened 2022-10-11
Added 2405

wand3rlust:
Added "dom.battery.enabled" config


rusty-snake:
Did you don't any resarch and/or testing?

PR #1580, opened 2022-11-09
Format date and time consistently.

junoslukan:
Currently, in the .bat files, the date and time used for renaming backup copies of files are formatted in a way that is locale dependent.  For example, in my locale calling

```cmd
echo %date%_%time%
```

outputs `sre. 09. 11. 2022_13:07:47,71`.
This does not sort nicely and in addition includes unneeded information, such as the weekday and the hundredths of second.

In this pull request, I explicitly requested the format of date and time, which translates into consistent filenames of backup files. I used [the method suggested by an answer on StackOverflow](https://stackoverflow.com/a/203116/4341436). The format was chosen to mirror `updater.sh` , while I also included the ISO format as an alternative in a remark.

I also bumped the version of the `updater.bat` (but not `prefsCleaner.bat`) and `SET v=4.17` accordingly, which was lagging one version number behind.

### DIFF
--- a/prefsCleaner.bat
+++ b/prefsCleaner.bat
@@ -31,8 +31,11 @@ IF NOT EXIST "prefs.js" (CALL :abort "prefs.js not found in the current director
 CALL :strlenCheck
 CALL :FFcheck
 CALL :message "Backing up prefs.js..."
-SET "_time=%time: =0%"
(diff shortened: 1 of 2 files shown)


Thorin-Oakenpants:
thanks, I agree dates should be year-month-day-24hr ...  I'll ping @earthlng and when he turns up he can push this after double checking it

earthlng:
nice, thanks for the pull request!

PR #1583, opened 2022-11-12
Improve Usability by Clarifying Desired Action

iTawkins:
Running`prefsCleaner.sh` prompts the user for command selection before continuing. However, it fails to tell the user how to proceed and instead just displays the following options:
`1) Start
2) Help
3) Exit`
Common sense would dictate that the user should enter one of the commands to proceed, but the script only accepts the numerical input. For example `Start` and `start` fail to resume the script, but `1` does.

After a few tries, the user should be able to figure out how to proceed, but appending this short line (or equivalent) increases usability, specifically for inexperienced users. It's not a huge improvement, but it's a small improvement in clarity that could save a few keystrokes.


earthlng:
nice, thanks!

Issue #1587, opened 2022-11-19
Updater.sh is not updating user.js


Opening-Button-8988:
I am using Arch Linux.

When I execute the updater.sh file, it does nothing. When I run it in the terminal, it shows 

```
This script will update to the latest user.js file and append any custom configurations from user-overrides.js. Continue Y/N? 
y
Status: user.js has been backed up and replaced with the latest version!
Status: Override file appended: user-overrides.js
```

Updater.sh is in the same location as the user-overrides.js, user.js, updater.bat and everything else that was in the zip file.

I know the .sh file is doing nothing because changes I make in user-overrides.js do not reflect to user.js, which is the intended purpose, right? If there's something I'm missing, please let me know.

icpantsparti2:
Hope I am not stating the obvious, but my first thought is have you placed the updater.sh etc in your firefox profile directory (find that location by opening `about:support` page in firefox).

#### also add an entry at the end of your user-overrides.js, something like:

```
user_pref("_user.js.parrot", "SUCCESS: user-overrides.js");
```

you can check `about:config` page in firefox for that pref name when you re-open firefox after the script has run.

#### When you run the updater script look out for any error messages

```
# change directory to your firefox profile (edit as required)
cd /home/*/.mozilla/firefox/*.default
./updater.sh
```

#### Other commands that might help you see what's happening:

\# list some files (with permissions/size/date shown) (edit as required):

```
find /home/*/.mozilla -iregex "^.*/\(prefs.js\|user.js\|user-overrides.js\|updater.sh\)$" -printf '%Y %M %u:%g %#6s %TY-%Tm-%Td %TH:%TM:%.2TS %p\n'
```

\# see if overrides are showing on the end of user.js:

```
tail user.js
```

TheNightRider12:
I too am having this issue. I emptied out my user-overrides to just have 
`user_pref("_user.js.parrot", "SUCCESS: user-overrides.js");`

Ran prefsCleaner, then ran updater.sh. All with Firefox closed, all from the correct directory. When I open Firefox and go into `about:config`, there is nothing showing up for "_user.js.parrot". 

I am on Fedora Linux 37.

Thorin-Oakenpants:
> nothing is showing

well it should, because the user.js is riddled with them. You should at the very least get to `SUCCESS: No no he's not dead, he's, he's restin'!`

or do you mean it's not showing `SUCCESS: user-overrides.js`?

TheNightRider12:
[...]

PR #1595, opened 2022-11-24
updater.sh: cope if run with sudo [withdrawn see #1651 instead]

icpantsparti2:
hi @earthlng, some small additions that 'might' help with [#1587](https://github.com/arkenfox/user.js/issues/1587). Thank you.

in version 3.5 if you run updater.sh with sudo then: (1) user.js may only be readable by root and not load for the Firefox user (2) if the script is run first time with sudo, then again without sudo, permissions from the previous run may cause issues (eg no backup or update of user.js) with little or no warning.

Therefore v3.6 to: (1) set some file/dir permissions (on those created) if run with sudo (2) show info if some access permissions are not as expected (3) added some further error messages

### DIFF
--- a/updater.sh
+++ b/updater.sh
@@ -2,7 +2,7 @@
 
 ## arkenfox user.js updater for macOS and Linux
 
-## version: 3.5
+## version: 3.6
 ## Author: Pat Johnson (@overdodactyl)
 ## Additional contributors: @earthlng, @ema-pe, @claustromaniac, @infinitewarp
 
@@ -202,6 +202,7 @@ update_updater() {
   fi
   mv "${tmpfile}" "$SCRIPT_FILE"
   chmod u+x "$SCRIPT_FILE"
+  chmod go+rx "$SCRIPT_FILE"
   "$SCRIPT_FILE" "$@" -d
   exit 0
 }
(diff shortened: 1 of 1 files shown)


CelestialNebula:
(I don't know how macOS works they may do different stuff for root.)

https://stackoverflow.com/questions/18215973/how-to-check-if-running-as-root-in-a-bash-script
I can't see any reason to use sudo, so I think this is simpler (added at
the start):
```
if [ "${EUID:-$(id -u)}" -eq 0 ]
then
	printf "You shouldn't run this as root.\n"
	exit 1
fi
```

I doubt it will matter but this may catch more https://stackoverflow.com/a/65358284:
```
ls /root > /dev/null 2>&1 && { printf "You shouldn't run this as root.\n"
			       exit 1; }
```

mhmdanas:
@CelestialNebula's suggestion makes sense, you should definitely not run a command or script with `sudo` unless it's absolutely necessary. I would be especially careful with running `sudo` for anything downloaded from the internet.

@CelestialNebula I have macOS and wouldn't mind testing anything if needed.

CelestialNebula:
@mhmdanas 
I heard that macos switched to zsh and it seems that supports EUID [1] and it does more or less the same thing.

If you could test this to confirm:

test.sh:
```
#!/usr/bin/env sh

if [ "${EUID:-$(id -u)}" -eq 0 ]
then
	printf "You shouldn't run this as root.\n"
	exit 1
fi
```
`$ sudo sh test.sh`

If it worked you should see: You shouldn't run this as root.
Then run
`$ echo $?`
and it should be 1

[1] https://zsh.sourceforge.io/Doc/Release/Parameters.html#Parameters-Set-By-The-Shell

mhmdanas:
@CelestialNebula apologies for the delay. I tested the script you provided with macOS's Dash, Bash, and Zsh, and it worked just fine under all of them.

>  Look for a line with something like root:*:0:root

There is no line in my /etc/groups file that begins with `root`.

mhmdanas:
Sorry for the delay, I totally forgot about your comment. I followed your steps with both Dash, Bash, and Zsh, and it worked with all of them. Note that the output was this with all shells:
```
It looks like this script was previously run with elevated privileges,
you will need to change ownership of the following files to your user:
.//zfakefile
```

icpantsparti2:
Withdrawn in favor of [#1651](https://github.com/arkenfox/user.js/pull/1651).  Closing without further action.

Issue #1596, opened 2022-11-28
Auto-update for prefsCleaner?


umbralOptimatum:
Just noticed my prefsCleaner has been one version behind for months. I find the updater's auto-update flag very helpful as I use a script to automate my Arkenfox updates with minimum attention needed from me. Any chance of a similar flag for prefsCleaner?

heretoenhance:
I support this, would be helpful

atomGit:
@claustromaniac @overdodactyl - combining the cleaner and updater scripts might be worth considering

earthlng:
prefsCleaner v1.6 now has auto-update functionality

PR #1601, opened 2022-12-08
PrefsCleaner.sh v1.6

iTawkins:
This pull request seeks to resolve issue [#1596](https://github.com/arkenfox/user.js/issues/1596). I took the update_updater function from the updater.sh file and implemented it into the prefsCleaner script. This allows the prefsCleaner.sh script to check for updates at the user's request. This can be done in the interactive menu or by passing flag `-u`. While I was at it, I also changed the way flags were handled in the script to make it easier to add new flags in the future.

### DIFF
--- a/prefsCleaner.sh
+++ b/prefsCleaner.sh
@@ -17,6 +17,17 @@ if [ -z "$sfp" ]; then sfp=${BASH_SOURCE[0]}; fi
 ## change directory to the Firefox profile directory
 cd "$(dirname "${sfp}")"
 
+## download method priority: curl -> wget
+DOWNLOAD_METHOD=''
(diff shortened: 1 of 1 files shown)


earthlng:
Hi,

thanks for the PR!
Next time please test it first and make sure the code is functional before submitting it.
Just a quick glance and I can see that download_file and SCRIPT_FILE are not available

iTawkins:
Hello,

I tested the code multiple times and it was working as intended (I just used it to update to the real v1.6). I'm new to this whole thing, so I actually forgot to commit the updater function. Sorry  for wasting your time, but I guess it worked out as your implementation was more eloquent than mine, anyway.

PR #1616, opened 2023-01-17
prefsCleaner.bat: add -unattended flag

kewiha:
Usage:
  prefsCleaner.bat -unattended

Skips the prompt for user input and proceeds when -unattended is specified. If omitted, default behaviour is unchanged.

Code taken from updater.bat for consistency.

Please review this PR closely as I'm not fluent in batch scripting. The revised script appears to function properly with and without -unattended, but I'm not aware of common pitfalls.

Signed-off-by: Keith Harrison <keithh@protonmail.com>

### DIFF
--- a/prefsCleaner.bat
+++ b/prefsCleaner.bat
@@ -7,6 +7,8 @@ REM ## version: 2.5
 
 CD /D "%~dp0"
 
+GOTO parse
+
 :begin
 ECHO:
 ECHO:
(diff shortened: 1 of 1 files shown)


Thorin-Oakenpants:
thanks, I'll leave this for earthlng to review for when he has time

earthlng:
Thanks @kewiha 👍

Issue #1625, opened 2023-02-05
Enable manual crash reporting


trytomakeyouprivate:
Firefox crashes for me at the moment. Trying to find help I find out that Arkenfox even blocks `about:crashes` from appearing!

This makes no sense, is contraproductive and harms the project.

Just keep
```
user_pref("browser.crashReports.unsubmittedCheck.autoSubmit2", false)
```

and remove the other lines.

Users should decide on their own if they want to send crash reports, helping them make their browser work again.

Great work, but this goes too far.

Thorin-Oakenpants:
> but this goes too far

Because AF modifies **_a lot of prefs_**, it is not productive to send AF profiles crash reports to Mozilla. Reproduce it in a new profile and THEN send your crash report

What really goes too far is wasting Mozilla devs' time. The first thing they would say, if e.g. you opened a bugzilla, would be (excluding a crash report) to send your about:support profile, and one look at that and they would say , like any good STRs and debugging, would be to ask if it happens in a new profile

PR #1626, opened 2023-02-05
re-enabled manual crash reporting

trytomakeyouprivate:
As far as I understand, crash reporting should still be an active choice. But it should not be disabled without an easy way to fix this (deleting the URL)


Thorin-Oakenpants:
see reply in #1625

trytomakeyouprivate:
makes sense

PR #1637, opened 2023-03-02
Disable Firefox start page sponsored sites "contile"

trytomakeyouprivate:
The start page can actually be useful. If users may reenable it, these sites would automatically make connections

https://www.kuketz-blog.de/mozilla-firefox-datensendeverhalten-desktop-version-browser-check-teil20/

### DIFF
--- a/user.js
+++ b/user.js
@@ -181,8 +181,8 @@ user_pref("app.normandy.api_url", "");
 
 /** CRASH REPORTS ***/
 /* 0350: disable Crash Reports ***/
-user_pref("breakpad.reportURL", "");
-user_pref("browser.tabs.crashReporting.sendReport", false); // [FF44+]
+user_pref("breakpad.reportURL", "https://www.mozilla.org/legal/privacy/firefox.html#crash-reporter");
+user_pref("browser.tabs.crashReporting.sendReport", true); // [FF44+]
    // user_pref("browser.crashReports.unsubmittedCheck.enabled", false); // [FF51+] [DEFAULT: false]
 /* 0351: enforce no submission of backlogged Crash Reports [FF58+]
  * [SETTING] Privacy & Security>Firefox Data Collection & Use>Allow Firefox to send backlogged crash reports  ***/
--- a/user.js
+++ b/user.js
@@ -363,6 +363,10 @@ user_pref("browser.formfill.enable", false);
  * [4] https://earthlng.github.io/testpages/visited_links.html (see github wiki APPENDIX A on how to use)
  * [5] https://lcamtuf.blogspot.com/2016/08/css-mix-blend-mode-is-bad-for-keeping.html ***/
    // user_pref("layout.css.visited_links_enabled", false);
+/* 0821: disable connection to sponsored sites on Firefox start page
+ * [1] https://www.kuketz-blog.de/mozilla-firefox-datensendeverhalten-desktop-version-browser-check-teil20/ ***/
+user_pref("browser.topsites.contile.enabled", false);
+
 
 /*** [SECTION 0900]: PASSWORDS
    [1] https://support.mozilla.org/kb/use-primary-password-protect-stored-logins-and-pas


Thorin-Oakenpants:
I don't read german. Anyway, looks like [5]. I'm not going to waste any time on this, but looks like FF would contact a mozilla server with I guess your region, in order to provide a list of tiles for your region. Is that it? What happens if you disable suggestions/sponsors in Activity Stream as per all the UI settings. Where is the privacy concern. Closing, until info providing

trytomakeyouprivate:
no the setting stops pages from connecting to their servers to load content. This only affects sponsored sites.

I only know that article, Firefox translate works well.

I would disable sponsored tabs on default, because people might like the start page, but is has to be debloated. Pocket is already disabled.

Thorin-Oakenpants:
connecting to mozilla servers to get a curated basic list of sites is not a privacy issue

debloated !== privacy. Please state what the privacy issue is

trytomakeyouprivate:
The contiles are the sponsored sites added to your start screen. If a user would enable it, these would contact the sponsored sites, e.g. tracking.

I dont know, is disabling the startscreen hardcoded? So one would have to change the user.js

But anyways, disabling it still useful, to guarantee privacy even if one wants to use the startscreen for their workflow, there is nothing bad about it. This setting will also not break anything in the future I guess.

trytomakeyouprivate:
literal quote from the website, using Firefox Translate : 

> [5] Connect to Mozilla to the host »contile.services.mozilla.com:
> ```
> GET /v1/tiles HTTP/2
> Host: contile.services.mozilla.com
> User Agent: Mozilla/5.0 (Windows NT 10.0; Win64; x64; rv:94.0) Gecko/201001 Firefox/94.0
> Accept: */*
> Accept-Language: en,en-US;q-0.7,en;q-0.3
> Accept-Encoding: gzip, deflate
> Sec-Fetch-Dest: empty
> Sec-Fetch-Mode: cors
> Sec-Fetch-Site: cross-site
> Te: trailers
> ```
> The remote station "contile.services.mozilla.com" is the [Contile Tile Server](https://github.com/mozilla-services/contile) from Mozilla. Background is the homepage or Tile view where tiles are loaded from Amazon, eBay, YouTube, Facebook, Wikipedia and Reddit. A server response looks like this (example Amazon):
> ```
> "id":74357,
> "name":"Amazon",
> "url":"https://www.amazon.de/?tag-admpdesktopde-21&ref-pd-sl-a70A23FDA6F3F4DC8F614AB, 
> "click-url":"https://bridge.sfo1.ap01.net/ctp?version-16.0.0&ci-1636359241037.12791&key-1636924003000000&ctag-7ED3A23FDA6F4DC8F614AB,"
> "image-url":"https://contile-images.services.mozilla.com/obgoOYObjIFea-bXuT5LbBJ8j425AD87S1HMD3BWg.9991.jpg,"
> "image size":200,
> "impression-url":"https://imp.mt48.net/static?id .7RHzfOIWHG7kJngfivnEnJrNWxnwmHF4%2Bj3%3DTIGfnxrxrEOCXfplfFIZJF8GF%2Bjjr3NHOIZjFXgFwqgCxkfZDr7n4NJEFYU5FwqgC8XfpqWfYIksGew5FwqgC8y,"
> "position":1
> ```
> With a call of the [about:config](https://support.mozilla.org/de/kb/konfigurationseditor-fur-firefox) and the following setting, the call can be prevented: browser.topsites.contile.enabled = false

trytomakeyouprivate:
So yes it seems Firefox proxies these partners, so the data is sent from their servers and not theirs. Still I think its useful to deactivate

Thorin-Oakenpants:
Mozilla already has app update checks, extension update checks, checks from other moz services for items like blocklists, crlite etc ... there is no PII and Mozilla are not interested in your IP - all it says is like the other 220mn FF users - hey update me, this IP has a Firefox - big deal. Even updating Safe Browsing lists via google is fine - there is no entropy or privacy issue here.

The same applies to this sponsored tile connection - all 220mn Firefox users with AS by default will be asking for a curated (maybe it's regional) set of tiles - big deal. There is no PII here.

If you want to use Activity Stream, there a whole settings page where you can tweak, including turning off sponsored items - I am not interested in duplication - we set newtab/hometab as blank by default. You still haven't even SHOWN me that if you use AS and you disable the sponsored items in the UI that the connection is still made

Stop conflating that all outgoing requests == bad for privacy. SHOW me that there is a privacy issue and that the user can't deal with this non-privacy issue on their own

trytomakeyouprivate:
okay fair point that it is not really a threat or anything. I just think it would be nice to have the backup. A user.js allowing users to change some settings and still be safe would be great, especially as without addons like Tabliss you have no start page with Arkenfox.

It is no privacy concern, but unnessecary bloat. I guess this connection will not be made when disabled, but currently have no Firewall to test that.

PR #1638, opened 2023-03-04
5007: adds browser.sessionstore.max_windows_undo

wavygecko:
Adds flag for preventing reopening closed windows in new session


Thorin-Oakenpants:
this was removed in AF v64 - see https://github.com/arkenfox/user.js/issues/575

PR #1651, opened 2023-04-01
updater.sh/prefsCleaner.sh: Check for root and abort

CelestialNebula:
Check if running as root and if any files have the owner/group as root|wheel. Abort on both.

Should (hopefully) prevent stuff like: https://github.com/arkenfox/user.js/issues/1587
Discussion: https://github.com/arkenfox/user.js/pull/1595

This does not attempt to correct the permissions, just notify the user
and abort.

If I followed https://github.com/arkenfox/user.js/issues/1587 right.
The original person didn't realize that it appended and subsequent
people tried to run with elevated privileges.  This should at least fix
the second.

### DIFF
--- a/prefsCleaner.sh
+++ b/prefsCleaner.sh
@@ -15,7 +15,7 @@ if [ "${EUID:-$(id -u)}" -eq 0 ]; then
 elif [ -n "$(find ./ -user 0 -o -group 0)" ]; then
 	printf 'It looks like this script was previously run with elevated privileges,
 you will need to change ownership of the following files to your user:\n'
-	find ./ -user 0 -o -group 0
+	find . -user 0 -o -group 0
 	exit 1
 fi
 
--- a/prefsCleaner.sh
+++ b/prefsCleaner.sh
@@ -2,11 +2,22 @@
 
 ## prefs.js cleaner for Linux/Mac
 ## author: @claustromaniac
-## version: 1.6
+## version: 1.7
 
 ## special thanks to @overdodactyl and @earthlng for a few snippets that I stol..*cough* borrowed from the updater.sh
 
 ## DON'T GO HIGHER THAN VERSION x.9 !! ( because of ASCII comparison in update_prefsCleaner() )
+
+# Check if running as root and if any files have the owner/group as root/wheel.
+if [ "${EUID:-$(id -u)}" -eq 0 ]; then
+	printf 'You should not run this with elevated privileges (such as with doas/sudo).\n'
+	exit 1
+elif [ -n "$(find ./ -user 0 -o -group 0)" ]; then
+	printf 'It looks like this script was previously run with elevated privileges,
+you will need to change ownership of the following files to your user:\n'
+	find ./ -user 0 -o -group 0
+	exit 1
+fi
 
 readonly CURRDIR=$(pwd)
 
@@ -138,7 +149,7 @@ echo -e "\n\n"
 echo "                   ╔══════════════════════════╗"
 echo "                   ║     prefs.js cleaner     ║"
 echo "                   ║    by claustromaniac     ║"
-echo "                   ║           v1.6           ║"
+echo "                   ║           v1.7           ║"
 echo "                   ╚══════════════════════════╝"
 echo -e "\nThis script should be run from your Firefox profile directory.\n"
 echo "It will remove any entries from prefs.js that also exist in user.js."
--- a/prefsCleaner.sh
+++ b/prefsCleaner.sh
@@ -10,7 +10,7 @@
 
 # Check if running as root and if any files have the owner/group as root/wheel.
 if [ "${EUID:-"$(id -u)"}" -eq 0 ]; then
-	printf 'You should not run this with elevated privileges (such as with doas/sudo).\n'
+	printf 'You shouldn't run this with elevated privileges (such as with doas/sudo).\n'
 	exit 1
 elif [ -n "$(find ./ -user 0 -o -group 0)" ]; then
 	printf 'It looks like this script was previously run with elevated privileges,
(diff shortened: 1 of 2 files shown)


prefsCleaner.sh
```diff
@@ -2,12 +2,23 @@
...
 ## DON'T GO HIGHER THAN VERSION x.9 !! ( because of ASCII comparison in update_prefsCleaner() )
 
+# Check if running as root and if any files have the owner/group as root/wheel.
+if [ "${EUID:-$(id -u)}" -eq 0 ]; then
+	printf 'You should not run this with elevated privileges (such as with doas/sudo).\n'
+	exit 1
+elif [ -n "$(find ./ -user 0 -o -group 0)" ]; then
+	printf 'It looks like this script was previously run with elevated privileges,
+you will need to change ownership of the following files to your user:\n'
+	find ./ -user 0 -o -group 0
```
mhmdanas, line 18:
```suggestion
	find . -user 0 -o -group 0
```

With `./` the output is shown with a double slash for me, which looks weird.

updater.sh
```diff
@@ -2,12 +2,23 @@
...
 ## DON'T GO HIGHER THAN VERSION x.9 !! ( because of ASCII comparison in update_updater() )
 
+# Check if running as root and if any files have the owner/group as root/wheel.
+if [ "${EUID:-$(id -u)}" -eq 0 ]; then
+	printf 'You should not run this with elevated privileges (such as with doas/sudo).\n'
+	exit 1
+elif [ -n "$(find ./ -user 0 -o -group 0)" ]; then
+	printf 'It looks like this script was previously run with elevated privileges,
+you will need to change ownership of the following files to your user:\n'
+	find ./ -user 0 -o -group 0
```
mhmdanas, line 18: Ditto:
```suggestion
	find . -user 0 -o -group 0
```

prefsCleaner.sh
```diff
@@ -2,12 +2,23 @@
...
 ## author: @claustromaniac
-## version: 1.6
+## version: 1.7
 
 ## special thanks to @overdodactyl and @earthlng for a few snippets that I stol..*cough* borrowed from the updater.sh
 
 ## DON'T GO HIGHER THAN VERSION x.9 !! ( because of ASCII comparison in update_prefsCleaner() )
 
+# Check if running as root and if any files have the owner/group as root/wheel.
+if [ "${EUID:-$(id -u)}" -eq 0 ]; then
```
mhmdanas, line 12:
```suggestion
if [ "${EUID:-"$(id -u)"}" -eq 0 ]; then
```

prefsCleaner.sh
```diff
@@ -2,12 +2,23 @@
...
+	printf 'You should not run this with elevated privileges (such as with doas/sudo).\n'
+	exit 1
+elif [ -n "$(find ./ -user 0 -o -group 0)" ]; then
+	printf 'It looks like this script was previously run with elevated privileges,
+you will need to change ownership of the following files to your user:\n'
+	find . -user 0 -o -group 0
```
mhmdanas, line 18: Is there a way to avoid running `find . -user 0 -o -group 0` twice?

CelestialNebula: You could do:
```
pc_ROOT_ID="$(find . -user 1001 -o -group 0)"
if [ "${EUID:-"$(id -u)"}" -eq 0 ]; then
	printf 'You should not run this with elevated privileges (such as with doas/sudo).\n'
	exit 1
elif [ -n "${pc_ROOT_ID}" ]; then
	printf 'It looks like this script was previously run with elevated privileges,
you will need to change ownership of the following files to your user:\n'
	printf '%s\n' "${pc_ROOT_ID}"
	exit 1
fi
```
but I think this makes the code more indirect for no real gain.

earthlng:
good idea, thanks!

updater.sh
```diff
@@ -10,7 +10,7 @@
 
 # Check if running as root and if any files have the owner/group as root/wheel.
 if [ "${EUID:-"$(id -u)"}" -eq 0 ]; then
-	printf 'You should not run this with elevated privileges (such as with doas/sudo).\n'
+	printf 'You shouldn't run this with elevated privileges (such as with doas/sudo).\n'
```
rusty-snake, line 13: This needs to be escaped/quoted, otherwise it is a syntax error breaking the script.

rusty-snake:
```suggestion
	printf 'You shouldn'"'"'t run this with elevated privileges (such as with doas/sudo).\n'
```

dngray:
btw, this is still incorrect and broken.

```
line 13: syntax error near unexpected token `('
line 13: `	printf 'You shouldn\'t run this with elevated privileges (such as with doas/sudo).\n''
```

Really ought to run these shell scripts through [Shellcheck](https://www.shellcheck.net/) as well and fix the other errors.

prefsCleaner.sh
```diff
@@ -2,12 +2,23 @@
...
 ## special thanks to @overdodactyl and @earthlng for a few snippets that I stol..*cough* borrowed from the updater.sh
 
 ## DON'T GO HIGHER THAN VERSION x.9 !! ( because of ASCII comparison in update_prefsCleaner() )
 
+# Check if running as root and if any files have the owner/group as root/wheel.
+if [ "${EUID:-"$(id -u)"}" -eq 0 ]; then
+	printf 'You shouldn't run this with elevated privileges (such as with doas/sudo).\n'
+	exit 1
+elif [ -n "$(find ./ -user 0 -o -group 0)" ]; then
+	printf 'It looks like this script was previously run with elevated privileges,
```
dngray, line 16:
```suggestion
	printf "It looks like this script was previously run with elevated privileges,
        you will need to change ownership of the following files to your user:\n"
```

Don't use single quotes.

PR #1652, opened 2023-04-07
Add prefs enabling Encrypted Client Hello (ECH)

KOLANICH:


fxbrit:
these prefs are not even in early beta, I'm not sure we want to flip them in any case

rusty-snake:
Dies anyone has a list oft Servers supporting ECH? IIRC it isn't used in real world yet.

KOLANICH:
>Does anyone has a list oft Servers supporting ECH? IIRC it isn't used in real world yet.

IRL I see no servers supporting it. Cloudflare was saying it is already using ECH, but when I connect to websites behind Cloudflare I see no ECH TLS extensions in Wireshark. It is because ECH requires additional records within DNS, but noone bothers to set them up. Also several states (i.e. Russian Federation and People Republic of China) have claimed they will block any website using ECH. While I don't see Russia blocking it currently, it is likely because noone uses it IRL. I guess website owners and hosters will en-masse avoid any moves that will make them blocked.

Issue #1667, opened 2023-04-23
Fix did not fix the syntax error in updater.sh


phroi:
There is a syntax error in the [newest updater.sh](https://github.com/arkenfox/user.js/blob/7a1d0a92af95d152349830d027648fa299b9bcba/updater.sh#L13):
```
user@host:~$ cd ~/.mozilla/firefox/; ./updater.sh -s -b -p xumnlhr8.Default/
./updater.sh: line 13: syntax error near unexpected token `('
./updater.sh: line 13: `	printf 'You shouldn\'t run this with elevated privileges (such as with doas/sudo).\n'
```
Cheers, Phroi

phroi:
Or maybe print [`should not` instead of `shouldn't`](https://github.com/arkenfox/user.js/pull/1665)? 🧐

PR #1668, opened 2023-04-23
Fix escape character/syntax error in updater.sh

bol0gna:
Sorry to add yet another updater.sh PR to the pile, this should get rid of the syntax error from #1667

### DIFF
--- a/updater.sh
+++ b/updater.sh
@@ -2,7 +2,7 @@
 
 ## arkenfox user.js updater for macOS and Linux
 
-## version: 3.7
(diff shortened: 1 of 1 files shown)


adrian5:
Using double quotes, you no longer need to escape the single quote:

```sh
printf "You shouldn\'t run this with elevated privileges (such as with doas/sudo).\n"
                   ↑
```

earthlng:
THX! sorry for repeatedly fucking this up ;)

Issue #1142, opened 2021-03-14
question: network.protocol-handler.external.ms-windows-store


crssi:
What exactly is a reason for `network.protocol-handler.external.ms-windows-store` to be active and different as is default?

Thank you and cheers

Thorin-Oakenpants:
so that is `disable links launching Windows Store on Windows 8/8.1/10`

Do you really want a link for ilovetaylorswift.com opening up somewhere else other than your default browser?

github-account1111:
>If you click on it in a browser on Windows 10, the games' page on the Microsoft Store websites gets opened, and so does the Windows Store application and the game's page there.

It doesn't though.
Nothing will open until you click Get, at which point you quite obviously want to open Microsoft Store and install the app.

Thorin-Oakenpants:
https://bugzilla.mozilla.org/show_bug.cgi?id=1756209

Thorin-Oakenpants:
Note that windows store is one of a very few handlers that have no warning dialogs
- see https://searchfox.org/mozilla-central/source/browser/app/profile/firefox.js#975-982

also FYI
- FF101+: [1756209](https://bugzilla.mozilla.org/show_bug.cgi?id=1756209) Support attribution from the Microsoft Store
- this is a nothing burger, a campaign id is not PII and no-one cares, especially me since we nix the handler anyway

---

So this was added in FF24: [866065](https://bugzilla.mozilla.org/show_bug.cgi?id=866065) 

Which leads to bugzilla [1677753](https://bugzilla.mozilla.org/show_bug.cgi?id=1677753) Show permission prompt before opening macOS or windows app store via external protocol.

That bugzilla is confusing - it shows that there is a permission prompt, so IDFK

Anyway the pref stays as is

Thorin-Oakenpants:
url: https://www.microsoft.com/en-nz/p/monster-go/9nblggh2s4gt?rtc=1&activetab=pivot:overviewtab
pref: `network.protocol-handler.external.ms-windows-store`
os: windows 11
test: click `get`

pref true (default) - opens MS store (I have no ms account, btw) - no prompting
pref false (AF) - nothing happens

If you're on windows, I guess you can just open edge IMO

Thorin-Oakenpants:
then again, I agree, you wanted to install it .. so yup, let's remove that pref

PR #1703, opened 2023-08-14
v115.1

Thorin-Oakenpants:
-  network.protocol-handler.external.ms-windows-store see #1142

### DIFF
--- a/user.js
+++ b/user.js
@@ -1,7 +1,7 @@
 /******
 *    name: arkenfox user.js
-*    date: 29 August 2023
-* version: 116
+*    date: 27 August 2023
+* version: 115
 *     url: https://github.com/arkenfox/user.js
 * license: MIT: https://github.com/arkenfox/user.js/blob/master/LICENSE.txt
 
--- a/user.js
+++ b/user.js
@@ -571,9 +571,6 @@ user_pref("media.peerconnection.ice.default_address_only", true);
 /* 2020: disable GMP (Gecko Media Plugins)
  * [1] https://wiki.mozilla.org/GeckoMediaPlugins ***/
    // user_pref("media.gmp-provider.enabled", false);
-/* 2021: disable widevine CDM (Content Decryption Module)
- * [NOTE] This is covered by the EME master switch (2022) ***/
-   // user_pref("media.gmp-widevinecdm.enabled", false);
 /* 2022: disable all DRM content (EME: Encryption Media Extension)
  * Optionally hide the setting which also disables the DRM prompt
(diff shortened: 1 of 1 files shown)


Thorin-Oakenpants:
deleted comment

> As I just read the corresponding source code and value 2 uses browser.download.dir (not browser.download.lastDir) I suggest the following purely cosmetic update:
> ```js
> /* 5016: discourage downloading to default location
>  * 0=desktop, 1=downloads (default), 2=custom
>  * [SETTING] To set your default "downloads": General>Downloads>Save files to ***/
>    // user_pref("browser.download.folderList", 2);
> ```

OK. so [here](https://searchfox.org/mozilla-central/source/browser/components/preferences/main.js#3391) it does say it will check the custom download dir. But testing shows it's not that simple.

My custom dir is set as `D:/Work`, folderList = 2, and I prompt all downloads to ask where to save. In a new session the first download asks where to save and starts with `D:/Work` which I changed to `D:\SomewhereElse\`. The next download prompts and starts the save dialog at `D:\SomewhereElse\`. In a new session we start over. So technically `last used` isn't 100% correct.

IDK what happens if `browser.download.dir` is not set

user.js
Thorin-Oakenpants, line 1: I meant 5016

Thorin-Oakenpants: Is there something wrong with I added? Not every file type is prompt, users can autosave some, all or none, it depends on if they have checked the save dialog box. IF you have modified the last directory used, then that is what is used, even if autosaving

Thorin-Oakenpants: read my earlier comment, test for yourself .. it uses the last dir despite what your custom download dir is, if you changed it in the session

Thorin-Oakenpants: > Yes, that happens no matter if `browser.download.folderList` is set to ... 2

so the description is correct then :)

Thorin-Oakenpants: > is set to 0, 1 or 2

now that's not something I ever bothered with. There is no need to be snippy. I have far more important things to do than piss around with a slightly vague description. Don't get me wrong, I appreciate it corrections. 

AF uses prompt by default. The original description was perfectly fine if not technically 100% correct. I mean I could have last saved in X (browsed to a different dir) but on a new session it would start off prompting in Y (custom). Hence the last edit.

The point is, I'm trying to as helpful as possible for our users. If I remove "last used", I bet someone else will come along and say .. "BUT, if I changed my last save dir, this is no longer true. Here is a PR"

I'll see what I can do - thanks

Thorin-Oakenpants: I know it's two prefs .. I eat prefs for breakfast. I was describing the behavior. Anyway, now you've _ghosted_ me :)

PR #1752, opened 2023-10-27
Disabled block_mozAddonManager, very outdated

trytomakeyouprivate:
All the linked issues are fixed since v60.0 (!)

It is intentional that Addons cannot run on AMO, to avoid addons injecting malicious .xpi downloads and tricking users into installing ANYTHING.

I am wondering if there are more of these extremely outdated settings...

### DIFF
--- a/user.js
+++ b/user.js
@@ -761,8 +761,10 @@ user_pref("privacy.window.maxInnerWidth", 1600);
 user_pref("privacy.window.maxInnerHeight", 900);
 /* 4503: disable mozAddonManager Web API [FF57+]
  * [NOTE] To allow extensions to work on AMO, you also need 2662
- * [1] https://bugzilla.mozilla.org/buglist.cgi?bug_id=1384330,1406795,1415644,1453988 ***/
-user_pref("privacy.resistFingerprinting.block_mozAddonManager", true);
+ * [1] https://bugzilla.mozilla.org/buglist.cgi?bug_id=1384330: FIXED in 57.0
+ * [2] 1406795,1415644 allowed addons Access to AMO and install malicious addons, FIXED in 60.0
+ * [3] 1453988 forbids Addons to run on AMO, intentional against malicious addons***/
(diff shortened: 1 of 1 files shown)


Thorin-Oakenpants:
> All the linked issues are fixed since v60.0 (!)

That's not how it works. The links are to provide documentation - their state doesn't matter, and what they do could be "good" OR "bad"

> It is intentional

Of course it is, it's one of the bugzillas linked. Do you think I just link random bugzillas and don't read them?

> I am wondering if there are more of these extremely outdated settings...

Did you read the user.js?
- `4502` (which was added for Tor Browser and I was there!!) in arkenfox is active
   - ^ it requires `2662` which is inactive
      -^ Do you really think I would compromise security by default?

Do you really think the pref is outdated? Hint - it is NOT, it still works exactly as expected

trytomakeyouprivate:
To make sure I understood this right, "blockmozaddonmanager" allows addons to access mozillas protected websites like AMO.

There are two reasons here, pro (Mozilla is a horrible tracking company) and contra (Some random addon could manipulate links and install malicious ones). 

This has to be evaluated, there is no clear yes or no? I would say security is most important, and this is a basic protection mechanism. But others may say "dont install any addons apart from xyz" which could also be "the Arkenfox way". But its not simple, and for sure needs some clearer documentation instead of, again, enumerating some bugzilla reports that vary

iam-py-test:
As I understand it (and I could be wrong), this does not give extensions access to the AMO, it just blocks the AMO from access certain APIs.
Thanks

Thorin-Oakenpants:
> Mozilla is a horrible tracking company

bullshit - they are not in the tracking business - getting whatever analytics (engagement/usage) for their own platform(s) e.g. AMO is not fucking tracking

> "blockmozaddonmanager" allows addons to access mozillas protected websites

no it does not - read what I already said, read the user.js. it requires ANOTHER pref

PR #1754, opened 2023-10-27
added formfill bugs

trytomakeyouprivate:
The original bug is very old and dead, maybe its still reproducible but I couldnt do it

But there are two new bugs, one about the clarification that the formdata will not be separated by URL origin, which could be a security risk, and another one about some weird bug with wrong input, when using some weird keyboard.

If this is such a vulnerable component, its probably best to keep it off

### DIFF
--- a/user.js
+++ b/user.js
@@ -337,7 +337,9 @@ user_pref("browser.urlbar.weather.featureGate", false); // [FF108+] [DEFAULT: fa
  * [NOTE] We also clear formdata on exit (2811)
  * [SETTING] Privacy & Security>History>Custom Settings>Remember search and form history
  * [1] https://blog.mindedsecurity.com/2011/10/autocompleteagain.html
- * [2] https://bugzilla.mozilla.org/381681 ***/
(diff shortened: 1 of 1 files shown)


Thorin-Oakenpants:
> The original bug is very old and dead, maybe its still reproducible but I couldnt do it

We don't deal in maybes

Thorin-Oakenpants:
see https://github.com/arkenfox/user.js/wiki/3.2-Overrides-[Common]#-common where I link to some bugzillas

There are a whole raft of form (and search is bundled with the pref) issues with a mix of usability vs security vs privacy issues. It's complex to say the least. And there will be edge cases which aren't really noteworthy - such as the virtual keyboard one. The current references are sufficient

trytomakeyouprivate:
? This PR literally only added bugs to the original user.js, not any random override. As the argumentation of the current disabling is based on 10 years old bugs, which is very weird and confusing to users. took some time to find out all the other and actually still applying bugs, so having them here is crucial in my opinion. Why close it and not just add actual documentation? Not sure about the formatting though

PR #1767, opened 2023-11-21
Include Pixiv in XOriginPolicy site breakage.

I2pRandom:
XOriginPolicy setting results in artworks on Pixiv not loading.


Thorin-Oakenpants:
thanks, but I already have examples and the pref is tagged and listed in common items in the wiki

PR #1822, opened 2024-04-09
user.js: fix section numbering + spacing

glitsj16:


SkewedZeppelin (approved):

Thorin-Oakenpants:
thanks - i'll fixup the wrong number in the other patch. We only do patches per release (when we need one)

the bugzilla number is meant to be indented an extra space so the numbers are right aligned and the `- descriptions` are left aligned and line up - like a spreadsheet - for easy visual parsing

PR #1843, opened 2024-05-22
Fix pref reference

sonbua:


Thorin-Oakenpants:
thanks ... 🍕 🧀 🍺 . It's easier for me to just add this to https://github.com/arkenfox/user.js/pull/1816 for the next major update

PR #1892, opened 2024-09-05
v133

Thorin-Oakenpants:


Thorin-Oakenpants:
> Are we going to add `browser.startup.windowsLaunchOnLogin.enabled` as `false` for `1005`?

why? https://github.com/arkenfox/user.js/issues/1860

Thorin-Oakenpants:
> it seems Arkenfox User.js has this idea of never reusing previously-removed numbers.

well, we have - done a few re-numberings of sections at times, and old numbers have either been repurposed or old ones since removed re-used. We just happen to have two `4511`'s at the same time (I want to keep the two system color prefs next to each other) - the ESR128 deprecated section will get removed _next_ PR (I could remove it now)

Issue #1696, opened 2023-07-29
Linux auto updater (update when Firefox starts)


ShellCode33:
Dear fellow Linux users,

I think we can all agree that running the updater script manually is cumbersome.
I tend to forget things, and updating the user.js is for sure on my top 5 of things to forget.
That's why I try to automate as much stuff as possible and arkenfox user.js is no exception.
Allow me to share with you a small hack I did to automate everything.

First of all, and this is very important, my PATH variable looks like this:

```
/usr/local/bin:/usr/bin:/home/shellcode/.local/bin
```
Notice how `/usr/local/bin` is **before** `/usr/bin`.

Then I created a `firefox` script in `/usr/local/bin` with the following content:

<details>
  <summary><b>[CLICK TO SHOW SCRIPT]</b></summary>

```python
#!/usr/bin/python3

import os
import re
import sys
import os.path
import requests
from typing import List, Tuple
from configparser import ConfigParser

FIREFOX_DIR = os.environ["HOME"] + "/.mozilla/firefox"
USER_JS_OVERRIDES_PATH = f"{FIREFOX_DIR}/user-overrides.js"

USER_JS_URL = "https://raw.githubusercontent.com/arkenfox/user.js/master/user.js"
USER_PREF_REGEX = r'user_pref\("(?P<pref_key>.+)"'

def get_profiles() -> List[Tuple[str, str]]:

    profiles = []

    config = ConfigParser()
    config.read(FIREFOX_DIR + "/profiles.ini")
    
    for section in config.sections():
        if section.startswith("Profile"):
            name = config[section]["Name"]
            path = config[section]["Path"]
            profiles.append((name, path))

    return profiles

def install(profile_path: str, new_user_js_content: str) -> None:
    
    user_js = f"{profile_path}/user.js"

    with open(user_js, "w") as stream:
        stream.write(new_user_js_content)

        with open(USER_JS_OVERRIDES_PATH, "r") as stream_override:
            stream.write("\n\n// ------ OVERRIDES START HERE ------\n\n")
            stream.write(stream_override.read())

def clean(profile_path: str) -> None:
    """
    Remove all entries from prefs.js that are in user.js regardless of whether they are active or not.
    They will be set back from user.js next time Firefox starts.

    Firefox must be closed for this to work because prefs.js is overwritten on exit.
    """
    user_js = f"{profile_path}/user.js"
    prefs_js = f"{profile_path}/prefs.js"

[...]

PR #1908, opened 2024-10-17
Shell script rewrite to comply with POSIX and best practices

### LINKED ISSUE #1810: updater.sh returns with exit code 0 when failing in update_userjs
updater.sh returns with exit code 0 when failing to download user.js file, for example, when no internet connection. The assumed result would be to return a non-zero exit code. This way, if the updater fails, one can retry after some time.
One way to solve would be by explicitly returning 0 at the end of update_userjs function when it completes succesfully and calling update_userjs || exit 1 to exit when the update_userjs function returns non-zero value.

### LINKED ISSUE #1855: Make Updater.sh shell agnostic
My distros use busybox ash and the Updater.sh is using bash to work. I think making shell agnostic is better for just works in all systems.

Wha i expect is it just run as normal and i get this instead.

`env: can't execute 'bash': No such file or directory`

9ao9ai9ar:
# The shell scripts in my now archived fork have some flaws that I would not be addressing here and now. Please use the shell scripts included in the main repository and only fall back to this experimental version if all else failed.

Resolves #1855
Resolves #1446
Fixes #1810

This is not the first POSIX solution to have ever been requested or presented, though certainly the most comprehensive one yet. Regrettably, the robustness goal hasn't truly been realized, so it would need to be addressed in a separate pull request. I'd also like to point out that alternatives like the one in #1696 exist and are worth considering.

## Target Platforms for Modern Desktop Versions of Firefox

Based on the [supported build hosts and targets of Firefox](https://firefox-source-docs.mozilla.org/contributing/build/supported.html).

### Tier-1

- [x] <s>Windows</s> (WSL, Cygwin, MSYS2, etc.) - Not targeted by the shell scripts.
- [x] macOS - Automated testing only.
- [x] Fedora
- [x] Debian
- [x] Ubuntu

### Tier-2

- [x] openSUSE
- [x] Arch
- [x] Alpine
- [x] NixOS - Requires the `psmisc` or `lsof` package for prefsCleaner.sh to work.

### Tier-3

- [x] FreeBSD
- [x] OpenBSD
- [x] NetBSD
- [x] Solaris - Version 11 and above only.

### Non-Tiered

- [x] DragonFly
- [x] OpenIndiana
- [x] Haiku (Iceweasel)

## POSIX Compliant Shells

- [x] bash 3 (default macOS `/bin/sh`) - Automated testing only.
- [x] bash (default Linux and Haiku `/bin/sh` and user shell)
- [x] dash (default Debian and Ubuntu `/bin/sh`)
- [x] ash (default BusyBox, FreeBSD and NetBSD `/bin/sh` and user shell)
- [x] ksh (default SunOS `/bin/sh` and user shell)
- [x] oksh (default OpenBSD `/bin/sh` and user shell)
- [x] zsh (default macOS user shell) - Requires `emulate sh` before sourcing.
- [x] XPG4 sh (alternative Solaris POSIX `sh`)
- [x] pdksh (alternative NetBSD user shell)
- [x] mksh
- [x] yash
- [x] posh
- [x] gwsh
- [x] bosh - Known issue: [Alias is messing with the case statement pattern which results in a syntax error](https://codeberg.org/schilytools/schilytools/issues/94).
- [x] osh - Known issue: https://github.com/oils-for-unix/oils/issues/2228.
- [x] <s>hush</s> - Not targeted for the time being due to missing crucial features: `command -p` and the `noglob` shell option.
- [x] <s>gash</s> - Not targeted for the time being due to missing crucial feature: `command -p`.
- [x] <s>mrsh</s> - Not targeted for the time being due to many deficiencies, e.g. https://github.com/emersion/mrsh/issues/196 and https://github.com/emersion/mrsh/issues/204.

### DIFF
--- a/prefsCleaner.sh
+++ b/prefsCleaner.sh
@@ -1,185 +1,1129 @@
-#!/usr/bin/env bash
+#!/bin/sh
 
-## prefs.js cleaner for Linux/Mac
-## author: @claustromaniac
-## version: 2.1
+# prefs.js cleaner for macOS, Linux and Unix-like operating systems
+# authors: @claustromaniac, @earthlng, @9ao9ai9ar
+# version: 3.0
 
-## special thanks to @overdodactyl and @earthlng for a few snippets that I stol..*cough* borrowed from the updater.sh
+# IMPORTANT: The version string must be on the 5th line of this file
+# and must be of the format "version: MAJOR.MINOR" (spaces are optional).
+# This restriction is set by the function arkenfox_script_version.
 
-## DON'T GO HIGHER THAN VERSION x.9 !! ( because of ASCII comparison in update_prefsCleaner() )
+# Example advanced script usage:
+# $ yes | tr -d '\n' | env WGET__IMPLEMENTATION=wget ./prefsCleaner.sh 2>/dev/null
+# $ TERM=dumb . ./prefsCleaner.sh && arkenfox_prefs_cleaner
 
-readonly CURRDIR=$(pwd)
+# This ShellCheck warning is just noise for those who know what they are doing:
+# "Note that A && B || C is not if-then-else. C may run when A is true."
+# shellcheck disable=SC2015
 
-## get the full path of this script (readlink for Linux, greadlink for Mac with coreutils installed)
-SCRIPT_FILE=$(readlink -f "${BASH_SOURCE[0]}" 2>/dev/null || greadlink -f "${BASH_SOURCE[0]}" 2>/dev/null)
+###############################################################################
+####                   === Common utility functions ===                    ####
+#### Code that is shared between updater.sh and prefsCleaner.sh, inlined   ####
+#### and duplicated only to maintain the same file count as before.        ####
+###############################################################################
 
-## fallback for Macs without coreutils
-[ -z "$SCRIPT_FILE" ] && SCRIPT_FILE=${BASH_SOURCE[0]}
+# Save the starting errexit shell option for later restoration.
+# Note that we do not choose the restoration method of running eval on
+# the saved output of `set +o` as that would be problematic because:
+# 1. bash turns off the errexit option in command substitutions
+#    and also does not clear errexit in a command substitution
+#    if `shopt -s inherit_errexit` is in effect:
+#    https://unix.stackexchange.com/a/383581.
+# 2. oksh fails to restore the errexit option using this method.
+# 3. oksh turns off the interactive option in command substitutions,
+#    and trying to toggle this option results in an error.
+# 4. it clutters up the command history with lots of set commands.
+# It is also more trouble than it is worth to try to work around
+# these limitations just to restore one shell option we may have changed,
+# not counting the posix and pipefail options as we want to keep them enabled.
+case $- in
+    *e*) _STARTING_SH_OPTION_ERREXIT=1 ;;
+    *) _STARTING_SH_OPTION_ERREXIT=0 ;;
+esac && {
+    # Unset function of the same name so that it is not invoked in place of
+    # the `command` regular built-in utility that we are going to be using:
+    # https://pubs.opengroup.org/onlinepubs/9699919799/utilities/V3_chap02.html#tag_18_09_01_01.
+    # In some ksh88 derivatives like the pdksh and XPG4 sh, the exit status
+    # will not be 0 if the function to unset is not already defined.
+    # In bash, there is no guarantee that the operation will succeed as:
+    # 1. readonly functions can not be unset.
+    # 2. all built-in utilities, including special built-in utilities
+    #    like unset and exit, can be overridden by functions.
+    # 3. to make matters worse, even the reserved words can be overridden
+    #    by aliases when not in POSIX mode.
+    # So we have no choice but to trust that the job is done faithfully.
+    # If secure shell scripting is ever a possibility, it is not found in bash.
+    \unset -f command && \: Suppress errexit if enabled.
+    \: Set a zero exit status for the grouping command.
+} &&
+    # https://pubs.opengroup.org/onlinepubs/9799919799/utilities/command.html
+    # The `command` built-in, when used without options, serves two purposes:
+    # 1. it causes the shell to treat the arguments as a simple command
+    #    that is not subject to alias substitution and shell function lookup.
+    # 2. it suppresses the special characteristics of the special built-ins,
+    #    so that an error does not cause a non-interactive script to abort.
+    # Though some shells do not respect point 2 above,
+    # so a prior test run in a subshell is still required.
+    # alias/unalias are not implemented in posh, hush and gash, and executing
+    # either in gash will exit the shell, so a safeguard is needed.
+    if (\command alias >/dev/null 2>&1) &&
+        (\command unalias -a 2>/dev/null); then
+        # Save the starting aliases for later restoration, then unset
+        # all aliases asap as alias substitution occurs right before parsing:
+        # https://pubs.opengroup.org/onlinepubs/9699919799/utilities/V3_chap02.html#tag_18_03_01.
+        {
+            # The `-p` option used by several implementations is absolutely
+            # required in order to properly save and restore alias definitions.
+            (\command alias -p >/dev/null 2>&1) &&
+                _STARTING_ALIASES=$(\command alias -p) ||
+                _STARTING_ALIASES=
+        } &&
+            \command unalias -a
+    else
+        _STARTING_ALIASES=
+    fi && {
+    # Enable POSIX mode. Needed for yash, as otherwise parsing will fail on
+    # non-ASCII characters if not supported by LC_CTYPE.
+    (command set -o posix 2>/dev/null) && command set -o posix
+    : Set a zero exit status for the grouping command.
+} &&
+    # Detect spoofing by external, readonly functions.
+    set -o errexit
+# In case of a variable assignment error (or any other shell error):
+# "In all of the cases shown in the table where an interactive shell
+# is required not to exit, the shell shall not perform any further processing
+# of the command in which the error occurred."
+# ―https://pubs.opengroup.org/onlinepubs/9699919799/utilities/V3_chap02.html#tag_18_08_01
+# We are being extra cautious here by checking the exit status
+# in a separate command because the words "any further processing of
+# the command in which the error occurred" are too vague to be relied on.
+# For example, if the following shell script is dot sourced in bash:
+# ```sh
+# readonly r
+# r= || echo continue in OR list
+# echo next command
+# ```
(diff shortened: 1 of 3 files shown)


MagicalDrizzle:
prefCleaner works fine for me, however updater doesn't seem to be able to append overrides.

MagicalDrizzle:
Seems to be just a broken check? This fixes it and updater seems to work fine now.
`if ! [ "$SKIPOVERRIDE" ];` > `if [ "$SKIPOVERRIDE" = false ];`

9ao9ai9ar:
> Seems to be just a broken check? This fixes it and updater seems to work fine now. `if ! [ "$SKIPOVERRIDE" ];` > `if [ "$SKIPOVERRIDE" = false ];`

Ah, that's it! The intricacies of the UNIX shell scripting language. I shouldn't have blindly followed my reading of the `test` man page without verifying, it's the innocent-looking parts that trip me up instead.

I will be generally unavailable for about a week, then I will return to seriously read the wiki and relevant docs and make the scripts more robust.

MagicalDrizzle:
Error in both scripts - you were checking for the literal file named `SCRIPT_FILE`:
```sh
// from
SCRIPT_FILE=$(preadlink "$0") && [ -f SCRIPT_FILE ] || exit 1
// to
SCRIPT_FILE=$(preadlink "$0") && [ -f $SCRIPT_FILE ] || exit 1
```

9ao9ai9ar:
@MagicalDrizzle Nice catch! My `/bin/sh` question is for esoteric OSes like NixOS and Solaris where, to my knowledge, it used to be the case that `/bin/sh` didn't exist or pointed to an ancient, almost primordial Bourne shell, respectively. You don't have to answer that question when it's clearly a typo or logic error that wouldn't even work as intended on Bash.

I've rewritten more parts of the scripts, like substituting `grep -A` with `awk` because `grep`'s outputs seem to be less deterministic and the `-A` option isn't specified in POSIX. I am not an AWK/sed expert though, so hopefully someone more knowledgeable about how to write portable and efficient AWK/sed scripts can pitch in when I release my final draft, hopefully by the end of this week.

jamesericdavidson:
@9ao9ai9ar Thoughts on using `#!/usr/bin/env`? `sh` could be busybox, the gimped version of Bash (which is *not* POSIX-compliant), etc.

updater.sh
```diff
@@ -96,312 +214,291 @@ Optional Arguments:
...
+SKIPOVERRIDE=false
+VIEW=false
+PROFILE_PATH=false
+ESR=false
+if [ $# != 0 ]; then
+    # Display usage if first argument is -help or --help
+    if [ "$1" = '--help' ] || [ "$1" = '-help' ]; then
+        usage
+        exit 0
     else
```
sertonix, line 438: This `else` could be replaced with a `fi` and then the rest could be intended less.

updater.sh
```diff
@@ -96,312 +214,291 @@ Optional Arguments:
...
+UPDATE='check'
+CONFIRM='yes'
+OVERRIDE='user-overrides.js'
+BACKUP='multiple'
+COMPARE=false
+SKIPOVERRIDE=false
+VIEW=false
+PROFILE_PATH=false
+ESR=false
+if [ $# != 0 ]; then
```
sertonix, line 433: I think this check does nothing and could be removed completely.

updater.sh
9ao9ai9ar: I have removed these non-POSIX options, but if you specify them the usage would still be displayed.

9ao9ai9ar:
> @9ao9ai9ar Thoughts on using `#!/usr/bin/env`? `sh` could be busybox, the gimped version of Bash (which is _not_ POSIX-compliant), etc.

The POSIX way to do it is to make an install script and overwrite the shebang with the value of `$(getconf PATH)/sh`, but I think that is overkill and would just ask the users to change it themselves if needed. My personal opinion is `/bin/sh` is better or at least no worse than `/usr/bin/env sh` because `/usr` might not be mounted, and the former allows specifying options to `sh` as an added bonus. I might reconsider the choice if `/usr/bin/env sh` turns out to work better for the subset of operating systems the scripts are intended to support, with real users (maybe Solaris 10?).

9ao9ai9ar:
Latest work-in-progress commit might have broken things, so disclaimer: it's for review purposes only. All feedbacks are welcome, especially those regarding my general approaches and the TODO/FIXME items.

Not sure if there's a reason behind the discrepancy, but I noticed the batch version updater calls prefsCleaner, whereas the UNIX shell version updater doesn't.

Just a heads-up: before I mark this branch as ready to be pulled, I'd squash all commits into one with thorough commit messages, because as things stand now the commit diffs are not that helpful and would only waste time for reviewers.

aartoni:
Hey @9ao9ai9ar @MagicalDrizzle I've noticed that progress on this PR is slowing down, so I'm willing to create a checklist of things to ensure that the script is working properly, and then make a GitHub workflow to test the script on different platforms.

Here are some things to be checked:
- [ ] normal execution returns `EX_OK`
- [ ] check every non-zero `EX_*`
- [ ] `user.js` gets updated
- [ ] `user-overrides.js` is respected

This will tick the [GitHub available runners](https://docs.github.com/en/actions/writing-workflows/choosing-where-your-workflow-runs/choosing-the-runner-for-a-job#standard-github-hosted-runners-for-public-repositories) off the list of things that testers have to check manually. Manual tester can then copy the workflow to test locally.

What do you think? An alternative would be splitting this PR into smaller PRs.

9ao9ai9ar:
@aartoni Sorry I had been procrastinating badly, but have picked up pace this week and will definitely make the PR ready for review before the end of this month, maybe even next week. I appreciate you volunteering to make a GitHub workflow to cover some of the platforms in the automated testing. Please go ahead!

9ao9ai9ar:
@MagicalDrizzle I've just done the one and only forced push, so please remember to clone my fork again. I also want to ask a favor of you: could you help test the two shells, ash and hush, included in BusyBox? I don't mind you do the testing on Windows, but I don't plan to resolve issues that are specific to the Windows port of BusyBox, or issues pertaining to only WSL/Cygwin/MSYS2 or whatnot.

MagicalDrizzle:
@9ao9ai9ar prefsCleaner works, updater only partly - works only if there's already an user.js file. It won't download a new user.js if doesn't exist, then will try to backup the nonexistent user.js file which fails.
```
+ userjs_backup=/home/demo/Downloads/profile/userjs_backups/user.js.backup.2024-12-01_0511
+ mkdir -p /home/demo/Downloads/profile/userjs_backups
+ cp /home/demo/Downloads/profile/user.js /home/demo/Downloads/profile/userjs_backups/user.js.backup.2024-12-01_0511
cp: cannot stat '/home/demo/Downloads/profile/user.js': No such file or directory
+ userjs_backup=
```

9ao9ai9ar:
@MagicalDrizzle It's the same in the old code, the only difference being that I passed up the error instead of suppressing it. I will check similar blocks of code and fix them in batches after enough issues have accumulated. Keep these bug reports coming!

The new pref.js cleaner might work under normal circumstances, but I know it can fail in more cases than the old code, so I'm still struggling to rewrite it, specifically the part where the actual cleaning happens.

aartoni (commented):
I really like your improvements to the pipeline, I find it way more readable than my version. Are you planning on getting this merged now? I'd be interested in contributing a few more changes after this gets merged, such as an extra step for checking for normal execution

9ao9ai9ar:
> Are you planning on getting this merged now?

There are still, I think, three things I'd like to do before getting this merged:

1. Split the `arkenfox_updater_check_nonroot` function into two to restore the old behavior of checking for root owned files only after running the self update process, so that the script can still update itself in the absence of a Firefox profile.
2. Properly squash, redo and consolidate the commits and commit messages ***one more time*** (apologies for breaking my previous promise). I am aiming for three commits after the rebase: my shell scripts rewrite, @aartoni's testing workflow contributions, and my improvements to the workflow file. The rebase will do away with misleading commit messages (e.g. https://github.com/9ao9ai9ar/user.js/commit/8317f14fea5c4efd03de94c8e5efa072d6e3359d isn't actually runnable on less popular platforms yet) and track the rename of `test.yml` to `portability.yml` for a better diff output, which somehow got messed up by the IDE's Git integration. Initially, I only pushed my branch to GitHub to have a remote backup without expecting people to review or fork it, but I was clearly wrong. Thankfully, this is more a rewrite than a refactor, so it's less important to do clean and piecemeal commits, in my humble opinion.
3. Now that I have put it behind me, reword the comments in the GitHub workflow file in a more pleasant tone?

> I'd be interested in contributing a few more changes after this gets merged, such as an extra step for checking for normal execution

You might want to check out [ShellSpec](https://github.com/shellspec/shellspec). Writing the tests in the workflow file is cumbersome, limiting and a form of vendor lock-in. For example, I can't just run the tests in my VM without setting it up as a self-hosted runner on GitHub or relying on tools such as [act](https://github.com/nektos/act), which you mentioned has some differing behaviors from the hosted runners. Ideally, the only workflow step you need after installing and setting up the shells is to call ShellSpec or some command to invoke the test scripts.

prefsCleaner.sh
```diff
@@ -1,185 +1,519 @@
...
+# This is so that we may determine if the script is sourced or not by
+# comparing it to the basename of the canonical path of $0.
+[ -z "$ARKENFOX_PREFS_CLEANER_NAME" ] &&
+    readonly ARKENFOX_PREFS_CLEANER_NAME='prefsCleaner.sh'
+
+###############################################################################
+####                   === Common utility functions ===                    ####
+#### Code that is shared between updater.sh and prefsCleaner.sh, inlined   ####
+#### and duplicated only to maintain the same file count as before.        ####
+###############################################################################
```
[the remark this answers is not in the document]
9ao9ai9ar: GitHub doesn't seem to notify users of edits to a comment belonging to their subscriptions, so I'll repost my edited reply in https://github.com/9ao9ai9ar/user.js/pull/1#issuecomment-2495417965 here:

> Currently, the amount of shared code is still pretty small and maintainable, so I am inclined to just inline them in the scripts to mirror the layout of their sister batch files.

> I'm fine with the names `libarkenfox.sh` or `.arkenfoxrc`. If we want to put all the scripts in a different directory, I suggest we put them in `scratchpad-scripts` and rename the directory to just `scripts`, as [Scratchpad](https://firefox-source-docs.mozilla.org/devtools-user/deprecated_tools/index.html#scratchpad) had been removed since Firefox 72.

Alternatively, we could combine `updater.sh` and `prefsCleaner.sh` into `arkenfox.sh`, keeping `prefsCleaner.sh`'s menu selection for interactive use. But for this pull request, I'd like to keep the status quo.

aartoni:
Thanks for the pointers!

PR #1909, opened 2024-10-23
Add missing -command argument in powershell call

MagicalDrizzle:
it seems like at least with powershell core - you need the `-command` argument passed to powershell to execute a built-in function.
currently, the backup filename is not `prefs-backup-(date/time).js` but `prefs-backup-All parameters are case-insensitive.js`, because this is the output of the powershell command
```
The argument 'get-date' is not recognized as the name of a script file. Check the spelling of the name, or if a path was included, verify that the path is correct and try again.
...
All parameters are case-insensitive.
```

### DIFF
--- a/prefsCleaner.bat
+++ b/prefsCleaner.bat
@@ -37,7 +37,7 @@ CALL :strlenCheck
 CALL :FFcheck
 
 CALL :message "Backing up prefs.js..."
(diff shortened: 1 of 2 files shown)


strotee (commented):
Pro tip: -command allows one to shorten it by writing -c.

earthlng:
thanks @MagicalDrizzle

PR #1914, opened 2024-11-09
Fixes version parsing and display in `updater.bat`

HorlogeSkynet:
This commit is a (partial) backport of HorlogeSkynet/thunderbird-user.js#59


HorlogeSkynet:
Little up @earthlng ! Thanks :bow:

HorlogeSkynet:
Hey @Thorin-II, would it be possible to merge this fix as it appears unlikely we'll get a review from @earthlng **and** the patch has been live _downstream_ at thunderbird-user.js for almost two years ?
Thanks, bye :wave:

PR #1926, opened 2025-01-03
Update user.js

rusty-snake:
Add missing pref for 3pcd


Thorin-Oakenpants:
https://searchfox.org/mozilla-central/source/modules/libpref/init/StaticPrefList.yaml#12484

sure, I'll add it in #1892

PR #1945, opened 2025-03-10
v140

Thorin-Oakenpants:


sertonix:
Since letterboxing isn't enabled by default anymore the `flip this pref` comment should probably be changed.

Thorin-Oakenpants:
> https://github.com/arkenfox/user.js/blob/7f852e94fc80fd0aa19f1c948319974e3301187f/user.js#L92
> 
> nit: ffx uses `chrome://browser/content/blanktab.html` for years; is there a benefit in using `about:blank` instead or is it just outdated?

IDK if it really matters, but they changed it from `about:blank` in FF108 to the chrome:// schema in https://bugzilla.mozilla.org/show_bug.cgi?id=1676492 (AFAICT). I'll change the pref value to match - not going to read the bugzilla but I do know about:blank can load webcontent, and we can access about:blank via opener from a webpage etc - it's not always simple or clear what the principal is - perhaps this chrome:// schema is more robust/secure - but TBH, I doubt about:blank has any real issues as a landing page (nothing else can interact with it AFAIK)

PR #1947, opened 2025-03-13
Add HIDDEN label to privacy.clearOnShutdown_v2.downloads

rusty-snake:
The merge target is the v136 PR


Thorin-Oakenpants:
I don't think it's hidden, because regardless of it's setting, it is set at runtime?

rusty-snake:
For me, `privacy.clearOnShutdown_v2.downloads` gets not set from the dialog clearOnShutdown dialog on about:preference. Neither true nor false.

The sync pref for this pref is not hidden.

Thorin-Oakenpants:
toggling settings (in the UI) in nightly doesn't create the pref 

So in https://phabricator.services.mozilla.com/D194136 scroll down to `browser/base/content/sanitize_v2.xhtml` line 234 it was added as the pref behind a checkbox called `item-download-history` at the end under site settings

but if you look at https://searchfox.org/mozilla-central/source/browser/base/content/sanitize_v2.xhtml  it's gone

Thorin-Oakenpants:
closing: I'll add a hidden tag to the next version PR because it's clearly hidden, and I literally don't care if the pref is used or not - i'll leave it in out of an abundance of caution

PR #1967, opened 2025-06-08
Turn off an advertising reporting feature

dmarti:
The "dom.private-attribution.submission.enabled" preference controls a feature called "Privacy-Preserving Attribution" co-developed by Mozilla and Meta. This feature creates privacy risks that normal product telemetry does not.

### DIFF
--- a/user.js
+++ b/user.js
@@ -1223,6 +1223,9 @@ user_pref("toolkit.telemetry.firstShutdownPing.enabled", false); // [FF57+]
 user_pref("toolkit.telemetry.coverage.opt-out", true); // [HIDDEN PREF]
 user_pref("toolkit.coverage.opt-out", true); // [FF64+] [HIDDEN PREF]
 user_pref("toolkit.coverage.endpoint.base", "");
+/* 8504: disable advertising reporting
+ * [1] https://www.privacyguides.org/articles/2024/07/14/mozilla-disappoints-us-yet-again-2/ ***/
+user_pref("dom.private-attribution.submission.enabled", false);
 
 /*** [SECTION 9000]: NON-PROJECT RELATED ***/
 user_pref("_user.js.parrot", "9000 syntax error: the parrot's cashed in 'is chips!");


dmarti:
According to a Bugzilla discussion, this ad feature is turned off by default when all telemetry is disabled.

[1908312 - Update settings UI to clarify that PPA is disabled when telemetry is disabled](https://bugzilla.mozilla.org/show_bug.cgi?id=1908312)

However, some users may want to leave the normal telemetry on, because the data feeds into product improvements that benefit all users, while turning the ad tracking features off, because they enable deceptive surveillance advertising practices. (Even though PPA does not track me as an individual, I [choose to turn it off](https://blog.zgp.org/why-turn-off-firefox-ad-tracking/) and leave telemetry turned on)

Thorin-Oakenpants:
Hey :) I read all your entries at https://planet.mozilla.org .. nice work 👍 or as they say, I drink and I know things (cuz I read)

So the reason it wasn't added is because it's covered by the master telemetry switch. I recently moved telemetry into it's own section so I could explain it (Telemetry ) was harmless, and I guess, now with that explained, some users might choose to add override prefs to enable it all back on (it's quite a few prefs) but IDK if anyone would actually bother.

I have just arrived back from a month away in Europe (work related) so I need some time to catch up - that and I need a break. I will need to revisit _exactly_ what PPA does. If it has any privacy concerns (I need to refresh my memory) then I wouldn't put in the telemetry section since that is all about how it isn't a privacy concern - if there is a privacy concern then more likely I will add it to `section 0300` as a fallback (active, i.e not commented out) should users enable telemetry

Don't fret if nothing happens for a while, I think I'll mostly be AWOL til July

dmarti:
Thank you. Personally I leave the regular telemetry on, because my interests and the developer's interests in having a working browser are aligned. The results of product telemetry are fed into making a better product for all users, which imho is a win-win.

But -- especially as we shift to a lower-trust economy -- I don't have that kind of shared interest with the kinds of players who would best be able to use the PPA data. (more: [why I’m turning off Firefox ad tracking: the PPA paradox](https://blog.zgp.org/why-turn-off-firefox-ad-tracking/))

(I mostly agree with noyb.eu on this feature: [Firefox tracks you with “privacy preserving” feature](https://noyb.eu/en/firefox-tracks-you-privacy-preserving-feature))

Breaking PPA out as a separate item that people can easily turn off without turning off telemetry would be doing Mozilla a favor. When they eventually drop this feature (as they have done with previous, far less problematic, in-browser ads) it will be better for them to have more users left with the regular telemetry still on.

Enjoy your break and happy to help when you get back.

Thorin-Oakenpants:
> Extensions I’m currently not using but want to .... Cookie Auto Delete

I have some news for you @dmarti [scroll down to cookie extensions](https://github.com/arkenfox/user.js/wiki/4.1-Extensions#-dont-bother) ... Cookie Auto Delete does not work with partitioning which has been the default in Firefox for about 2 years from memory - and this is essential. The fact they tell users to disable that just makes it even all the more worse and they should be banned and avoided at all costs - they clearly have no clues

Thorin-Oakenpants:
Hah, so it's never been used (wasn't interested in following it since the UI was hidden and covered by another pref we have) - not much point adding anything then

PR #1975, opened 2025-07-16
Turn off sending daily pings to Mozilla

Rafee-M:
Disable setting under  Privacy & Security> Firefox Data Collection and Use> Send daily usage ping to Mozilla


Rafee-M:
Ah my bad. Don't know how I missed that. Cheers!

PR #2029, opened 2026-01-22
Support the XDG base directory based default profile path

junkblocker:
Starting with Firefox 147, Firefox supports XDG base directory specification

### DIFF
--- a/updater.sh
+++ b/updater.sh
@@ -159,6 +159,8 @@ readIniFile() { # expects one argument: absolute path of profiles.ini
 getProfilePath() {
   declare -r f1=~/Library/Application\ Support/Firefox/profiles.ini
   declare -r f2=~/.mozilla/firefox/profiles.ini
+  declare -r f3="${XDG_CONFIG_HOME:-"$HOME/.config"}/mozilla/firefox/profiles.ini"
+  declare -r f4=~/.config/mozilla/firefox/profiles.ini
 
   if [ "$PROFILE_PATH" = false ]; then
     PROFILE_PATH="$SCRIPT_DIR"
@@ -167,6 +169,10 @@ getProfilePath() {
       readIniFile "$f1" # updates PROFILE_PATH or exits on error
     elif [[ -f "$f2" ]]; then
       readIniFile "$f2"
+    elif [[ -f "$f3" && $f3 == /* ]]; then
+      readIniFile "$f3"
+    elif [[ -f "$f4" ]]; then
+      readIniFile "$f4"
     else
       echo -e "${RED}Error: Sorry, -l is not supported for your OS${NC}"
       exit 1


updater.sh
```diff
@@ -158,7 +158,8 @@ readIniFile() { # expects one argument: absolute path of profiles.ini
 
 getProfilePath() {
   declare -r f1=~/Library/Application\ Support/Firefox/profiles.ini
-  declare -r f2=~/.mozilla/firefox/profiles.ini
+  declare -r f2=~/.config/mozilla/firefox/profiles.ini
```
Neurognostic: You should also check the environment variable `XDG_CONFIG_HOME` as the `mozilla` directory may be located there instead of the default path.

junkblocker: Done.

9ao9ai9ar: You should also check the environment variable `XDG_CONFIG_DIRS`, too, which defines the set of base directories where additional configuration files should be searched. More importantly, all the environment variables defined by the XDG Base Directory Specification must be set to absolute paths or ignored otherwise. Oh, and the read-only configs are searched after the read-write config, and it is up to the implementation to define what happens if the same information appears in multiple configs.

But I doubt Mozilla, like most other vendors under the sun, cares enough to properly implement the spec which is considered a failure even by its authors. For the record, Flatpak disregards the XDG spec entirely. Since the Arkenfox shell scripts already provide the option to specify the path to your Firefox profile, your first commit is sufficient since no beginner user knows about or should mess with `XDG_CONFIG_HOME`.

rusty-snake: You forgot to say that mozilla missed that `XDG_CONFIG_HOME` is only for config files and data/state files belong in other xdg directories.
Anyway the updater script must not implement the spec but reimplement the logic of firefox.

And btw https://searchfox.org/firefox-main/rev/c8bed8c9f4cdcbfa8d24fc355ca72adff8c85364/toolkit/xre/nsXREDirProvider.cpp#1423:
>   // Explicitly check for rv failure because in case we get passed an env
  // value that is an invalid dir by the XDG specification level, it should
  // be ignored. Per
  // https://specifications.freedesktop.org/basedir-spec/0.8/:
  // "If an implementation encounters a relative path in any of
  // these variables it should consider the path invalid and ignore it."

9ao9ai9ar: Mozilla Firefox is just one of many apps that put their data/state files under `XDG_CONFIG_HOME`, and sometimes what counts as config vs data/state is not really clear-cut, so I do not consider the misplacement the biggest problem with the XDG spec or the apps that try to support it.

Reading back on my previous comment, I should clarify that my somewhat sarcastic remark was directed at the poor state of adequate support and the perceived design shortcomings of the XDG spec, not at the OP or anyone else, so if I have offended anyone, I apologize for how it was poorly communicated as that was not my intention.

I anticipate this migration to XDG paths is going to confuse/upset at least some users. What happens when a user has both `~/.config/mozilla/firefox/profiles.ini` and `~/.mozilla/firefox/profiles.ini` present? I think for backward compatibility, most apps would only apply the configuration at the old location.

junkblocker: None taken. I agree with this analysis and have updated the PR.

updater.sh
```diff
@@ -168,8 +169,10 @@ getProfilePath() {
       readIniFile "$f1" # updates PROFILE_PATH or exits on error
     elif [[ -f "$f2" ]]; then
       readIniFile "$f2"
-    elif [[ -f "$f3" ]]; then
+    elif [[ -f "$f3" ]] && [[ $f3 == /* || $f3 == ~* ]]; then
```
Neurognostic:
```bash
    elif [[ -f "$f3" && $f3 == /* ]]; then
```

`f3` will never contain the literal string `~` and if somehow it did it would not be compliant with the XDG spec. so it should not pass the conditional. The `~` in the parameter assignments above expand to the value of `$HOME` (e.g., `/home/myuser`). Simply checking that it starts with `/` is enough to validate whether the path is an absolute path and is in compliance with the spec.

junkblocker: Addressed.

costowell:
Bump! I would love to see this!

Issue #2017, opened 2025-11-11
Add more prefs to 7016


Tiagoquix:
`privacy.trackingprotection.emailtracking.enabled` is set to true with Strict it seems. could be added to the customize etp settings thing.

there is also `privacy.trackingprotection.consentmanager.skip.pbmode.enabled` but I am not sure if it's set with Strict.

Thanks.

Thorin-Oakenpants:
section 7000's is a `DON'T BOTHER` - it's not a "customize etp settings" it's a "users reset this shit when running prefsCleaner thing" because many users get bad advice or settings from elsewhere, or change things in about:config and forget about it

OK, so lets find out from diffs when these were added to code and what they actually do - like I mean WTF is emailtracking protection

Thorin-Oakenpants:
user_pref("network.cookie.cookieBehavior.optInPartitioning", true); // [ETP FF132+]

this doesn't get set to true anymore with ETP strict - we might be able to remove it, I'm just not sure when that change happened and that this pref actually does

Tiagoquix:
> section 7000's is a DON'T BOTHER - it's not a "customize etp settings" it's a "users reset this shit when running prefsCleaner thing" because many users get bad advice or settings from elsewhere, or change things in about:config and forget about it

yes. what I meant is to add it there as to be part of the list of the don't bother things.

Thorin-II:
also see https://github.com/arkenfox/user.js/pull/2056/changes/7602fda9fc4058f53abe5e528942a2afe0bccf74

PR #2056, opened 2026-06-02
v153

Thorin-II:
@tomrittervg added a link to your blog about telemetry

### DIFF
--- a/user.js
+++ b/user.js
@@ -764,7 +764,8 @@ user_pref("_user.js.parrot", "4000 syntax error: the parrot's bereft of life!");
    1653987 - limit font visibility to bundled and "Base Fonts" (Windows, Mac, some Linux) (FF80)
    1461454 - spoof smooth=true and powerEfficient=false for supported media in MediaCapabilities (FF82)
     531915 - use fdlibm's sin, cos and tan in jsmath (FF93, ESR91.1)
-   1692609 - reduce JS timing precision to 16.67ms (previously FF55+ was 100ms) (FF102)
+   1692609 - reduce timing precision via rAF (requestAnimationFrame) to 16.67ms | i.e FPS are 60hz (FF102)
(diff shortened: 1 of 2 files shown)


user.js
any1here: `network.cookie.cookieBehavior.optInPartitioning` hasn't been set to `true` in ETP Strict [since FF137](https://bugzilla.mozilla.org/show_bug.cgi?id=1951443)

SkewedZeppelin:
typo found by @cerox
https://github.com/arkenfox/user.js/blob/8fe9905c35a1025d1e6df69479f7625585fd956d/user.js#L942

user.js
Thorin-II: https://github.com/arkenfox/user.js/issues/2017#issuecomment-3522642571

Thorin-II: let's just remove it

any1here: > [#2017 (comment)](https://github.com/arkenfox/user.js/issues/2017#issuecomment-3522642571)

> and that this pref actually does

It blocked all non‑CHIPS third‑party cookies, unless they were on an exemption list or certain [heuristics](https://developer.mozilla.org/en-US/docs/Web/Privacy/Guides/State_Partitioning#storage_access_heuristics) also allowed them.

HorlogeSkynet:
```suggestion
 * [1] https://bugzilla.mozilla.org/show_bug.cgi?id=1981587 ***/
```

Rustlyr:
What new update. ?

62f1346c-4576-42b3-afcb-23ef1820da34:
Please add a warning like this somewhere:

If _browser.profiles.enabled_ is set to _true_ (default since FF150) all profiles share the same settings for _Permissions and data>Firefox data collection and use_ and the corresponding user.js prefs are _not_ honored.
You might want to check and manually adjust those settings in the GUI of one of your profiles.

